PR #71, opened 2025-01-11
moved some time/sync funcs/consts out from syscall_utils

0xeb:
...to utils/timer.hpp


src/common/utils/timer.hpp
```diff
@@ -23,4 +27,80 @@ namespace utils
...
+    using __time64_t = int64_t;
+#endif
+
+    inline LARGE_INTEGER convert_unix_to_windows_time(const __time64_t unix_time)
+    {
+        LARGE_INTEGER windows_time{};
+        windows_time.QuadPart = (unix_time + EPOCH_DIFFERENCE_1601_TO_1970_SECONDS) * HUNDRED_NANOSECONDS_IN_ONE_SECOND;
+        return windows_time;
+    }
 }
```
momo5502: i would create a time.hpp/time.cpp (instead of timer, as that really is only a timer and imo should stay that way)
and then all the inline functions can be moved into the cpp and only the declarations can stay in the time.hpp

Mizokuiam (commented):
.

Mizokuiam:
Great observation! Let me add some thoughts...

momo5502 (approved):

momo5502:
LGTM. As soon as CI passes, this can be merged

PR #75, opened 2025-01-12
54 cross platform api set support

robert-yates:
This PR provides configurable APISET data for the PEB

There are multiple ways the APISET data can be provided:

* host -> read the APISET from the host process
* file -> read the APISET from an external file
* w11  -> use an inbuilt dump from windows 11
* w10  -> use an inbuilt dump from windows 10

by default on windows it uses the "host" mode.
by default on linux/mac it uses the "w11" mode.

with the file mode you can use the dump-apiset tool to create a dump.
however the inbuilt w10 works fine on w11 so I don't expect it will be needed.


.gitmodules
```diff
@@ -14,3 +14,6 @@
 [submodule "deps/googletest"]
 	path = deps/googletest
 	url = https://github.com/google/googletest.git
+[submodule "deps/zlib"]
+	path = deps/zlib
+	url = https://github.com/madler/zlib.git
```
momo5502, line 19: I tend to use dev branch. Not sure if it makes sense. But ignore dirty makes sense as the zlib cmake scripts modifies some files
```suggestion
	url = https://github.com/madler/zlib.git
	branch = develop
	ignore = dirty
```

src/tools/dump-apiset/dump-apiset.cpp
```diff
@@ -0,0 +1,118 @@
...
+    return (PVOID)__readfsdword(0x30);
+#endif
+}
+
+int main()
+{
+    printf("Dump API-SET\n");
+    printf("------------\n\n");
+
+    const auto peb = (PPEB64)GetCurrentProcessPeb();
```
momo5502, line 25: do we have PEB32 structs already? might make sense to just use PPEB here so that it defaults to PEB32 on x86 processes, as GetCurrentProcessPeb supports 32 bit

src/windows-emulator/windows_emulator.cpp
```diff
@@ -133,15 +137,55 @@ namespace
...
+            throw std::runtime_error("The APISET host location is not supported on this platform");
 #endif
+        case apiset_location::file: {
+            auto apiset = utils::io::read_file("api-set.bin");
+            if (apiset.empty())
+                throw std::runtime_error("Failed to read file api-set.bin");
+            auto buffer = utils::compression::zlib::decompress(apiset);
+            if (buffer.empty())
+                throw std::runtime_error("Failed to decompress API-SET");
+            return buffer;
```
momo5502: maybe extract those 4 lines into a function. they seem to repeat quite a lot?

.gitmodules
robert-yates: after some git fighting i think i changed it :)

src/tools/dump-apiset/dump-apiset.cpp
robert-yates: not yet, structs that end in 64 are explicit 64 ones that i didnt template because i wasnt sure they would map 1:1 or i was too fatigued to do

src/windows-emulator/windows_emulator.cpp
robert-yates: yup, done !

PR #76, opened 2025-01-12
Changes to make the 'emulator' project suitable for submodules

0xeb:
- Added the MOMO_BUILD_AS_LIBRARY CMake option to allow ease of using of the emulator project as a library/submodule.
- windows_emulator project also brings in the 'common' dependency
- 'common' dependency now renamed to 'emulator-common'


momo5502:
Great addition. I'm afraid this will need an exported interface to be really usable (as in visibility default/dllexport annotation at a few classes/methods). But that can be added once we see it doesn't work

momo5502 (approved):

0xeb:
Thanks. I already use the emulator as a static library in a project. 

Let’s see how this evolves.

PR #77, opened 2025-01-13
Add Android support

momo5502:


src/windows-emulator/registry/registry_manager.hpp
```diff
@@ -33,7 +33,7 @@ class registry_manager
 {
   public:
     using hive_ptr = std::unique_ptr<hive_parser>;
-    using hive_map = std::unordered_map<std::filesystem::path, hive_ptr>;
+    using hive_map = std::unordered_map<std::u16string, hive_ptr>;
```
momo5502: not good. these keys need to be in a canonicalized form

momo5502: done. tthere is now a thing called path_key

PR #79, opened 2025-01-15
Exposing high level callbacks

0xeb:
The idea here is to start exposing various high level callbacks from various sub components (syscalls, device io, etc.).

Consumer of `windows_emulator` can then hook into any callback.

Please give your feedback on this initial commit.

### DIFF
--- a/src/windows-emulator/windows_emulator.hpp
+++ b/src/windows-emulator/windows_emulator.hpp
@@ -16,7 +16,8 @@ struct emulator_callbacks
                        std::string_view syscall_name)>
         inline_syscall{};
     std::function<void(uint32_t syscall_id, x64_emulator::pointer_type address, std::string_view mod_name,
-                       std::string_view syscall_name, x64_emulator::pointer_type prev_address, std::string_view prev_mod_name)>
+                       std::string_view syscall_name, x64_emulator::pointer_type prev_address,
+                       std::string_view prev_mod_name)>
         outofline_syscall{};
 };
 
@@ -127,7 +128,6 @@ class windows_emulator
     bool silent_until_main_{false};
     std::unique_ptr<x64_emulator> emu_{};
     std::vector<instruction_hook_callback> syscall_hooks_{};
-    std::function<void(std::string_view)> stdout_callback_{};
 
     process_context process_;
     syscall_dispatcher dispatcher_;
--- a/src/windows-emulator/windows_emulator.hpp
+++ b/src/windows-emulator/windows_emulator.hpp
@@ -9,6 +9,17 @@
 
 std::unique_ptr<x64_emulator> create_default_x64_emulator();
 
+struct emulator_callbacks
+{
+    std::function<void(std::string_view)> stdout_callback{};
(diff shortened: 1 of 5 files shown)


src/windows-emulator/windows_emulator.hpp
```diff
@@ -88,13 +100,12 @@ class windows_emulator
         this->syscall_hooks_.push_back(std::move(callback));
     }
```
momo5502, line 101: i like the idea. however, i'm not sure if the distinction of inline/outofline syscalls should be done by the emulator. this is an analysis that feels out of place at the moment. in my head, the windows emulator should only perform the emulation. any kind of analysis should be done ontop of it. as e.g. theremight be use cases for the emulator that don't need the overhead of the analysis. so maybe it might make sense to just reuse this generic syscall_hook and and move it into the callbacks and the distinction of inline/outofline can be done outside, within the analyzer

src/windows-emulator/windows_emulator.hpp
```diff
@@ -26,7 +37,8 @@ class windows_emulator
 {
   public:
     windows_emulator(std::unique_ptr<x64_emulator> emu = create_default_x64_emulator());
-    windows_emulator(emulator_settings settings, std::unique_ptr<x64_emulator> emu = create_default_x64_emulator());
+    windows_emulator(emulator_settings settings, emulator_callbacks callbacks = {},
```
momo5502, line 40: at the moment this is passed through the constructor. this means if i want to dynamically register a callback, i can't. iguess this is fine for now. but maybe in the future we can have a way of dynamicall updating the callbacks (e.g. register/deregister callbacks on demand)

momo5502 (changes requested):

src/windows-emulator/windows_emulator.hpp
0xeb: correct, windows emulator should just emulate. however, lots of hardcoded detection logic inside of it that I want to extract out.
Re-using hooks API in various cases looks counter-intuitive. Since this is a "windows emulator" it should expose higher level logic so one can hook on top of the windows emulator and not go down to a deeper level and then try to reason from a lower level.

let me surface more callbacks and we can have more discussion.

src/windows-emulator/windows_emulator.hpp
0xeb: yes, let's keep figuring out for now until we refactor cleanly.

momo5502:
> This process will take some time and back and forth to get it started.
> 
> `windows_emulator` should not log anything, it should give enough high level callbacks. the `analyzer` app should then use the callbacks and do the logging.
> 
> the `analyzer` should be the reference consumer of the library `windows_emulator`.
> 
> in most cases `windows_emulator` will hide low level details and expose the synthesis via usable / relevant windows-like callbacks.
> 
> the consumer should not repeat windows logic/semantics. consumers should benefit from having `windows_emulator` expose convenient high level observations (hooks, relevant windows-like events), etc.

i very much like this idea.

0xeb:
thanks for merging this now. we can refactor everything soon. i plan to add a lot of callbacks (we can organize later) and figure out a cleaner way to go about that. but one step at a time.

PR #83, opened 2025-01-16
Exclude 'zlib' shared project from emulator lib

0xeb:
Shared targets are not compatible with the emulator lib.


momo5502 (changes requested):
I woud move that to deps/zlib.cmake and always disable it, not only if building as library. i don't think we need the shared project

Issue #53, opened 2025-01-06
Create file system abstraction


momo5502:
Files are accessed using fopen/std::filesystem
There needs to be an abstraction to encapsulate accessing files and resolving paths for cross platform support

This also opens up the possibility to redirect certain files (like qiling supports it :3)

momo5502:
Maybe something likee this:


rootfs/
├─ filesystem/
│  ├─ C/
│  │  ├─ Users/
│  │  │  ├─ EmulatorUsername/
│  │  │  │  ├─ ...
│  │  ├─ Windows/
│  │  │  ├─ System32/
│  │  │  │  ├─ ntdll.dll
│  │  │  │  ├─ ...
├─ api-set.bin
├─ registry/
│  ├─ NTUSER.DAT
│  ├─ SYSTEM
│  ├─ ...

momo5502:
mac & linux are case sensitive, which probably needs to be respected

0xeb:
I like the rootfs idea. I removed the comments. We can repurpose '-r' to point to rootfs when implemented.

PR #84, opened 2025-01-16
added '-r' switch to the analyzer

0xeb:
command line switch to specify the registry path to use with the analyzer

### DIFF
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -206,12 +206,11 @@ namespace
             }
             else if (arg == "-r")
             {
-                // Make sure there is an argument following "-r"
                 if (args.size() < 2)
                 {
                     throw std::runtime_error("No registry path provided after -r");
                 }
-                arg_it = args.erase(arg_it); // Remove the "-r" token from args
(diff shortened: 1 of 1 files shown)


src/analyzer/main.cpp
```diff
@@ -202,6 +204,16 @@ namespace
             {
                 options.concise_logging = true;
             }
+            else if (arg == "-r")
+            {
+                // Make sure there is an argument following "-r"
```
momo5502: no need for the two comments. the code is self explanatory

momo5502, line 209: we can do it like this for now. but as part of #53 i want to introduce a root-fs concept similar to qiling that will probably change how the registry is loaded and also everything else

momo5502:
we can do it like this for now. but as part of https://github.com/momo5502/emulator/issues/53 i want to introduce a root-fs concept similar to qiling that will probably change how the registry is loaded and also everything else

0xeb:
I like the rootfs idea. I removed the comments. We can repurpose '-r' to point to rootfs when implemented.

PR #86, opened 2025-01-17
fixed typo in 'committed_region' struct

0xeb:


momo5502 (approved):
lol. i thought i fixed that already. maybe i had that typo twice 😂

PR #88, opened 2025-01-18
Refactor GDB stub

momo5502:
The intention of this PR is not yet to extend/fix the GDB functionality, but rather to get rid the `mini-gdbstub` dependency.

Its functionality should have been reimplemented as closely as possible.

The main reason for this is better portability and security fixes (there are numerous OOB read/write vulnerabilities in `mini-gdbstub`)


0xeb:
the code changes look ok, however it seems debugging with IDA stopped working:

image

Can you verify and tell me to test again?

0xeb:
I fixed the first error by resetting all the saved answers (hidden ida dialogs) and now I have another issue that does not repro in main branch.

image


I have a breakpoint at the EXE entry, but IDA reports errors (unlike from the last time as in main branch).

image

src/gdb-stub/gdb_stub.cpp
```diff
@@ -0,0 +1,492 @@
...
+            }
+
+            connection.send_reply("OK");
+        }
+
+        void read_single_register(const connection_handler& connection, debugging_handler& handler,
+                                  const std::string& payload)
+        {
+            size_t reg{};
+            rt_assert(sscanf_s(payload.c_str(), "%zx", &reg) == 1);
```
robert-yates, line 214: this reads the incorrect register, it should be %zd i believe

src/gdb-stub/gdb_stub.cpp
```diff
@@ -0,0 +1,492 @@
...
+            }
+        }
+
+        void write_single_register(const connection_handler& connection, debugging_handler& handler,
+                                   const std::string_view payload)
+        {
+            const auto [reg, hex_data] = split_string(payload, '=');
+
+            size_t register_index{};
+            rt_assert(sscanf_s(std::string(reg).c_str(), "%zx", &register_index) == 1);
```
robert-yates, line 237: probably this also. @0xeb maybe you can test this :)

src/gdb-stub/gdb_stub.cpp
momo5502: Interesting. I copied the implementation from here: https://github.com/momo5502/mini-gdbstub/blob/main/lib/gdbstub.c#L196

momo5502: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#Packets

image

%zx seems correct

momo5502:
> I fixed the first error by resetting all the saved answers (hidden ida dialogs) and now I have another issue that does not repro in main branch.
> 
> image
> 
> I have a breakpoint at the EXE entry, but IDA reports errors (unlike from the last time as in main branch).
> 
> image

i can reproduce it :( will look into it

image

src/gdb-stub/gdb_stub.cpp
robert-yates: ah ok, then i remembered wrong !

momo5502:
> i can reproduce it :( will look into it
> 
> image

Fixed, this is because not all registers are supported right now, so returning false if register reading failed is not good for now.

src/gdb-stub/gdb_stub.cpp
momo5502: image

same for this. %zx should be fine

robert-yates (approved):

0xeb:
it works, confirmed too! thank you both!

PR #91, opened 2025-01-19
Introducing reflection levels concept into core components

0xeb:
The idea is to allow as much internal information into inner components. to not burden all builds, the reflection level can be controlled via the MOMO_REFLECTION_LEVEL (where 0 means no reflection code is included).

more reflection variables will be introduced as needed.

For now, the memory manager's layout state version is used to track whether the memory layout is changed or not (at the lowest level). the win_emu API consumer can use this reflection to decide to whether to refresh or not expensive computations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,10 @@ cmake_minimum_required(VERSION 3.26.4)
 option(MOMO_ENABLE_SANITIZER "Enable sanitizer" OFF)
 option(MOMO_BUILD_AS_LIBRARY "Configure and Build the emulator as a shared library (without the samples and tests)" OFF)
 
+set(MOMO_REFLECTION_LEVEL "0" CACHE STRING "Reflection level for the build")
+message(STATUS "Reflection level is set to: ${MOMO_REFLECTION_LEVEL}")
+add_compile_definitions(MOMO_REFLECTION_LEVEL=${MOMO_REFLECTION_LEVEL})
+
 if(MOMO_BUILD_AS_LIBRARY)
     add_compile_definitions(MOMO_BUILD_AS_LIBRARY=1)
 else()
--- a/src/emulator/memory_manager.hpp
+++ b/src/emulator/memory_manager.hpp
@@ -1,9 +1,7 @@
 #pragma once
 #include <map>
-#if MOMO_REFLECTION_LEVEL > 0
 #include <atomic>
 #include <cstdint>
-#endif
 
 #include "memory_region.hpp"
 #include "address_utils.hpp"
@@ -127,12 +125,10 @@ class memory_manager
         return reserved_regions_;
     }
 
-#if MOMO_REFLECTION_LEVEL > 0
     std::uint64_t get_memory_layout_state_ver() const
     {
-        return memory_layout_state_ver_.load(std::memory_order_relaxed);
+        return memory_layout_state_version_.load(std::memory_order_relaxed);
     }
-#endif
 
   private:
     reserved_region_map reserved_regions_{};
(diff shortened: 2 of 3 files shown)


src/emulator/memory_manager.hpp
```diff
@@ -133,6 +144,15 @@ class memory_manager
     virtual void apply_memory_protection(uint64_t address, size_t size, memory_permission permissions) = 0;
 
   protected:
+#if MOMO_REFLECTION_LEVEL > 0
+    std::atomic<std::uint64_t> memory_layout_state_ver_{0};
+
+    void inc_memory_layout_state_ver()
```
momo5502: I'm not a fan of abreviations.
i think something like `increase_memory_layout_state` might be better

CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ cmake_minimum_required(VERSION 3.26.4)
 option(MOMO_ENABLE_SANITIZER "Enable sanitizer" OFF)
 option(MOMO_BUILD_AS_LIBRARY "Configure and Build the emulator as a shared library (without the samples and tests)" OFF)
 
+set(MOMO_REFLECTION_LEVEL "0" CACHE STRING "Reflection level for the build")
+message(STATUS "Reflection level is set to: ${MOMO_REFLECTION_LEVEL}")
+add_compile_definitions(MOMO_REFLECTION_LEVEL=${MOMO_REFLECTION_LEVEL})
+
```
momo5502, line 11: is this needed? all this code essentially does is a relaxed atomic increment, which should be bretty cheap. 
in my opinion the gained performance is neglectable, but the ifs polute the code.

if this is needed, i would move the if into inc_memory_layout_state_ver so that it's not needed around every call

CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ cmake_minimum_required(VERSION 3.26.4)
...
 if(MOMO_BUILD_AS_LIBRARY)
```
momo5502, line 12: also, I really hate the MOMO prefix 😂 if you have any recommendations for a better prefix, that is maybe more related to the project, I'd greatly appreciate it :D

CMakeLists.txt
0xeb: For now, I am doing increment for the memory manager. but what if later I am exposing more internal stuff that we don't want outside in regular builds?

For instance, think capstone and detail level. you need to compile the detail level in to get more behavior. i can remove that for now, but not sure what other internal stuff i might need down the line.

what's your take?

CMakeLists.txt
0xeb: I guess, we have two choices: `EMU_` or `EMULATOR_`.

Whatever we pick, we have to rename the other defines that are `MOMO_` prefixed.

CMakeLists.txt
robert-yates: i haven't looked too much at the general architecture of things yet, but at some point i guess there will be a lot of interface changes to expose out important elements for apis and bindings.

when it comes to exposing internals i think its important to define end user's use cases and what would be required and work backwards.

so with that in mind here are some of my own end user ideas and thus things i think are important.

Emulation integration into RE tools: IDA and BinaryNinja:

* cpu registers
* memory layout
* modules
* on demand snapshot save and load, queries, diffing, analysis etc etc ...
* event notifications:
 - execution left main modules image size
 - user hooks on memory range read/write

i envision a workflow where the emulator could be used represent a processes state as part of a large scale RE workflow so that you can with easy have a reproducible static environment, reset snapshots , jump back and forth in time, track and monitor various elements, and wrap that all up into IDA/Binja into a nice GUI plugin that communicates to the emulators api.

interested in what others think, thats my rambley large scale future ideas :) however first i want to get cross platform working for general windows RE/fuzzing and small CTFs (-:

/rob.

0xeb: I agree and share the same goals. All the small changes and exposing of stuff I am doing is to meet a very practical goal. Give me couple of weeks to show the end result. 

I am onboard about refactoring and cleaning as we solidify the use cases and workflow.

momo5502 (approved):

CMakeLists.txt
momo5502: i like both. maybe i'll create a PR to change everything in a bulk in the coming days

CMakeLists.txt
momo5502: i  alsoreally like those ideas and share the same goals :D

CMakeLists.txt
0xeb: yes, exactly. i was waiting for your reply to make a separate rename PR.
Go for it with your bulk changes then!

PR #96, opened 2025-01-22
some refactoring with optional_function

0xeb:
- wrapped std::function into utils::optional_function 
  - cleaned the code accordingly in windows_emulator
- using the 'emulator'/'windows_emulator' dependency implies the emulator_common as well.

### DIFF
--- a/src/common/utils/function.hpp
+++ b/src/common/utils/function.hpp
@@ -1,0 +1,48 @@
+#pragma once
+
+#include <functional>
+
+namespace utils
+{
+    template <typename Signature>
+    class optional_function;
+
+    template <typename Ret, typename... Args>
+    class optional_function<Ret(Args...)>
+    {
+      private:
+        std::function<Ret(Args...)> func;
+
+      public:
+        optional_function() = default;
+        optional_function(std::function<Ret(Args...)> f)
+            : func(std::move(f))
+        {
+        }
+
+        optional_function& operator=(std::function<Ret(Args...)> f)
+        {
+            func = std::move(f);
+            return *this;
+        }
+
+        Ret operator()(Args... args) const
+        {
+            if (func)
+            {
+                return func(std::forward<Args>(args)...);
+            }
+            else
+            {
+                if constexpr (!std::is_void_v<Ret>)
+                {
+                    return Ret();
+                }
+            }
+        }
+        explicit operator bool() const
+        {
+            return static_cast<bool>(func);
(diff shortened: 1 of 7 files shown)


0xeb:
@momo5502 does it make sense to remove the function_wrapper.hpp and move its code to function.hpp? or prefer one file per class?

0xeb:
@momo5502 can you please help with the GCC build. Not sure how to make the compiler happy.

image

src/common/utils/function.hpp
```diff
@@ -0,0 +1,48 @@
...
+        {
+        }
+
+        optional_function& operator=(std::function<Ret(Args...)> f)
+        {
+            func = std::move(f);
+            return *this;
+        }
+
+        Ret operator()(Args... args) const
```
momo5502, line 36: Args&&...

src/common/utils/function.hpp
```diff
@@ -0,0 +1,48 @@
...
+            {
+                return func(std::forward<Args>(args)...);
+            }
+            else
+            {
+                if constexpr (!std::is_void_v<Ret>)
+                {
+                    return Ret();
+                }
+            }
```
momo5502:
```suggestion

            if constexpr (!std::is_void_v<Ret>)
            {
                return Ret();
            }
```

momo5502:
> @momo5502 can you please help with the GCC build. Not sure how to make the compiler happy.
> 
> image

my first guess is that the compiler doesn't perform the required conversions, lambda -> function -> optional_function. 
you might need a templated constructor that forwards all paramters to std::function so that things compile

src/common/utils/function.hpp
momo5502: this is not fixed yet

src/common/utils/function.hpp
```diff
@@ -0,0 +1,53 @@
...
+            return *this;
+        }
+
+        Ret operator()(Args... args) const
+        {
+            if (func)
+            {
+                return func(std::forward<Args>(args)...);
+            }
+            if constexpr (!std::is_void_v<Ret>)
```
momo5502, line 43: empty line pls :3

```suggestion

            if constexpr (!std::is_void_v<Ret>)
```

0xeb: yes! the formatter did not catch that.

momo5502 (approved):

PR #99, opened 2025-01-23
Add NtUserRegisterClassExWOW syscall implementation

reddead-coder:
This PR introduces the implementation of the NtUserRegisterClassExWOW syscall. The purpose of this addition is to handle the registration of window classes in the emulation layer more efficiently.

Key changes:

Implemented a method to read and store class names from memory.
Added logic to allocate atom indices and prevent duplication.
Logged the full class name and atom for debugging purposes.
Request for Feedback:
I would appreciate the community's feedback on this implementation. Specifically:

Is the logic for handling atoms and avoiding duplicates robust enough?
Are there any potential issues or edge cases that I might have missed?
Do you have any suggestions to optimize the performance of this function?
Please let me know your thoughts or suggest any improvements. Thank you for reviewing this!

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -3413,6 +3413,87 @@ namespace
         c.win_emu.yield_thread();
         return STATUS_SUCCESS;
     }
+    typedef struct _UNICODE_STRING
+    {
+        USHORT Length;
+        USHORT MaximumLength;
+        PWSTR Buffer;
+    } UNICODE_STRING;
+    typedef UNICODE_STRING* PUNICODE_STRING;
+    typedef struct _CLSMENUNAME
+    {
+        LPSTR pszClientAnsiMenuName;
+        LPWSTR pwszClientUnicodeMenuName;
+        PUNICODE_STRING pusMenuName;
+    } CLSMENUNAME, *PCLSMENUNAME;
+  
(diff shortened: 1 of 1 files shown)


src/windows-emulator/syscalls.cpp
```diff
@@ -3413,6 +3413,84 @@ namespace
         c.win_emu.yield_thread();
         return STATUS_SUCCESS;
     }
+    typedef struct _UNICODE_STRING
+    {
+        USHORT Length;
+        USHORT MaximumLength;
+        PWSTR Buffer;
+    } UNICODE_STRING;
+    typedef UNICODE_STRING* PUNICODE_STRING;
```
momo5502: unicode string already exists as struct: `UNICODE_STRING<EmulatorTraits<Emu64>>`

src/windows-emulator/syscalls.cpp
```diff
@@ -3413,6 +3413,84 @@ namespace
...
+                    if (diff > 1)
+                    {
+                        index = *last_entry + 1;
+                    }
+                }
+            }
+
+            last_entry = entry.first;
+        }
+        c.win_emu.log.print(color::gray, "Full ClassName: %ls , atom : %u\n", name.c_str(), index);
```
momo5502: feels like a lot of code is duplicated from NtAddAtomEx. can this be unified? e.g. this calls handle_NtAddAtom or can the logic be extracted?

PR #101, opened 2025-01-23
Optimize handle_NtUserRegisterClassExWOW Implementation

reddead-coder:
hi

I’ve refactored and optimized the handle_NtUserRegisterClassExWOW function. Leveraging handle_NtAddAtomEx for better reusability and consistency.

Issue:
The current implementation allocates memory (allocate_memory) for the atom but does not free it afterward. I'm unsure how to properly release this memory, so any advice or suggestions on handling this would be greatly appreciated.

Please review the changes and share your feedback or potential solutions to address the memory management issue!

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -3413,6 +3413,31 @@ namespace
         c.win_emu.yield_thread();
         return STATUS_SUCCESS;
     }
+    typedef UNICODE_STRING<EmulatorTraits<Emu64>> *PUNICODE_STRING;
+    typedef struct _CLSMENUNAME
+    {
+        LPSTR pszClientAnsiMenuName;
+        LPWSTR pwszClientUnicodeMenuName;
+        PUNICODE_STRING pusMenuName;
+    } CLSMENUNAME, *PCLSMENUNAME;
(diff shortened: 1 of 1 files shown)


alextrack2013:
why use AI to write the PR comment lol

0xeb:
> why use AI to write the PR comment lol

language barrier. I think the OP is from `.ir`.

src/windows-emulator/syscalls.cpp
```diff
@@ -3413,6 +3413,31 @@ namespace
...
+
+        UNICODE_STRING<EmulatorTraits<Emu64>> ClassName = pustrClassName.read();
+
+        const auto allocatedAddress = c.emu.allocate_memory(0x1000, memory_permission::read_write);
+
+        emulator_object<RTL_ATOM> atom{c.emu, allocatedAddress};
+
+        handle_NtAddAtomEx(c, ClassName.Buffer, ClassName.Length, atom, /*Flags=*/0);
+
+        return atom.read();
```
momo5502, line 3438: this is not exactly what i meant. my intent was rather to extract common parts. like what both functions do is they loop through all existing atoms, to check if it already exists, or they add a new one. so you could extract that into a `uint16_t store_atom(const syscall_context& c, std::u16string atom)` function and just call that one from both syscall handlers

Issue #97, opened 2025-01-22
threads not working correctly


0xeb:
Please use this sample: [threads_1.zip](https://github.com/user-attachments/files/18499465/threads_1.zip) (pass: repro).

If executed, you expect something like:

![Image](https://github.com/user-attachments/assets/2473d9a7-9ba5-4da1-8447-e546c4245b3e)

But if emulated, you get:

![Image](https://github.com/user-attachments/assets/cf2fe3fc-f6ad-4211-8de0-edada761f050)

...which implies that only one yield by thread before the emulation stops.

can you please investigate?

PR #102, opened 2025-01-24
added one more threads test

0xeb:
- use Windows API to test the threads
- (don't mind the memory leaks on failure)

My goal was to repro issue #97 but did not succeed.

The new test works and passes, but the sample binary in issue #97 does not look right under emulation as not enough thread switching takes place when testing it with analyzer.exe; can you repro that issue?

### DIFF
--- a/src/samples/test-sample/test.cpp
+++ b/src/samples/test-sample/test.cpp
@@ -287,6 +287,55 @@ bool test_native_exceptions()
     return test_access_violation_exception() && test_illegal_instruction_exception();
 }
 
+bool test_threads_winapi()
+{
+    struct ctx_t
+    {
+        int iterations;
+        int result;
+    };
+
+    static LPTHREAD_START_ROUTINE thread_proc = [](LPVOID lpParameter) -> DWORD {
+        ctx_t& c = *(ctx_t*)lpParameter;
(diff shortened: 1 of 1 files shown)


momo5502:
is this PR needed then, if it doesn't reproduce the issue?

momo5502:
the issue was a bug in NtWaitForMultipleObjects btw.

src/samples/test-sample/test.cpp
```diff
@@ -287,6 +287,55 @@ bool test_native_exceptions()
...
+        int iterations;
+        int result;
+    };
+
+    static LPTHREAD_START_ROUTINE thread_proc = [](LPVOID lpParameter) -> DWORD {
+        ctx_t& c = *(ctx_t*)lpParameter;
+        c.result = 0;
+        for (int i = 1; i <= c.iterations; i++)
+        {
+            ++c.result;
```
momo5502, line 303: missing a sleep here?

momo5502: or rather a yield. sleep only costs time

0xeb:
should we still use this test at all or I close the issue? @momo5502

momo5502:
i think every test we have is good

PR #141, opened 2025-02-11
added more callbacks

0xeb:


src/windows-emulator/windows_emulator.hpp
```diff
@@ -161,11 +164,18 @@ class windows_emulator
         return this->use_relative_time_;
     }
 
+    emulator_callbacks& callbacks()
```
momo5502: why add a getter again?

momo5502: i removed them as they served no purpose, you can just access the raw object instead

src/common/utils/function.hpp
```diff
@@ -27,19 +27,13 @@ namespace utils
         {
         }
 
-        optional_function& operator=(std::function<Ret(Args...)> f)
+        template <typename F, typename = std::enable_if_t<std::is_invocable_r_v<Ret, F, Args...>>>
```
momo5502:
```suggestion
        template <typename F>
                requires(std::is_invocable_r_v<Ret, F, Args...>)
```

src/windows-emulator/windows_emulator.cpp
```diff
@@ -220,6 +220,10 @@ void windows_emulator::setup_process(const application_settings& app_settings, c
     const auto& emu = this->emu();
     auto& context = this->process;
 
+    mod_manager.on_module_load = [this](mapped_module& mod) { this->callbacks().module_loaded(mod); };
+    mod_manager.on_module_unload = [this](mapped_module& mod) { this->callbacks().module_unloaded(mod); };
+    context.on_create_thread = [this](handle h, emulator_thread& thr) { this->callbacks().thread_created(h, thr); };
+
```
momo5502: doing that adds a level of indirection and thus runtime overhead. might not be that important for those specific events, as they don't happen that often, but what about just passing a reference/pointer to the callbacks object into the mod manager?

momo5502 (changes requested):

0xeb:
sorry, the formatting checks is schizophrenic. the syscalls.cpp is formatted with clang formatter and it keeps complaining.
@momo5502

robert-yates:
> sorry, the formatting checks is schizophrenic. the syscalls.cpp is formatted with clang formatter and it keeps complaining. @momo5502

i have this problem at work also sometimes. different clang format binaries can have slight changes. in this case it appears the workflow is using clang-format v19, check you have the same version

0xeb:
@robert-yates , I use Ctrl-K, Ctrl-D shortcut from within visual studio 2022 (latest update).

what else do I need to use for that? any detailed pointers?

0xeb:
@robert-yates thanks, let's see

image

momo5502:
> @robert-yates , I use Ctrl-K, Ctrl-D shortcut from within visual studio 2022 (latest update).
> 
> what else do I need to use for that? any detailed pointers?

that uses the integrated visual studio formatter which won't use clang-format.
I use this plugin: https://marketplace.visualstudio.com/items?itemName=caphyon.ClangPowerTools
Then you only have to click this button:

image

In the settings you can also configure format-on-save, which is nice.

0xeb:
Thanks for the plugin hint @momo5502. I think that’s the one you told me about before. Anyway, now the formatter works and I could tell it which tool version to use. The PR should be ready now.

momo5502:
> Thanks for the plugin hint @momo5502. I think that’s the one you told me about before. Anyway, now the formatter works and I could tell it which tool version to use. The PR should be ready now.

yep. directx install hung up. sadly that happens sometimes :( need to investigate why. i retriggered the workflow. let's see if it's green :D

src/windows-emulator/windows_emulator.cpp
```diff
@@ -220,6 +220,10 @@ void windows_emulator::setup_process(const application_settings& app_settings, c
     const auto& emu = this->emu();
     auto& context = this->process;
 
+    mod_manager.on_module_load = std::move(callbacks.module_loaded);
+    mod_manager.on_module_unload = std::move(callbacks.module_unloaded);
+    context.on_create_thread = std::move(callbacks.thread_created);
+
```
momo5502, line 226: this should be done in the constructor. setup process may be called multiple times (e.g. in the unit test), which results in moving an already moved object and overwriting the values.

momo5502 (dismissed):

0xeb:
Ping @momo5502 ?

PR #152, opened 2025-03-14
Update README.md

SleekZ:
Fix "prototypy" typo


momo5502:
prototypy was used as an adjective here. seems like it's really wrong. something like `prototypical` maybe fits better

SleekZ:
Ahh gotcha 👍

momo5502:
prototype is still better than prototypy tho 😂

momo5502 (approved):

PR #184, opened 2025-04-09
Improve momo_strip_target and add  unicord-emulator and windows-emulator-test

jlucaso1:
Tested on linux locally in release mode.
Build size before:
105.9 MiB

Build size after:
46.4 MiB

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -323,7 +323,7 @@ function(momo_strip_target target)
     return()
   endif()
 
-  if(MSVC OR MOMO_ENABLE_SANITIZER OR CMAKE_SYSTEM_NAME STREQUAL "iOS" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
+  if(MSVC OR MOMO_ENABLE_SANITIZER OR CMAKE_SYSTEM_NAME STREQUAL "Android")
     return()
   endif()
 
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -323,7 +323,7 @@ function(momo_strip_target target)
     return()
   endif()
 
-  if(MSVC OR MOMO_ENABLE_SANITIZER OR CMAKE_SYSTEM_NAME STREQUAL "Android")
+  if(MSVC OR MOMO_ENABLE_SANITIZER OR CMAKE_SYSTEM_NAME STREQUAL "iOS" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
     return()
   endif()
 
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -323,7 +323,7 @@ function(momo_strip_target target)
     return()
   endif()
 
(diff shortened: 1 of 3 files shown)


cmake/utils.cmake
```diff
@@ -323,29 +323,34 @@ function(momo_strip_target target)
     return()
   endif()
 
-  if(NOT MSVC)
-    # TODO: detect LLVM IR bitcode and abort
+  if(MSVC)
```
momo5502: maybe also return if MOMO_ENABLE_SANITIZER is set, to retain stack traces for sanitizer builds.
```suggestion
  if(MSVC OR MOMO_ENABLE_SANITIZER)
```

cmake/utils.cmake
```diff
@@ -323,29 +323,34 @@ function(momo_strip_target target)
     return()
   endif()
 
-  if(NOT MSVC)
-    # TODO: detect LLVM IR bitcode and abort
+  if(MSVC OR MOMO_ENABLE_SANITIZER)
```
momo5502: probably also disable it for android and ios for now 😂
```suggestion
  if(MSVC OR MOMO_ENABLE_SANITIZER OR CMAKE_SYSTEM_NAME STREQUAL "iOS" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
```

jlucaso1: i think macos too

momo5502: macos should work, it previously did, i think the flags are just different. i might be wrong tho.

cmake/utils.cmake
```diff
@@ -323,29 +323,38 @@ function(momo_strip_target target)
...
-    set(IN_FILE "$<TARGET_FILE:${target}>")
-    set(OUT_FILE "$<TARGET_FILE_DIR:${target}>/$<TARGET_FILE_PREFIX:${target}>$<TARGET_FILE_BASE_NAME:${target}>-unstripped$<TARGET_FILE_SUFFIX:${target}>")
+  set(STRIP_FLAGS "-s")
 
-    add_custom_command(TARGET ${target} POST_BUILD 
-      COMMAND ${CMAKE_COMMAND} -E copy ${IN_FILE} ${OUT_FILE}
-      COMMAND "${STRIP_COMMAND}" ${STRIP_FLAGS} "${IN_FILE}"
-      COMMENT "Strippping ${target}"
-    )
+  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
jlucaso1, line 338: Added this.
I was in doubt between "-x" and "-S"

momo5502: question is, do we still need to ignore ios with this fix? i think it might work

momo5502 (approved):
maybe test ios again? otherwise looks good :D thanks for your contribution

cmake/utils.cmake
jlucaso1: good question. I will try this

momo5502: ios is still broken xD

PR #185, opened 2025-04-09
feat: spoof rdtsc timings

natimerry:
Previously, RDTSC in the VM always returned a constant value of 4, which broke any timing-based operations, or caused detections in heuristics of malware and ANTI-VM tools.

This patch introduces a spoofed rdtsc_fake counter that tracks and adjusts timing deltas to simulate realistic TSC increments. Can be extended to simulate rdtsc timings based on CPU clock speed.

### DIFF
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -40,16 +40,13 @@ namespace utils
         /// TODO: find better solution for ARM and Figure out better CPU base frequency heuristics
         virtual uint64_t timestamp_counter()
         {
-#if defined(_MSC_VER)
-#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86)
-            return __rdtsc(); // 64-bit with MSVC intrinsic
-#endif
-
-#elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__) // If we are using clang or gcc
-            return __rdtsc(); // 64-bit with clang/gcc intrinsic
-#endif
-            int64_t count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86) || defined(__x86_64__) || defined(__i386__) || \
+    defined(__amd64__)
+            return __rdtsc(); // any x86 system will have this instrinsic
+#else
+            const auto count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
             return static_cast<uint64_t>((count * 38LL) / 10LL);
+#endif
         }
     };
 
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -6,6 +6,8 @@
 #if defined(_MSC_VER)
 #include <intrin.h>
 #pragma intrinsic(__rdtsc)
+#elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__)
+#include <x86intrin.h>
 #endif
 
 constexpr auto HUNDRED_NANOSECONDS_IN_ONE_SECOND = 10000000LL;
@@ -44,12 +46,10 @@ namespace utils
 #endif
 
 #elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__) // If we are using clang or gcc
-            unsigned int lo, hi;
-            __asm__ __volatile__("rdtsc" : "=a"(lo), "=d"(hi));
-            return ((uint64_t)hi << 32) | lo;
+            return __rdtsc(); // 64-bit with clang/gcc intrinsic
 #endif
-            return static_cast<uint64_t>(std::chrono::high_resolution_clock::now().time_since_epoch().count()) *
-                   3.8; // should be base cpu frequency here;
+            int64_t count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+            return static_cast<uint64_t>((count * 38LL) / 10LL);
         }
     };
 
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -36,7 +36,7 @@ namespace utils
             return std::chrono::steady_clock::now();
         }
 
-        // Returnds the current timestamp counter value. RDTSC on x86/x64, or just time since epoch for ARM
+        // Returns the current timestamp counter value. RDTSC on x86/x64, or just time since epoch for ARM
         /// TODO: find better solution for ARM and Figure out better CPU base frequency heuristics
         virtual uint64_t timestamp_counter()
         {
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -4,11 +4,9 @@
 
 #include "../platform/platform.hpp"
 #if defined(_MSC_VER)
-#define ARCH_x86
 #include <intrin.h>
 #pragma intrinsic(__rdtsc)
 #elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__)
-#define ARCH_x86
 #include <x86intrin.h>
 #endif
 
@@ -42,10 +40,14 @@ namespace utils
         /// TODO: find better solution for ARM and Figure out better CPU base frequency heuristics
         virtual uint64_t timestamp_counter()
         {
-#if defined(ARCH_x86)
-            return __rdtsc(); 
-#endif // We are using x86, regardless of compiler the instrinsic is the same
+#if defined(_MSC_VER)
+#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86)
+            return __rdtsc(); // 64-bit with MSVC intrinsic
+#endif
 
+#elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__) // If we are using clang or gcc
+            return __rdtsc(); // 64-bit with clang/gcc intrinsic
+#endif
             int64_t count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
             return static_cast<uint64_t>((count * 38LL) / 10LL);
         }
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -3,6 +3,10 @@
 #include <chrono>
 
 #include "../platform/platform.hpp"
+#if defined(_MSC_VER)
+#include <intrin.h>
+#pragma intrinsic(__rdtsc)
+#endif
 
 constexpr auto HUNDRED_NANOSECONDS_IN_ONE_SECOND = 10000000LL;
 constexpr auto EPOCH_DIFFERENCE_1601_TO_1970_SECONDS = 11644473600LL;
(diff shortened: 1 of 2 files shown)


src/windows-emulator/windows_emulator.cpp
```diff
@@ -448,9 +448,24 @@ void windows_emulator::setup_hooks()
     });
 
     this->emu().hook_instruction(x64_hookable_instructions::rdtsc, [&] {
-        const auto instructions = this->executed_instructions_;
-        this->emu().reg(x64_register::rax, instructions & 0xFFFFFFFF);
-        this->emu().reg(x64_register::rdx, (instructions >> 32) & 0xFFFFFFFF);
+        static bool first_rdtsc = 0;
+        static uint_fast64_t rdtsc_prev = 0;
+
+        uint64_t rdtsc_real = __rdtsc();
```
momo5502: this doesn't work on non intel devices.
question is, what exactly does rdtsc return? afaik the cpu ticks? which probably somewhat correlates to the amount of executed instructions?

also, this makes the emulation non-deterministic. i think it might make sense to either extend the clock interface, or to precisely reimplement rdtsc. 

question is, what happens if you just remove the hook and use the default implementation of the underlying emulator?

natimerry:
> Would you mind extending the test.cpp sample with an ANTI-VM technique, so there is a test that actually shows the technique detects the emulator?

Sure, im on linux for a bit now, but I will provide one soon.

src/windows-emulator/windows_emulator.cpp
natimerry: > this doesn't work on non intel devices. question is, what exactly does rdtsc return? afaik the cpu ticks? which probably somewhat correlates to the amount of executed instructions?
> 
> also, this makes the emulation non-deterministic. i think it might make sense to either extend the clock interface, or to precisely reimplement rdtsc.
> 
> question is, what happens if you just remove the hook and use the default implementation of the underlying emulator?

Yeah i just realized the project builds for non x86 systems. This does build on non intel CPUs though considering im using AMD myself

natimerry:
```C++
#include <stdio.h>
#include <windows.h>
#include <intrin.h>
#include <stdlib.h>
#include <stddef.h>

static inline unsigned long long rdtsc() {
    UINT64 ret;
    ret = __rdtsc();
    return ret;
}

int main() {
    int suspicious = 0;
    const int samples = 20;
    UINT64 timings[samples];

    for (int i = 0; i < samples; ++i) {
        UINT64 start = rdtsc();

        Sleep(1);

        UINT64 end = rdtsc();
        timings[i] = end - start;

    }

    // Analyze for constant deltas or suspiciously low / high values
    for (int i = 1; i < samples; ++i) {
        if (timings[i] == timings[i - 1]) suspicious++;
        if (timings[i] < 10000 || timings[i] > 100000000) suspicious++; // too fast
    }

    printf("Suspicious readings: %d out of %d\n", suspicious, samples);

    for (int i = 0; i < samples; ++i)
        printf("Cycle delta %2d: %llu\n", i, timings[i]);

    return 0;
}
```


##  Output without emulation:
```
Suspicious readings: 0 out of 20
Cycle delta  0: 25066206
Cycle delta  1: 61219330
Cycle delta  2: 61030394
Cycle delta  3: 61007366
Cycle delta  4: 62863932
Cycle delta  5: 21880438
Cycle delta  6: 13442842
Cycle delta  7: 32590700
Cycle delta  8: 18656708
Cycle delta  9: 6117848
Cycle delta 10: 43652538
Cycle delta 11: 59116068
Cycle delta 12: 24686054
Cycle delta 13: 60704772
Cycle delta 14: 61635924
Cycle delta 15: 60556496
Cycle delta 16: 57766878
Cycle delta 17: 60610783
Cycle delta 18: 60751740
Cycle delta 19: 61516391
```

## Output through emulator

```
Cycle delta  0: 114
Cycle delta  1: 114
Cycle delta  2: 114
Cycle delta  3: 114
Cycle delta  4: 114
Cycle delta  5: 114
Cycle delta  6: 114
Cycle delta  7: 114
Cycle delta  8: 114
Cycle delta  9: 114
Cycle delta 10: 114
Cycle delta 11: 114
Cycle delta 12: 114
Cycle delta 13: 114
Cycle delta 14: 114
Cycle delta 15: 114
Cycle delta 16: 114
Cycle delta 17: 114
Cycle delta 18: 114
Cycle delta 19: 114
```

natimerry:
There are two potential solutions in my opinion: one is to reimplement rdtsc, and the other is to rely on the host to provide ticks. The latter could involve conditionally compiling based on platform-specific checks, such as `_M_X64` or `_M_ARM64`, to implement spoofing accordingly.

natimerry:
Im not entirely sure how you want me to extend the tests, do you want me to just add the heuristic I sent above to the tests and make sure they pass?

momo5502:
I think we can do it like this:

There is a clock interface that provides an abstraction to the clock: https://github.com/momo5502/emulator/blob/main/src/common/utils/time.hpp#L23
The reason this exists is to allow precise control of the clock from outside. 

What you can do is add a method, something like `virtual uint64_t timestamp_counter()`. The default implementation can return __rdtsc on intel using the _M_X64/x86_64 checks and for every other platform, idk, just maybe return `std::chrono::high_resolution_clock::now().count()` or something similar.

then, you need to provide a different implementation for the tick_clock below, that clock should be 100% deterministic and reproducible, for unittests, so you can maybe just `return this->ticks()` (which equals to the amount of instructions), maybe add some random offset as well or so.

in the rdtsc hook, you can then just call this->clock.timestamp_counter()

momo5502:
> Im not entirely sure how you want me to extend the tests, do you want me to just add the heuristic I sent above to the tests and make sure they pass?

i think extending the tests does not make sense at the moment, as the tests run with a deterministic clock, to prevent any timing/scheduling related inaccuracies. just test locally with your sample for now, if you are happy with the result, feel free to push

natimerry:
Alright, thanks for the details. Helps me get started a lot faster than foraging around the codebase. I will get to it tomorrow, will update you on the details / ask for advise if I need it.

natimerry:
Wow the build really failed over a `3.8` being treated as a double? Let me fix and push that

natimerry:
I updated the code to be not have inline assembly and should build properly now. I will setup the CI/CD on my own repo and try to build it

natimerry:
Ok, I have squashed all the commits that fixed cling tidy and related errors to [0271e39](https://github.com/momo5502/emulator/pull/185/commits/0271e39474ab69fb3ce174fd673155144f32f59f). The github actions succeed for my repo at https://github.com/natimerry/emulator/actions/runs/14389530530/job/40354296615 

This should be good to merge imo unless there are more things to consider

src/windows-emulator/windows_emulator.cpp
```diff
@@ -448,9 +448,19 @@ void windows_emulator::setup_hooks()
     });
 
     this->emu().hook_instruction(x64_hookable_instructions::rdtsc, [&] {
-        const auto instructions = this->executed_instructions_;
-        this->emu().reg(x64_register::rax, instructions & 0xFFFFFFFF);
-        this->emu().reg(x64_register::rdx, (instructions >> 32) & 0xFFFFFFFF);
+        uint64_t ticks = this->clock_->timestamp_counter();
+        static uint64_t fake_ticks = ticks;
+        static uint64_t prev_ticks = 0;
```
momo5502: don't do that. there can be multiple instances of the emulator with different clocks. i don't think there is a need to store the previous ticks.
just write ticks into rax/rdx

src/common/utils/time.hpp
```diff
@@ -29,6 +35,22 @@ namespace utils
...
+        {
+#if defined(_MSC_VER)
+#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86)
+            return __rdtsc(); // 64-bit with MSVC intrinsic
+#endif
+
+#elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__) // If we are using clang or gcc
+            return __rdtsc(); // 64-bit with clang/gcc intrinsic
+#endif
+            int64_t count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
```
momo5502:
```suggestion
            const auto count = std::chrono::high_resolution_clock::now().time_since_epoch().count();
```

src/common/utils/time.hpp
```diff
@@ -29,6 +35,22 @@ namespace utils
...
+        virtual uint64_t timestamp_counter()
+        {
+#if defined(_MSC_VER)
+#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86)
+            return __rdtsc(); // 64-bit with MSVC intrinsic
+#endif
+
+#elif defined(__x86_64__) || defined(__i386__) || defined(__amd64__) // If we are using clang or gcc
+            return __rdtsc(); // 64-bit with clang/gcc intrinsic
+#endif
```
momo5502: aren't those two blocks the same? :D 
```suggestion
#if defined(_M_X64) || defined(_M_AMD64) || defined(_M_IX86) || defined(__x86_64__) || defined(__i386__) || defined(__amd64__)
            return __rdtsc(); // 64-bit with MSVC intrinsic
#endif
```

src/common/utils/time.hpp
natimerry: Alright will do, I initially did use auto but thought it was supposed to be the other way around

src/windows-emulator/windows_emulator.cpp
natimerry: I will do that after some testing. 
If we do need to keep the previous time, would thread_local be an appropriate solution?

src/common/utils/time.hpp
natimerry: I did exactly this and committed earlier, however clang tidy got mad at it, and I can't seem to get it to run locally (it throws error in stuff like qemu and parts of ur code)

momo5502: clang tidy might complain because you might need to put the std::chrono::high... lines in the #else block, as it might get recognized as dead code otherwise

natimerry: Fixed this but i can't get clang-tidy to work on my system to test

src/common/utils/time.hpp
```diff
@@ -29,6 +35,18 @@ namespace utils
         {
             return std::chrono::steady_clock::now();
         }
+
+        // Returnds the current timestamp counter value. RDTSC on x86/x64, or just time since epoch for ARM
```
momo5502:
```suggestion
        // Returns the current timestamp counter value. RDTSC on x86/x64, or just time since epoch for ARM
```

src/windows-emulator/windows_emulator.cpp
```diff
@@ -448,9 +448,9 @@ void windows_emulator::setup_hooks()
     });
 
     this->emu().hook_instruction(x64_hookable_instructions::rdtsc, [&] {
-        const auto instructions = this->executed_instructions_;
-        this->emu().reg(x64_register::rax, instructions & 0xFFFFFFFF);
-        this->emu().reg(x64_register::rdx, (instructions >> 32) & 0xFFFFFFFF);
+        uint64_t ticks = this->clock_->timestamp_counter();
```
momo5502:
```suggestion
        const auto ticks = this->clock_->timestamp_counter();
```

natimerry:
Resolved the formatting error on [daeefb4](https://github.com/momo5502/emulator/commit/daeefb4a7f989ad7eb4de60ac0de464376a1c800)

momo5502 (approved):

momo5502:
everything's green now :3 thanks for the contribution :D

natimerry:
Yaay, looking forward for rdtscp :)

PR #191, opened 2025-04-11
refactor(windows-emulator/syscalls): partial refactor of syscalls.cpp…

RektInator:
refactor(windows-emulator/syscalls): partial refactor of syscalls.cpp, splits syscall implementations over multiple files


momo5502 (approved):
good pr bro. long needed change :D

PR #211, opened 2025-04-19
Add more atom-related syscalls

Igoorx:
While trying to run some executables I noticed handlers for `NtUserRegisterClassExWOW`/`NtUserUnregisterClass` were necessary, so this PR aims to add them. Also, while working on this PR I noticed that atoms should be ref-counted, so this PR also does that.

I found a previous abandoned PR that aimed to add it (https://github.com/momo5502/emulator/pull/101), so I took some of the feedback from that and added a common function in process_context to handle the atoms, I hope that's ok.


momo5502:
Thank you for your PR, from the first glance, it looks good :D
I will review it in detail tomorrow

momo5502 (approved):

momo5502 (approved):

PR #212, opened 2025-04-19
Fix NtOpenSection and add 3 new syscall handlers

Igoorx:
This PR aims to fix an issue with NtOpenSection where it would fail with "Unsupported section" if the RootDirectory was "BASE_NAMED_OBJECTS_DIRECTORY", it also adds handlers for "NtOpenThread" (stub), "NtUserGetProcessWindowStation" (stub) and "NtFlushBuffersFile".

As you can see from the changes, the `NtOpenThread` handler simply returns `STATUS_NOT_SUPPORTED`. I don't know if there's anything better I could do, but this was enough to get around it in the executable I was trying to run.

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -651,7 +651,7 @@ namespace syscalls
 
     NTSTATUS handle_NtUserGetProcessWindowStation()
     {
-        return NULL;
+        return 0;
     }
 }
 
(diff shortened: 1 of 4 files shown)


momo5502:
if you're having trouble with the formatter, i can recommend this visual studio extension: https://clangpowertools.com/

otherwise clang-format will do the job

Igoorx:
Thanks, but it looks like Visual Studio uses clang-format by default:

>By default, Visual Studio runs clang-format.exe in the background and applies formatting as you type. You can also specify to run it only for manually invoked formatting commands Format Document (Ctrl+K, Ctrl+D) or Format Selection (Ctrl+K, Ctrl+F).
(https://learn.microsoft.com/en-us/visualstudio/ide/reference/options-text-editor-c-cpp-formatting?view=vs-2022#configuring-clangformat-options)

I think the problem was that I was using ReSharper's reformat (which I thought was also clang-format, but seems to behave differently?), so hopefully it should work now 🤞

momo5502 (approved):

PR #213, opened 2025-04-20
registry: Implement case-preserving key names and enumeration syscalls

Igoorx:
This PR aims to improve registry handling by keeping the key names casing in the map intact so they can be queried and the caller gets them in their proper casing. It also implements the `NtEnumerateKey` and `NtEnumerateValueKey` syscalls.

### DIFF
--- a/src/common/utils/container.hpp
+++ b/src/common/utils/container.hpp
@@ -18,8 +18,39 @@ namespace utils
         }
     };
 
+    struct insensitive_string_hash
+    {
+        using is_transparent = void;
+
+        size_t operator()(const std::string_view str) const
+        {
+            size_t hash = 0;
+            constexpr std::hash<int> hasher{};
+            for (unsigned char c : str)
+            {
+                hash ^= hasher(std::tolower(c)) + 0x9e3779b9 + (hash << 6) + (hash >> 2);
+            }
+            return hash;
+        }
+    };
+
+    struct insensitive_string_equal
+    {
+        using is_transparent = void;
+
+        bool operator()(const std::string_view lhs, const std::string_view rhs) const
+        {
+            return std::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end(),
+                              [](unsigned char c1, unsigned char c2) { return std::tolower(c1) == std::tolower(c2); });
+        }
+    };
+
     template <typename T>
     using unordered_string_map = std::unordered_map<std::string, T, string_hash, std::equal_to<>>;
+
+    template <typename T>
+    using unordered_insensitive_string_map =
+        std::unordered_map<std::string, T, insensitive_string_hash, insensitive_string_equal>;
 
     using unordered_string_set = std::unordered_set<std::string, string_hash, std::equal_to<>>;
 }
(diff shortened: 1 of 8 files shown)


Igoorx:
Ok, I installed the `Clang Power Tool` extension, but somehow it gives a lot of false positives. I think I have fixed the relevant problems now though.
image

momo5502:
> Ok, I installed the `Clang Power Tools` extension, but somehow it gives a lot of false positives. I think I have fixed the relevant problems now though. image

I only use the clang-format feature from that plugin. I haven't used clang-tidy

src/windows-emulator/registry/registry_manager.cpp
```diff
@@ -114,10 +114,8 @@ std::optional<registry_key> registry_manager::get_key(const utils::path_key& key
     return {std::move(reg_key)};
 }
 
-std::optional<registry_value> registry_manager::get_value(const registry_key& key, std::string name)
+std::optional<registry_value> registry_manager::get_value(const registry_key& key, const std::string& name)
```
momo5502, line 117: you can use std::string_view here

momo5502: you can use std::string_view here:

```suggestion
std::optional<registry_value> registry_manager::get_value(const registry_key& key, const std::string_view name)
```

src/common/utils/container.hpp
```diff
@@ -18,8 +18,39 @@ namespace utils
...
+    {
+        using is_transparent = void;
+
+        size_t operator()(const std::string_view str) const
+        {
+            size_t hash = 0;
+            constexpr std::hash<int> hasher{};
+            for (unsigned char c : str)
+            {
+                hash ^= hasher(std::tolower(c)) + 0x9e3779b9 + (hash << 6) + (hash >> 2);
```
momo5502: you can include "string.hpp" and do this:

```suggestion
            for (const char c : str)
            {
                hash ^= hasher(utils::string::char_to_lower(c)) + 0x9e3779b9 + (hash << 6) + (hash >> 2);
```

that automatically takes care of the int/unsigned char weirdness std::tolower brings

src/common/utils/container.hpp
```diff
@@ -18,8 +18,39 @@ namespace utils
...
+    };
+
+    struct insensitive_string_equal
+    {
+        using is_transparent = void;
+
+        bool operator()(const std::string_view lhs, const std::string_view rhs) const
+        {
+            return std::ranges::equal(
+                lhs, rhs, [](unsigned char c1, unsigned char c2) { return std::tolower(c1) == std::tolower(c2); });
```
momo5502:
```suggestion
                lhs, rhs, [](const char c1, const char c2) { return utils::string::char_to_lower(c1) == utils::string::char_to_lower(c2); });
```

momo5502 (changes requested):
A few minor things, but otherwise looks really good. Thank you for your contributions, I really appreciate it :D

Igoorx:
You're welcome 😄

momo5502 (approved):

momo5502:
> You're welcome 😄, hopefully I will have some more to come, I'm currently trying to track down why `GetTimeZoneInformation` isn't working but it's being quite difficult.

Sounds great, my first guess would be something related to KUSER_SHARED_DATA (kusd_mmio.cpp), but I might be wrong :D

Issue #128, opened 2025-02-07
Support NtAlpcConnectPort/NtAlpcSendWaitReceivePort


acheron2302:
Overview
When I tried to emulate a file with comctl32.dll, the emulator fail when tried to run through comctl32.dll entry point.

Reason
user32.dll call `CsrClientConnectServer` which in turn call "CsrClietnCallServer" and "NtAlpcSendWaitReceivePort" in order to get the information from csrss.exe but "handle_NtAlpcSendWaitReceivePort" hasn't been implemented to support the case.

Version of windows:
windows 11 pro (10.0.22631 build 22631)

Reproduce:
Make change to test.cpp, compile the code and run the command `.\analyzer.exe -e "root" -r "registry" "C:\temp\test.exe"`

```
bool test_tls()
{
    std::atomic_bool kill{false};
    std::atomic_uint32_t successes{0};
    constexpr uint32_t thread_count = 2;

    std::vector<std::thread> ts{};
    kill = false;

    for (size_t i = 0; i < thread_count; ++i)
    {
        ts.emplace_back([&] {
            while (!kill)
            {
                std::this_thread::yield();
            }

            if (tls_var.num == GetCurrentThreadId())
            {
                ++successes;
            }
        });
    }

    LoadLibraryA("d3dcompiler_47.dll");
    LoadLibraryA("dsound.dll");
    LoadLibraryA("comctl32.dll"); // the change is here
    /*LoadLibraryA("d3d9.dll");
    LoadLibraryA("dxgi.dll");
    LoadLibraryA("wlanapi.dll");*/

    kill = true;

    for (auto& t : ts)
    {
        if (t.joinable())
        {
            t.join();
        }
    }

    return successes == thread_count;
}

```

Idea for solutions
Idea 1: just port every argument when the emulator call NtAlpcSendWaitReceivePort to the real NtAlpcSendWaitReceivePort syscall.
Idea 2: Implement a specific case for NtAlpcSendWaitReceivePort of user32.dll and only emulate that case, allocate a memory buffer and copy the content of a fake NtAlpcSendWaitReceivePort to that buffer.
Idea 3 (hardest): dump csrss.exe and fully emulate NtAlpcSendWaitReceivePort.

momo5502:
Can you test if it works for you now?

momo5502:
Loading comctl32.dll should work now, but NtAlpcSendWaitReceivePort does not. That will be needed sooner or later and I'm going to use this issue to track the progress

[...]

PR #223, opened 2025-04-22
Improvements to locale syscalls, fix TimeZone query and add more syscalls

Igoorx:
This PR aims to do the following:
- Add the `NtUserGetAtomName` syscall.
- Add stub handlers for `NtUserSetWindowsHookEx`, `NtUserUnhookWindowsHookEx`, and `NtUserCreateWindowEx`.
- Implement `NtGetNlsSectionPtr` (for Code Pages), add `NtQueryDefaultUILanguage`, and improve `NtQueryInstallUILanguage`.
- Handle `SystemTimeZoneInformation`, `SystemCurrentTimeZoneInformation`, and `SystemDynamicTimeZoneInformation` in `NtQuerySystemInformation`.
- Skip `ThreadAffinityMask` in `NtSetInformationThread`.
- Fix `NtConnectPort` to allow TimeZone queries to work properly. This is probably the most problematic change here because it required messing with the shared section and hardcoding an offset (2504). Initializing the data in this offset seems necessary for calls like `GetTimeZoneInformation` to succeed, so maybe there's no other way around it.

### DIFF
--- a/src/windows-emulator/process_context.hpp
+++ b/src/windows-emulator/process_context.hpp
@@ -80,7 +80,10 @@ struct process_context
     uint64_t current_ip{0};
     uint64_t previous_ip{0};
 
+    uint64_t shared_section_address{0};
+    uint64_t shared_section_size{0};
     uint64_t dbwin_buffer{0};
+    uint64_t dbwin_buffer_size{0};
 
     std::optional<uint64_t> exception_rip{};
     std::optional<NTSTATUS> exit_status{};
--- a/src/windows-emulator/process_context.hpp
+++ b/src/windows-emulator/process_context.hpp
@@ -81,9 +81,9 @@ struct process_context
     uint64_t previous_ip{0};
 
     uint64_t shared_section_address{0};
-    size_t shared_section_size{0};
+    uint64_t shared_section_size{0};
     uint64_t dbwin_buffer{0};
-    size_t dbwin_buffer_size{0};
+    uint64_t dbwin_buffer_size{0};
 
     std::optional<uint64_t> exception_rip{};
     std::optional<NTSTATUS> exit_status{};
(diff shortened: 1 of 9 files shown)


momo5502 (approved):

src/windows-emulator/process_context.hpp
```diff
@@ -79,7 +80,10 @@ struct process_context
     uint64_t current_ip{0};
     uint64_t previous_ip{0};
 
+    uint64_t shared_section_address{0};
+    size_t shared_section_size{0};
```
momo5502: make sure to serialize and deserialize those values.

src/windows-emulator/process_context.hpp
```diff
@@ -79,7 +80,10 @@ struct process_context
...
     uint64_t dbwin_buffer{0};
+    size_t dbwin_buffer_size{0};
```
momo5502: i'd rather prefer this to be a uint64_t and then static cast it to size_t, where the warnings were, as this causes 32 and 64 bit serialized buffers to stay identical

src/windows-emulator/process_context.hpp
momo5502: see process_context.cpp lines 132 and 167

Igoorx:
Oops, I forgot to include the cast in the commit.

momo5502 (approved):
just static_cast in section.cpp 270/277

momo5502:
> Out of curiosity, I tested this change with the emulation roots available in the Github Actions, and it works fine for Windows 2022 and 2025, but for Windows 2019 the offset isn't `2504`, but rather `2672`.

yeah, the entire api port stuff is very fragile and strictly tied to specific versions of the kernel/ntdll. there is #128 to fix that, but i haven't done that yet.

PR #232, opened 2025-04-23
Various fixes, update to create-root.bat and a new syscall

Igoorx:
This PR aims to:
- [Update `create-root.bat`](https://github.com/momo5502/sogen/commit/21fc460db83192464ce0f8773a92ecedb5acbe4f) with more dlls and NLS-related files.
- [Retry loading using normal path when wow6432node path is not found](https://github.com/momo5502/sogen/commit/16e7cac48af4e88c3ee3710a6e3a335e1486adb5), I'm not sure if this is the right thing to do, but I guess Windows does something like this behind the scenes because the saved `wow6432node` doesn't contain all the keys that exist when you look at it in regedit.
- [Add stub for NtAreMappedFilesTheSame and modify NtSetInformationKey to return success](https://github.com/momo5502/sogen/commit/8dfcf2755c3fca1da36a2334ad40c4e55728a048).
- [Simplify the TimeZone query fix](https://github.com/momo5502/sogen/commit/39d40a7f2f588fdba9d3292d1faa1357789b72da), this was necessary because... turns out all the stuff I wrote in NtConnectPort was unnecessary and caused some regressions, all I needed to change was `NtMapViewOfSection`.
- [Fix vm crash when teb.ThreadLocalStoragePointer is null](https://github.com/momo5502/sogen/commit/134b45d1e8107ad1637531c7d084fa22b1143ad3), but I'm not sure if this was a small oversight or something more fundamentally wrong.

### DIFF
--- a/src/windows-emulator/registry/registry_manager.cpp
+++ b/src/windows-emulator/registry/registry_manager.cpp
@@ -105,7 +105,22 @@ std::optional<registry_key> registry_manager::get_key(const utils::path_key& key
         return {std::move(reg_key)};
     }
 
-    const auto* entry = iterator->second->get_sub_key(reg_key.path.get());
+    auto path = reg_key.path.get();
+    const auto* entry = iterator->second->get_sub_key(path);
+
+    if (!entry)
+    {
+        constexpr std::wstring_view wowPrefix = L"wow6432node\\";
+
+        const auto pathStr = path.wstring();
+        if (pathStr.starts_with(wowPrefix))
+        {
+            path = pathStr.substr(wowPrefix.size());
+            reg_key.path = path;
+            entry = iterator->second->get_sub_key(path);
+        }
+    }
+
     if (!entry)
     {
         return std::nullopt;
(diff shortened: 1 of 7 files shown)


src/windows-emulator/registry/registry_manager.cpp
```diff
@@ -105,7 +105,22 @@ std::optional<registry_key> registry_manager::get_key(const utils::path_key& key
...
 
-    const auto* entry = iterator->second->get_sub_key(reg_key.path.get());
+    auto path = reg_key.path.get();
+    const auto* entry = iterator->second->get_sub_key(path);
+
+    if (!entry)
+    {
+        constexpr std::wstring_view wowPrefix = L"wow6432node\\";
+
+        const auto pathStr = path.wstring();
```
momo5502, line 115: isn't that only done for 32 bit applications?

momo5502 (approved):

src/windows-emulator/registry/registry_manager.cpp
Igoorx: I assume you’ve already found the answer, but x64 applications can request that key by using the KEY_WOW64_32KEY flag.

PR #235, opened 2025-04-24
Improvements to NtQueryDirectoryFileEx

Igoorx:
This PR aims to:
- [Stub FileBasicInformation in NtSetInformationFile](https://github.com/momo5502/sogen/commit/6d0ad1dd61454bacdc41ad869d5b82c17540dc9d)
- [Make sure a root `windows_path` have slash at the end](https://github.com/momo5502/sogen/commit/02ed4fbb033135844ec0986fb4ca6fac73b45a3e), this was necessary because `directory_iterator` fails when the path is `C:` (without slash)..
- [Improve NtQueryDirectoryFileEx](https://github.com/momo5502/sogen/commit/f6ec1fc9cca448fc8de031cba1a5dd492e530087) by adding filename filtering support, including more basic file information in the struct, and fixing an issue that prevented the last chunk of files from being enumerated correctly.

### DIFF
--- a/src/windows-emulator/syscalls/file.cpp
+++ b/src/windows-emulator/syscalls/file.cpp
@@ -21,6 +21,11 @@ namespace syscalls
             return STATUS_INVALID_HANDLE;
         }
 
+        if (info_class == FileBasicInformation)
+        {
+            return STATUS_NOT_SUPPORTED;
+        }
+
         if (info_class == FilePositionInformation)
         {
             if (!f->handle)
--- a/src/windows-emulator/syscalls/file.cpp
+++ b/src/windows-emulator/syscalls/file.cpp
@@ -5,6 +5,7 @@
 
 #include <iostream>
 #include <utils/finally.hpp>
+#include <utils/wildcard.hpp>
 
 #include <sys/stat.h>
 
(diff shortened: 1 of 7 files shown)


src/windows-emulator/syscalls/file.cpp
```diff
@@ -99,17 +104,33 @@ namespace syscalls
         }
     }
 
-    std::vector<file_entry> scan_directory(const std::filesystem::path& dir)
+    std::vector<file_entry> scan_directory(const std::filesystem::path& dir, std::u16string file_mask)
```
momo5502: you should be able to use std::u16string_view here

Igoorx:
I added some code from Wine that handles wildcards in file enumeration. Hopefully this will be fine, since Wine uses the LGPL licence.

momo5502:
> I added some code from Wine that handles wildcards in file enumeration. Hopefully this will be fine, since Wine uses the LGPL license.

I'd rather not have lgpl code in this repo, i'm trying to get away from gpl, which is why I added icicle support. I recommend rewriting the code. it doesn't look clean anyways and it's hard to grasp what it really does.

Igoorx:
Alright I think this new one should be better then. I have also included two more things in the PR:
- Add the NtQueryDirectoryFile syscall
- Fix an issue where NtOpenFile wouldn't work if the file was "." with the FILE_FLAG_BACKUP_SEMANTICS flag.

momo5502 (approved):

PR #243, opened 2025-04-28
Load environment variables from registry and other fixes

Igoorx:
This PR aims to:
- [Fix TimeZoneInformation](https://github.com/momo5502/sogen/commit/496fbd3a407e66a191e8cfa73fea77def2c6bdc6) by moving `wDayOfWeek` to the end of the `SYSTEMTIME` structure. This goes against the public definitions of the `SYSTEMTIME` struct but I can confirm that this yields correct values when `GetTimeZoneInformation` is called.
- [Load environment variables from registry](https://github.com/momo5502/sogen/commit/9d5338b1680a7b0e25b2547895c6cac934e02eeb).
- [Miscellaneous fixes](https://github.com/momo5502/sogen/commit/a629f77e31ef4a40ebe8e8b845d398478ecc6bdd), to be more specific:
  - Add `NtOpenProcess` syscall stub.
  - Add logging to `handle_file_enumeration`.
  - Fix `FILE_ATTRIBUTE_NORMAL` being together with `FILE_ATTRIBUTE_DIRECTORY` (`FILE_ATTRIBUTE_NORMAL` means "no attributes are set").
  - Properly handle relative filenames in `NtQueryFullAttributesFile`.
  - Properly handle "trap flag".
  - Handle "DbgPrint" interrupt.

### DIFF
--- a/src/common/platform/kernel_mapped.hpp
+++ b/src/common/platform/kernel_mapped.hpp
@@ -863,28 +863,26 @@ typedef struct _SYSTEM_TIMEOFDAY_INFORMATION64
     ULONGLONG SleepTimeBias;
 } SYSTEM_TIMEOFDAY_INFORMATION64, *PSYSTEM_TIMEOFDAY_INFORMATION64;
 
-#ifndef OS_WINDOWS
-typedef struct _SYSTEMTIME
+typedef struct _SYSTEMTIME64
 {
     WORD wYear;
     WORD wMonth;
-    WORD wDayOfWeek;
     WORD wDay;
     WORD wHour;
     WORD wMinute;
     WORD wSecond;
     WORD wMilliseconds;
-} SYSTEMTIME, *PSYSTEMTIME, *LPSYSTEMTIME;
-#endif
+    WORD wDayOfWeek;
+} SYSTEMTIME64, *PSYSTEMTIME64, *LPSYSTEMTIME64;
 
 typedef struct _SYSTEM_TIMEZONE_INFORMATION
 {
     LONG Bias;
     ARRAY_CONTAINER<char16_t, 32> StandardName;
-    SYSTEMTIME StandardDate;
+    SYSTEMTIME64 StandardDate;
     LONG StandardBias;
     ARRAY_CONTAINER<char16_t, 32> DaylightName;
-    SYSTEMTIME DaylightDate;
+    SYSTEMTIME64 DaylightDate;
     LONG DaylightBias;
 } SYSTEM_TIMEZONE_INFORMATION, *PSYSTEM_TIMEZONE_INFORMATION;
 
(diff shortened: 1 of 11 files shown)


src/windows-emulator/windows_emulator.cpp
```diff
@@ -515,14 +515,23 @@ void windows_emulator::setup_hooks()
...
         {
         case 0:
             dispatch_integer_division_by_zero(this->emu(), this->process);
             return;
         case 1:
-            this->log.print(color::pink, "Singlestep: 0x%" PRIx64 "\n", rip);
+            if ((eflags & 0x100) != 0)
+            {
+                this->log.print(color::pink, "Singlestep (Trap Flag): 0x%" PRIx64 "\n", rip);
+                this->emu().reg(x86_register::eflags, eflags & ~0x100);
```
momo5502, line 529: are you sure removing the trap bit is correct?

src/common/platform/kernel_mapped.hpp
```diff
@@ -863,39 +863,37 @@ typedef struct _SYSTEM_TIMEOFDAY_INFORMATION64
...
     WORD wMonth;
-    WORD wDayOfWeek;
     WORD wDay;
     WORD wHour;
     WORD wMinute;
     WORD wSecond;
     WORD wMilliseconds;
-} SYSTEMTIME, *PSYSTEMTIME, *LPSYSTEMTIME;
-#endif
+    WORD wDayOfWeek;
```
momo5502, line 875: oof, interesting. would you mind adding a test for that? the unit tests use a fake clock, so it should be easy to set that to a fixed value

src/windows-emulator/windows_emulator.cpp
Igoorx: Hm, that's a good question. Actually it looks like it should only be unset during the exception handler.

src/common/platform/kernel_mapped.hpp
Igoorx: This is actually part of the information returned by `NtQuerySystemInformationEx`, so it's already a fixed value.

src/windows-emulator/windows_emulator.cpp
Igoorx: Okay, I did some more in-depth testing, and it looks like removing the trap bit is the correct thing to do. I added a test for it just in case, and it passes on both the emulator and my host machine.

Igoorx:
Looks like icicle didn't like the interrupt test...
But the Windows 2019 fail is expected because that offset discrepancy mentioned in https://github.com/momo5502/sogen/pull/223#issuecomment-2821915290, I guess I will just write in that address as well to fix the issue.

Igoorx:
I guess the icicle failures are unrelated to my changes? I will disable the `Trap Flag` test for icicle for now.

momo5502:
> I guess the icicle failures are unrelated to my changes? I will try to disable the `Trap Flag` test on icicle for now.

well, no, you need to implement trap flag support

momo5502 (approved):

momo5502:
> > I guess the icicle failures are unrelated to my changes? I will try to disable the `Trap Flag` test on icicle for now.
> 
> well, no, you need to implement trap flag support. the way icicle does it is a bit abstract. you have to map the execptions, see handle_exception in icicle.rs

however, i don't think it's super urgent. if you want you can do that in another MR, or if not, the person who encounters the issue can take the time to fix it :P

Igoorx:
>well, no, you need to implement trap flag support. the way icicle does it is a bit abstract. you have to map the execptions, see handle_exception in icicle.rs

Ah, I see. I didn't look too deeply into the internals of the codebase, so I just assumed the icicle backend was completely abstracted away and that the failures were caused by something else.

>however, i don't think it's super urgent. if you want you can do that in another MR, or if not, the person who encounters the issue can take the time to fix it :P

This looks easy enough, even though I don't know much about Rust. I would guess that I just have to map the single-step exception to interrupt 1. I'll give it a try and submit a pull request if I'm successful.

Igoorx:
Never mind, it looks like Icicle just doesn't support the trap flag at all. Implementing it there doesn't look like it would be that difficult if I took the time, but it isn't that important to me, so I might as well leave it for future readers.

PR #246, opened 2025-04-28
Handle listen/accept/send/receive in afd_endpoint

Igoorx:
This PR adds support for listening, accepting, sending, and receiving to afd_endpoint.

A simple HTTP server is currently functional, but I'm likely missing something as I encounter a mapping violation on `select` when performing chunked downloads from the HTTP server.

Also, there are still some things I'm still uncertain about:

- Where the `delayed_ioctl_` event should be reset? It wasn't being reset previously, which caused issues. In this PR, I'm resetting it with every IO operation.
- Is `AFD_LISTEN_RESPONSE_INFO` correct? The struct Windows expects appears to be 20 bytes long, so I had to remove some fields from the structure.
- `pragma pack(push, 1)` seems to be necessary, but it doesn't seem to be used in the other structures, maybe they fit neatly in the alignment?


momo5502:
@RektInator probably makes sense you look at this, as you were working on the same thing

MauriceKayser:
Might contain helpful info: https://www.huntandhackett.com/blog/improving_afd_socket_visibility

Igoorx:
@MauriceKayser Thank you, that had some useful stuff! Also, thanks to it, I found this repository:
https://github.com/A-Normal-User/NtSocket_NtClient_NtServer
Could be useful, but this server doesn't seem to be able to accept connections inside the emulator for some unknown reason.

Igoorx:
I did some more in-depth investigation today and managed to track down the cause(s) of the issues 😄. It seems to be working correctly now. I probably just need to find a better way to handle the completion event and do some touchups before opening the PR for review.

Igoorx:
@momo5502, I think it's all good now! Would you mind checking it out when you get a chance? Thanks!

src/windows-emulator/devices/afd_endpoint.cpp
```diff
@@ -547,6 +624,286 @@ namespace
                 return STATUS_ADDRESS_ALREADY_ASSOCIATED;
             }
 
+            if (c.io_status_block)
+            {
+                IO_STATUS_BLOCK<EmulatorTraits<Emu64>> block{};
+                block.Status = STATUS_SUCCESS;
+                block.Information = 0;
+                c.io_status_block.write(block);
+            }
```
momo5502: i don't think this is needed. this is automatically done here:
https://github.com/momo5502/sogen/blob/main/src/windows-emulator/io_device.hpp#L115

src/windows-emulator/devices/afd_endpoint.cpp
```diff
@@ -547,6 +624,286 @@ namespace
...
+            {
+                return STATUS_INVALID_PARAMETER;
+            }
+
+            if (c.io_status_block)
+            {
+                IO_STATUS_BLOCK<EmulatorTraits<Emu64>> block{};
+                block.Status = STATUS_SUCCESS;
+                block.Information = 0;
+                c.io_status_block.write(block);
```
momo5502: same here and in all the other handlers as well

src/windows-emulator/devices/afd_endpoint.cpp
```diff
@@ -602,7 +959,12 @@ namespace
                     timeout = utils::convert_delay_interval_to_time_point(win_emu.clock(), info.Timeout);
                 }
 
-                this->delay_ioctrl(c, timeout);
+                this->delay_ioctrl(c, {}, timeout, [&](const io_device_context& dc) {
```
momo5502: i would explicitly write out what is being captured here. due to delaying the callback execution, there is high chance of UAFs, so better be explicit in order to prevent these.
```suggestion
                this->delay_ioctrl(c, {}, timeout, [&win_emu](const io_device_context& dc) {
```

momo5502 (approved):

Issue #69, opened 2025-01-11
Add semantic logging


momo5502:
Logging colors are chosen individually for each logged output. But this doesn't make sense. Logging should be based on semantics: Colors should be selected based on how informative the logged information is. So different logging categories like warning, error, but also different info levels are needed: e.g. syscalls being executed is nice to see, but almost never really useful. however certain memory accesses or special syscall events like hiding threads from a debugger need a higher info level.

Mizokuiam:
Great observation! Let me add some thoughts...

VelocityRa:
You guys probably know about it but [spdlog](https://github.com/gabime/spdlog) is a great plug 'n play solution for this, I use it everywhere.

momo5502:
> You guys probably know about it but [spdlog](https://github.com/gabime/spdlog) is a great plug 'n play solution for this, I use it everywhere.

I have used spdlog once a few years ago 😂 It has probably changed a lot since. I will definitely check it out and see if it fits into this project. Thank you for the hint :)

momogflovesbbc:
<img width="560" height="800" alt="Image" src="https://github.com/user-attachments/assets/ab7190de-23fc-447d-8cfa-2d7be79ecabd" />

PR #359, opened 2025-06-05
Move much of the analysis logic from the emulation core into the analyzer

momo5502:
Analysis should not be done in the core. Not everyone using the emulator needs the analysis.
Much of it was moved to the analyzer. Not all, but the rest will be done in a follow up PR.


momo5502:
This somewhat relates to #69

PR #362, opened 2025-06-10
Add temp environment variables to fix GetTempPath(2)W

3fault:
`GetTempPathW` in the following code will fail to retrieve a path from `TMP`, `TEMP`, then `USERPROFILE` environment variables (and `SystemTemp` in `GetTempPath2W`):
```
if ( GetTempPathW(0x104u, &PathName) - 1 > 0x102 || !GetTempFileNameW(&PathName, L"xx", 0, p_LibFileName) )
    return 3;
```

`GetTempFileNameW` will then call `CreateFileW` **many** times, before giving up (see [ReactOS implementation](https://doxygen.reactos.org/da/d76/filename_8c.html#a0963043522ee3e70101d2a764f2153e0)): 

failed

Adding the environment variables to the process context appears to resolve this:

success

### DIFF
--- a/src/windows-emulator/process_context.cpp
+++ b/src/windows-emulator/process_context.cpp
@@ -119,6 +119,10 @@ namespace
         env_map[u"USERNAME"] = u"momo";
         env_map[u"SystemDrive"] = u"C:";
         env_map[u"SystemRoot"] = u"C:\\WINDOWS";
+        env_map[u"SystemTemp"] = u"C:\\Windows\\SystemTemp";
+        env_map[u"TMP"] = u"C:\\Users\\momo\\AppData\\Temp";
+        env_map[u"TEMP"] = u"C:\\Users\\momo\\AppData\\Temp";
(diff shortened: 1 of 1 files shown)


momo5502 (approved):

3fault:
Okay so it seems that the problem runs a little deeper than I had originally thought, `CreateFileW` was originally failing with `ERROR_NOT_SUPPORTED (0x32)` and then converts the error to `ERROR_NOT_SAME_DEVICE (0x11)`, which doesn't break out of the loop. Supplying a root doesn't seem to change much.

My change still fails, just at the `GetFileAttributesW` call earlier on. I'm going to take a look into handle_NtQueryAttributesFile and `handle_NtCreateFile` later on tonight. Not sure if the change sure be reverted in the mean time until I find a fix?

momo5502:
I don't think there is the need to revert you change. It looms useful, even if it doesn't fix your issue

PR #381, opened 2025-06-14
Fix anti-debug detection for NtQueryInformationProcess ProcessDebugObjectHandle

Redcommander:
Small fix for ProcessDebugObjectHandle returning STATUS_SUCCESS instead of STATUS_PORT_NOT_SET, triggering common anti-debugging checks.
Tested on al-khaser, which now passes this part but still fails later on.


momo5502 (approved):

src/windows-emulator/syscalls/process.cpp
```diff
@@ -171,7 +166,12 @@ namespace syscalls
             const emulator_object<handle> info{c.emu, process_information};
             info.write(NULL_HANDLE);
 
-            return STATUS_SUCCESS;
+            if (return_length)
```
momo5502, line 169: now that I see it, this should not have been moved down. otherwise, if the size doesn't match, and STATUS_BUFFER_OVERFLOW is returned above, the caller has no way to know the correct size. I'll fix it on main

Redcommander: It's actually broken again now 😅, [al-khaser](https://github.com/ayoubfaouzi/al-khaser) detects it by calling the function with overlapping buffers, in which case the return length should be written in the buffer instead of the null-handle. (This was fixed in my initial code by moving it down but did indeed not account for that other case, so this should be a better solution overall.)

A simple fix to handle_query_internal would be to check for STATUS_SUCCESS before writing to the buffer.

```cpp
    if (result == STATUS_SUCCESS)
    {
        emu.write_memory(buffer, obj);
    }
```

(Thanks for this amazing project!)

momo5502: oh, yeah, that makes sense. feel free to create a PR :D
otherwise i can maybe fix it later that day

Redcommander: Done! https://github.com/momo5502/sogen/pull/382

PR #406, opened 2025-07-02
Implement guard pages

### LINKED ISSUE #21: Add guard page support

3fault:
Hey, looking for feedback on my guard page implementation. 

I wasn't sure on how to structure the additional windows specific memory flags, since the emulation backends won't like additional guard flag. I opted to create a new `memory_permission_ext` enum to hold the guard flag, and a `nt_memory_permission` struct to wrap the "common" memory permission flags, with the new extended flags. This struct implicitly coerces to the original `memory_permission` to reduce the amount of changes for the PR. 

This however meant that I changed signatures of `map_memory` and `apply_memory_protection` in `memory_interface` to accommodate this new structure, and was an afterthought.

The `map_nt_to_emulator_protection` function might also need some attention now, too. For future reference, windows uses [MiMakeProtectionMask](https://doxygen.reactos.org/d1/d9a/marea_8c.html#adfb66408771a4df77c1056cc2a99ef21) in ntoskrnl to map `PAGE_*` flags to [MM PTE constants](https://reactos.org/wiki/Techwiki:Memory_management_in_the_Windows_XP_kernel).

The test added to the `test-sample` binary seems to be passing.

Fixes #21

### DIFF
--- a/src/emulator/memory_region.hpp
+++ b/src/emulator/memory_region.hpp
@@ -2,7 +2,7 @@
 #include "memory_permission.hpp"
 #include <cstddef>
 
-template<typename PermissionType = memory_permission>
+template <typename PermissionType = memory_permission>
 struct basic_memory_region
 {
     uint64_t start{};


src/windows-emulator/memory_permission_ext.cpp
momo5502, line 1: what's the point of this file?

src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
```diff
@@ -193,9 +193,9 @@ namespace icicle
             icicle_map_mmio(this->emu_, address, size, read_wrapper, ptr, write_wrapper, ptr);
         }
 
-        void map_memory(const uint64_t address, const size_t size, memory_permission permissions) override
+        void map_memory(const uint64_t address, const size_t size, nt_memory_permission permissions) override
```
momo5502: why does backend need to know the nt_memory_permissions? feels like passing memory_permission is just fine here. nt permissions seems to be a platform specific detail that the backend doesn't care about

src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
```diff
@@ -335,9 +335,9 @@ namespace unicorn
                 this->mmio_[address] = std::move(cb);
             }
 
-            void map_memory(const uint64_t address, const size_t size, memory_permission permissions) override
+            void map_memory(const uint64_t address, const size_t size, nt_memory_permission permissions) override
```
momo5502: same here, memory_permission is just fine

src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
```diff
@@ -367,9 +367,9 @@ namespace unicorn
             }
 
             void apply_memory_protection(const uint64_t address, const size_t size,
-                                         memory_permission permissions) override
+                                         nt_memory_permission permissions) override
```
momo5502: same here, memory_permission

src/emulator/memory_interface.hpp
```diff
@@ -2,7 +2,7 @@
 #include <vector>
 #include <functional>
 
-#include "memory_permission.hpp"
+#include "windows-emulator/memory_permission_ext.hpp"
```
momo5502: the emulator has no dependency on the windows-emulator, and it should not need so. the emulator is platform agnostic. as mentioned previously, there should not be any need for that. memory_permission is just fine here

src/emulator/memory_region.hpp
```diff
@@ -1,12 +1,12 @@
 #pragma once
-#include "memory_permission.hpp"
+#include "windows-emulator/memory_permission_ext.hpp"
```
momo5502: same here.

src/windows-emulator/memory_manager.cpp
```diff
@@ -1,8 +1,10 @@
+#include "memory_permission.hpp"
```
momo5502: std_include.hpp should be the first include

src/windows-emulator/memory_manager.cpp
```diff
@@ -1,8 +1,10 @@
...
 #include "std_include.hpp"
 #include "memory_manager.hpp"
 
 #include "memory_region.hpp"
 #include "address_utils.hpp"
+#include "windows-emulator/memory_permission_ext.hpp"
```
momo5502: ../memory_permission_ext

src/windows-emulator/windows_emulator.cpp
```diff
@@ -11,6 +11,7 @@
 #include "apiset/apiset.hpp"
 
 #include "network/static_socket_factory.hpp"
+#include "windows-emulator/memory_permission_ext.hpp"
```
momo5502: ../memory_...

src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
3fault: Yes, originally I was checking for the guard flag in the `map_memory` method, which changed the interface. I have lifted the check into the `allocate_memory` method, which should resolve this.

src/samples/test-sample/test.cpp
```diff
@@ -647,9 +647,99 @@ namespace
...
+    }
+
+    bool test_guard_page_exception()
+    {
+        SYSTEM_INFO sys_info;
+        GetSystemInfo(&sys_info);
+
+        // Allocate a guarded memory region with the length of the system page
+        // size.
+        auto* addr = static_cast<LPBYTE>(
```
momo5502, line 694: should we free the memory? i'm usually totally against memory leaks, but as that's a test, is it ok? i don't know. you decide, if you cba, then it's fine I guess :P

src/windows-emulator/memory_permission_ext.hpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+        return extended;
+    }
+
+    // This just does memberwise equality on each of the members in declaration order
+    bool operator==(nt_memory_permission const&) const = default;
+
+    nt_memory_permission& operator=(memory_permission const& y)
+    {
+        this->common = y;
```
momo5502, line 92: I honestly would expect this operator to reset the extended permissions. Something like this:

```suggestion
        this->common = y;
        this->extended = memory_permission_ext::none;
```

src/windows-emulator/memory_manager.hpp
```diff
@@ -12,13 +12,22 @@ constexpr auto ALLOCATION_GRANULARITY = 0x0000000000010000ULL;
...
+// This maps to the `basic_memory_region` struct defined in
+// emulator\memory_region.hpp
+struct nt_basic_memory_region
+{
+    uint64_t start{};
+    size_t length{};
+    nt_memory_permission permissions{};
+};
+
+struct region_info : nt_basic_memory_region
```
momo5502: i would recommend making the basic_memory_region a template, something like this:

```C++
template<typename PermissionType = memory_permission>
struct basic_memory_region
{
    uint64_t start{};
    size_t length{}; // uint64_t?
    PermissionType permissions{};
};
```

that way, region_info can inherit from that with a custom permission type:
```suggestion
struct region_info : basic_memory_region<nt_memory_permission>
```

momo5502 (approved):

src/windows-emulator/memory_permission_ext.hpp
3fault: Good catch, I really wasn't too sure about this one. I wouldn't necessarily be apposed to just removing the operator to make things a little more explicit

PR #414, opened 2025-07-16
Bug fix, add new syscall, new dll file in collection, update compile flags in MinGW

### LINKED ISSUE #363: BUG: windows-emulator/syscalls/file.cpp:scan_directory will cause exception when handle a directory
I notice this file is update recently, I am not test the updated code, but the source code causing the exception is still there:

```
            files.emplace_back(file_entry{
                .file_path = file.path().filename(),
                .file_size = file.file_size(),
                .is_directory = file.is_directory(),
            });
```

That is, when `file.is_directory == true`, `file.file_size()` will throw exception (on linux environment):

```
Syscall threw an exception: 143 (0x18009fd42) - filesystem error: cannot get file size: Is a directory [/some/directory]
```

Change to this solved the problem:

```
            files.emplace_back(file_entry{
                .file_path = file.path().filename(),
                .file_size = file.is_directory() ? 0 : file.file_size(),
                .is_directory = file.is_directory(),
            });
```

I can issue a pull request (with some other improves) if you wish ;)

Hacksign:
1. FIX #363 
2. Add a new syscall which is `NtUserBuildHwndList`
3. Add 2 new files when construct root filesystem in `src/tools/create-root.bat`
4. Update MinGW compile flags to staticly link the dll
5. Merge upstream (your main branch) updates to the forked repo


deps/googletest
momo5502, line 1: is there any reason all these submodules are being changed?

src/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ add_subdirectory(windows-emulator)
 add_subdirectory(windows-gdb-stub)
 add_subdirectory(backend-selection)
 
+if (MINGW)
+  add_link_options(-static-libstdc++ -static -lwinpthread)
```
momo5502: these flags are all set already: https://github.com/momo5502/sogen/blob/main/cmake/compiler-env.cmake#L67

src/windows-emulator/syscalls.cpp
```diff
@@ -1,3 +1,4 @@
+#include "platform/status.hpp"
```
momo5502: this include is not needed

deps/googletest
Hacksign: I invesgated these changes a little bit, here is the conclusion. Let's focus to the dependency of `googletest` to simplify explanation. 

This problem comes from this commit in my forked repo [1460c49](https://github.com/Hacksign/sogen/commit/1460c49e80933c29872f9e34a9c1aa18e8221dec).

Below is how it comes:

After I synced the upstream on github, I'm trying to merge the the synced code, so I just execute `git pull` and `git merge` on a local repo. Then there comes conflicts .....

I found there is a new dependency `minidump_cpp`, so I just execute `git submodult init` (wrong operation) this lead the `deps/googletest` change to commit `e9092b1`, which is the local repos "current" state binding to "current upstream" (more specificlly: [4c600fd](https://github.com/momo5502/sogen/commit/4c600fd4c4e67f7ae36b221ef5643606341d4cb8) in upstream).

After I resolve the conflict, I pushed the code to my forked repo then issued a pull request. This action lead to the change of these stuffs.

So, I have a question here: is there a proper way to manage `src/deps` ? I've tried add `deps/` to `.gitignores`, no effect : (

momo5502: you probably forgot to do git submodule update --init --recursive and thus staged old submodules.

src/windows-emulator/syscalls.cpp
Hacksign: This is auto added by clangd, it can be removed.

And I have to add "-misc-use-internal-linkage,-misc-unused-parameters" in .clang-tidy to elimate below errors:

* static function error:

截图_选择区域_20250717145824

* unused parameter error:
截图_选择区域_20250717145924

But there is one question, even if I add this contents to `.clang-tidy`:

```
CheckOptions:
    ......
    - key: misc-include-cleaner.UnusedIncludes
      value: 0
```

There still have warnnings about un directly used includes:

截图_选择区域_20250717150247

Any idea to elimate this warnning ?

momo5502: just ignore those warnings
that's clang tidy not being compatible with precompiled headers

PR #416, opened 2025-07-17
Generate an output that the Tenet plugin can read

maskelihileci:
Generates a Tenet tracer output with the -t parameter. The original Tenet tracer plugin is outdated, so I made updates to support our emulator. The forked project below works well.

https://github.com/maskelihileci/tenet/tree/master/plugins_sogen-support


maskelihileci:
I'm not someone who contributes much to projects, so please excuse my inexperience. Only the main.cpp file was modified, and new files tenet_tracer.cpp and tenet_tracer.hpp were added.

Since I didn't make any changes to the other files, I assumed GitHub wouldn't include them and expected it to automatically detect only the modified parts. But it seems I made a mistake, and it looks like the whole files were marked as changed.

Instead of directly accepting the contribution, if you have time, you can compare the source code yourself and add the necessary parts. You can also reject and close this project if needed  I won't take it personally.

momo5502:
I did some cleanup. The changes are in this PR and will be merged in a few minutes: https://github.com/momo5502/sogen/pull/417
I hope I did not break anything. It would be cool if you could verify that :D

momo5502:
A summary of what I did:
I fixed the clang tidy warnings, changed all variables and classes to snake case (and members have underscore at the end instead of m_ prefix), cleaned up the includes a bit, changed the code to use early returns where possible to reduce nesting and I used the scoped_hook to automatically unhook emulator hooks.

PR #418, opened 2025-07-18
Module-based filtering added with -m option

maskelihileci:
Previously, filtering was only applied to .exe files, and the only output in the log file was related to that. However, considering that the user might want to monitor a specific module, I added the ability to filter based on the module as well.

### DIFF
--- a/src/analyzer/object_watching.hpp
+++ b/src/analyzer/object_watching.hpp
@@ -25,7 +25,14 @@ emulator_hook* watch_object(windows_emulator& emu, const std::set<std::string, s
             if (!verbose)
             {
                 static std::unordered_set<uint64_t> logged_addresses{};
-                if (is_main_access && !logged_addresses.insert(address).second)
+
+                bool is_new = false;
+                for (size_t j = 0; j < size; ++j)
+                {
+                    is_new |= logged_addresses.insert(address + j).second;
+                }
+
+                if (!is_new)
                 {
                     return;
                 }


momo5502:
While you are at it fixing things, I recommend that you look into fixing the line endings. Make sure that the line endings of the files are LF, not CRLF. Usually git takes care of that with autocrlf, but aparently net for you 😂 Notepad++ or VS Code can change them

maskelihileci:
I was using VSCode okay, I found it now. I changed CRLF to LF from the bottom right corner, so I guess there won't be any more issues. Actually, I enabled the “hide whitespace changes” feature on GitHub, but it seems there are still problems.

I assumed you were seeing it the same way I did, which is why I created the pull request. Otherwise, I would have fixed that issue so as not to upset you. But now that you've explained the solution, it'll be easier from now on.

image

src/analyzer/tenet_tracer.cpp
```diff
@@ -1,269 +1,279 @@
...
+            continue;
+        }
+
+        char* end_ptr = nullptr;
+        uint64_t address = std::strtoull(line.c_str() + rip_pos + 6, &end_ptr, 16);
+
+        bool is_line_inside = exe_module->is_within(address);
+        if (!is_line_inside && !traced_modules_.empty())
+        {
+            const auto* mod = win_emu_.mod_manager.find_by_address(address);
```
momo5502: you are printing the log at the end. that means there is a chance that some modules have been unloaded. also, there is the possibility that some other module has been loaded again at the same base address.
both of these cases can yield undesired results in the trace. the only fix is probably to move the filtering to the instruction execution hook. which is likely quite expensive. i guess for now, the way it is, is fine. but eventually, this might need to be fixed.

maskelihileci: Exactly, as you said, since it’s quite an expensive operation, I had to do it that way. But that can also be resolved if the added or removed modules are listed somewhere, we can check from there while filtering.

If the module we analyzed was unloaded, we can delete the lines after the one where we got that information. Of course, I’m not sure whether that operation is also expensive or not.

momo5502: i already have logic that prevents certain modules from being unloaded. i think if i enable that for the list of interesting modules, that should automatically fix everything for you, so you should not need to change anything

src/windows-emulator/io_device.cpp
```diff
@@ -49,7 +49,9 @@ std::unique_ptr<io_device> create_device(const std::u16string_view device)
         return create_security_support_provider();
     }
 
-    throw std::runtime_error("Unsupported device: " + u16_to_u8(device));
+    // For any other unhandled device, return a generic dummy device
+    // to prevent the emulator from crashing and allow the program to continue.
+    return std::make_unique<dummy_device>();
```
momo5502: uh, no. there is a reason this is in. dummy device is usually not a working solution. if it is for you, instead add it do the list above. but unknown devices should throw errors

src/windows-emulator/syscalls/file.cpp
```diff
@@ -764,17 +764,14 @@ namespace syscalls
...
+        // If the path after the prefix contains a drive letter (e.g., C:), it's a regular file path, not a device.
+        if (path.size() > 2 && path[1] == ':' && std::isalpha(static_cast<unsigned char>(path[0])))
         {
-            return path;
+            return std::nullopt;
         }
 
-        return std::nullopt;
+        // Otherwise, treat it as a device name.
+        return path;
```
momo5502: same here, not so sure about that. it could still be another mapped path, e.g network drive or whatnot.

src/windows-emulator/io_device.cpp
momo5502: https://github.com/momo5502/sogen/blob/main/src/windows-emulator/io_device.cpp#L21 you can add it here

src/windows-emulator/syscalls/file.cpp
maskelihileci: I had to add this because it doesn't recognize it as a device it thinks it's a file path. It wasn't even returning the "Unsupported device: " result from https://github.com/momo5502/sogen/blob/main/src/windows-emulator/io_device.cpp#L21. If you want, I can send you a sample PE file so you can check it yourself.

src/windows-emulator/io_device.cpp
maskelihileci: Yes, I saw the device structure, and I figured there must be a reason you implemented it that way. Of course, more analysis and testing is needed, but this worked well for me. The error I was getting was more about a running PE file's device creation call terminating with the error "Syscall threw an exception: 55", and the program wouldn't continue executing.

You're absolutely right  for proper results, the emulator needs to emulate those devices as well. But realistically, that's not feasible; there are thousands of devices, and actual communication would be required. That could go beyond the emulator's scope and potentially put the host system at risk.

What I did was simply make it think the device exists. But think of it this way: if that's a weakness in your emulator, a piece of malware could appear to communicate with a fake device and cause the emulator to stop. Of course, your suggestion is valid  you could manually add that device name to the structure. But most users want an emulator that works well out of the box, without having to check the source code first.

maskelihileci: Actually, what I did isn't very reliable either  thinking about it now, malware could still detect it's running in a virtual or emulated environment if the fake device returns `true` when it communicates with it.
So, I can’t really say my approach is safe either. The best option might be to return `"Unsupported device"` and indicate that the device doesn’t exist, allowing the analyzed program to continue executing.
I think this line:

```cpp
throw std::runtime_error("Unsupported device: " + u16_to_u8(device));
```

is what's causing the program to exit.

src/analyzer/tenet_tracer.cpp
maskelihileci: Currently, I'm logging the base address to the tenet_trace.log file using the module, and I'm performing a user-approved automatic rebase with the Tenet plugin. I did something related to unload, but I couldn't find an executable example that unloads the module.

I added add_module_load_callback and add_module_unload_callback to the Windows emulator source code. Without hooking anything (to avoid performance loss), I can track which modules are loaded and unloaded from there.

Once enough testing is completed, I'll upload it to GitHub and you can review it then.

src/windows-emulator/syscalls/file.cpp
momo5502: yes, send the sample

momo5502: but also, that seems to have nothing to do with your PR, module based filtering. maybe a second PR might be better

maskelihileci: [PtControlApp.zip](https://github.com/user-attachments/files/21334098/PtControlApp.zip)

You can check; it will crash when the syscall is called.

```cpp

  memset(hProcess, 0, sizeof(hProcess));
  memset(&SystemInfo, 0, sizeof(SystemInfo));
  lpFilename = (LPWSTR)operator new(0x208u);
  memset(lpFilename, 0, 0x208u);
  GetNativeSystemInfo(&SystemInfo);
  hDevice = CreateFileW(L"\\\\.\\WindowsIntelPtDev", 0x1F01FFu, 0, 0, 3u, 0, 0); -> Crash call
  GetLastError();
  if ( hDevice == (HANDLE)-1LL )
  {
    v14 = sub_140009730(OutBuffer);
    sub_1400010C0((__int64)L"Intel Processor Tracing support for this CPU: ");

```

PR #447, opened 2025-08-10
Added argument dumping framework

CarlTSpeak:
Argument reading missed activity by DLLs pre-executable main, modified.

Code now assesses if is an OS or non-OS PE making the call and dumps arguments if non-OS.

Code checks, sanitises and displays arguments ANSI / unicode / hex depending on contents.

Argument count list generated from arity table - easier to extend.

inline const std::unordered_set<std::string> kOsMods  holds OS module list.

### DIFF
--- a/src/analyzer/analysis.hpp
+++ b/src/analyzer/analysis.hpp
@@ -42,5 +42,110 @@ struct analysis_context
     std::vector<accessed_import> accessed_imports{};
 };
 
+inline void add_functions_with_arity(std::unordered_map<std::string_view, size_t>& map, int arity,
+                                     std::initializer_list<std::string_view> names)
+{
+    for (auto name : names)
+        map[name] = arity;
+}
+
+inline std::unordered_map<std::string_view, size_t> function_argument_count = [] {
+    std::unordered_map<std::string_view, size_t> map;
+
+    add_functions_with_arity(map, 1,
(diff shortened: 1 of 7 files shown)


src/analyzer/analysis.hpp
```diff
@@ -42,5 +42,127 @@ struct analysis_context
...
+                              "LdrLoadDll"});
+
+    add_functions_with_arity(map, 5,
+                             {"CreateDirectoryExA", "CreateDirectoryExW", "CreateDirectoryEx", "NtProtectVirtualMemory",
+                              "NtQueryInformationProcess"});
+
+    add_functions_with_arity(
+        map, 6, {"CreateProcessAsUserA", "CreateProcessAsUserW", "CreateProcessAsUser", "NtQueryValueKey"});
+
+    add_functions_with_arity(map, 11, {"NtCreateUserProcess"});
```
momo5502, line 84: honestly, i think most of these are not needed. almost all of them have a backing syscall that eventually prints the necessary details. e.g. LoadLibrary variants call some section syscall which prints the section details,
NtOpenKey as well, the syscall prints all details. 

i don't think it makes sense to build upon my current implementation. having to manually handle everything here feels off. i think a better way might be to generate function signature mapping based on PDB information. that way things are automated and always correct.

CarlTSpeak: Ack. Will park for now: need to harden against detection to progress an analysis.

PR #480, opened 2025-08-21
Fix exception when NtTerminateThread is called on the active thread.

Redcommander:
Fixes a bug where `perform_context_switch_work` method erases the thread from the threads list and call its destructor but doesn't invalidate the `active_thread` property causing the following switch_to_thread call to call `active_thread->save(emu);` when `active_thread` is already destroyed and thus throws an exception when assigning to `last_registers`.


momo5502 (approved):
Awesome, thank you for the fix :)

PR #481, opened 2025-08-22
Emulator Improvements and New API Support

maskelihileci:
1. Added support for `AllocConsole` call.
2. Added module base support for Tenet tracer; now we can retrieve the base addresses of modules.
3. Unsupported devices were causing the emulator to terminate; now, if a device is not found, it returns false instead just like in a real Windows environment allowing the program to continue execution.
4. Added support for `ProcessWow64Information` and `SystemSupportedProcessorArchitectures`. (Not entirely sure about this one we're returning a fixed value, and it works, but it might not yield the desired result in every scenario.)

Note :  I'm using Git for the first time  I hope the line endings are LF. It says it handles this automatically, so I hope that's true.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
         uses: nick-fields/retry@v3.0.2
         with:
           max_attempts: 5
-          timeout_minutes: 10
+          timeout_minutes: 15
           retry_wait_seconds: 60
           command: |
             wget https://apt.llvm.org/llvm.sh
@@ -252,7 +252,7 @@ jobs:
         if: "${{ matrix.platform == 'Linux x86_64 Clang' }}"
         with:
           max_attempts: 5
-          timeout_minutes: 10
+          timeout_minutes: 15
           retry_wait_seconds: 60
           command: |
             wget https://apt.llvm.org/llvm.sh


src/windows-emulator/windows_emulator.hpp
```diff
@@ -17,10 +17,13 @@ struct io_device;
 
 #define opt_func utils::optional_function
 
-struct emulator_callbacks : module_manager::callbacks, process_context::callbacks
+struct emulator_callbacks : process_context::callbacks
 {
     using continuation = instruction_hook_continuation;
 
+    std::vector<std::function<void(mapped_module& mod)>> on_module_load{};
+    std::vector<std::function<void(mapped_module& mod)>> on_module_unload{};
```
momo5502: just use an opt_func like for the other callbacks. using a vector is unnecessary. it just costs performance overhead while likely being unneeded. usually, one callback is enough, and if not, it can be implemented ontop by the consumer

src/windows-emulator/windows_emulator.hpp
```diff
@@ -155,6 +158,9 @@ class windows_emulator
         return this->executed_instructions_;
     }
 
+    void add_module_load_callback(const std::function<void(mapped_module& mod)>& callback);
+    void add_module_unload_callback(const std::function<void(mapped_module& mod)>& callback);
+
```
momo5502: not needed

src/windows-emulator/syscalls/system.cpp
```diff
@@ -92,6 +92,11 @@ namespace syscalls
     {
         switch (info_class)
         {
+        case SystemSupportedProcessorArchitectures: // The IsWow64Process call was giving the error "Unsupported system info class: B5", so a code change was made to fix it.
```
momo5502: remove the comment pls

src/windows-emulator/syscalls/process.cpp
```diff
@@ -22,6 +22,14 @@ namespace syscalls
         case ProcessEnclaveInformation:
             return STATUS_NOT_SUPPORTED;
 
+        case ProcessWow64Information:
+            return handle_query<ULONG>(c.emu, process_information, process_information_length, return_length,
+                                       [&](ULONG& info) {
+                                           // We are a 64-bit process emulating a 64-bit process,
+                                           // so we are not running under WOW64.
```
momo5502: remove the comment and rather give the constant `3` a name

momo5502:
thank you for your PR :) most of the comments in the code can be removed. will review again once CI is green :D

momo5502:
Just as a hint, this PR contains a lot of independent changes. I don't really mind, but I feel like it makes sense to split these up into different PRs. That makes reviewing easier and will probably get things merged more quickly.

maskelihileci:
> Just as a hint, this PR contains a lot of independent changes. I don't really mind, but I feel like it makes sense to split these up into different PRs. That makes reviewing easier and will probably get things merged more quickly.

I understand, you're right  creating a separate PR and uploading the tests there might be a sensible approach.

PR #558, opened 2025-10-19
Implement ALPC port abstraction and implement DNS resolver port

Igoorx:
This PR introduces an abstraction for ALPC ports to make them easier to manage in the future, and implements the DNS resolver port, at least enough to get host address queries working.  
There's a lot of code in this PR that I'm not very confident about, so don't hesitate on the feedback 😄 

image

### DIFF
--- a/src/windows-emulator/port.cpp
+++ b/src/windows-emulator/port.cpp
@@ -4,7 +4,6 @@
 #include "windows_emulator.hpp"
 #include "ports/api_port.hpp"
 #include "ports/dns_resolver.hpp"
-#include "utils/finally.hpp"
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


momo5502:
Holy canoly, ALPC abstraction is huge. I had that on my todo list for a while, but never got around to actually implementing it :D
That's awesome :)

src/tools/create-root.bat
```diff
@@ -136,6 +136,7 @@ CALL :collect srvcli.dll
 CALL :collect wlanapi.dll
 CALL :collect windowscodecs.dll
 CALL :collect mobilenetworking.dll
+CALL :collect FWPUCLNT.dll
```
momo5502, line 139: must be all lowercase

momo5502: fixed it myself on master 😂

PR #565, opened 2025-10-22
Various fixes and enhancements

Igoorx:
This PR does a number of fixes and enhancements:
- [Added FileFsAttributeInformation fs info class](https://github.com/momo5502/sogen/commit/c0fd8ebf51c158477a9dd97aaaa6dd7a4358548c)
- [Added FileIsRemoteDeviceInformation file info class](https://github.com/momo5502/sogen/commit/c8a18cf61c885911417c998d21d16ae5a55dd075)
- [Added FileIdInformation file info class](https://github.com/momo5502/sogen/pull/565/commits/2c82277b5e7ac7e43edefe60aaf6dfeaa36d620b)
- [Implemented NtUserEnumDisplaySettings syscall](https://github.com/momo5502/sogen/commit/54ebbe27c2171dee66d601647f2755b4a65f1350)
- [Fixed the DISPLAY_DEVICEW value](https://github.com/momo5502/sogen/commit/4fbd65a3020d3ba2c6577e16a2f607d80e0ec945), the old one had `StateFlags` set to 0 but that's incorrect since at least one display needs to be the primary display.
- [Added support for SHA256 in KSecDD](https://github.com/momo5502/sogen/commit/0d06e6822af69083d20d6cc73570eba33a2f4092)
- [Initialized StaticUnicodeString in TEB](https://github.com/momo5502/sogen/commit/1c89c4e3b38ae610d225754a7d07e21577693b87), this was necessary to get `RtlAnsiStringToUnicodeString` working.
- [Added '\??\pipe\' prefix to is_named_pipe_path](https://github.com/momo5502/sogen/commit/d20da120525ac9454bc5a61c6a594e34c55ac84d)
- [Stubbed NtSetSecurityObject](https://github.com/momo5502/sogen/commit/c07f5415284af9d1d273eb216158988e1f81d11f)
- [Don't request 'file_size' for directories](https://github.com/momo5502/sogen/commit/46b97fd91c9f64fd9faa92f25e2492ae4200521b), because this causes an exception in Linux.
- [Optimized module_manager::get_module](https://github.com/momo5502/sogen/commit/83aa3b663a867ce97fcc87480086a045d3766ac7), this seems to increase the emulator's speed by around 10%.

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -912,37 +912,58 @@ namespace syscalls
         return FALSE;
     }
 
-    NTSTATUS handle_NtUserEnumDisplayDevices(const syscall_context& /*c*/,
+    NTSTATUS handle_NtUserEnumDisplayDevices(const syscall_context& c,
                                              const emulator_object<UNICODE_STRING<EmulatorTraits<Emu64>>> str_device, const DWORD dev_num,
                                              const emulator_object<EMU_DISPLAY_DEVICEW> display_device, const DWORD /*flags*/)
     {
-        if (str_device && dev_num != 0)
+        if (!str_device)
         {
-            return STATUS_UNSUCCESSFUL;
-        }
+            if (dev_num > 0)
+            {
+                return STATUS_UNSUCCESSFUL;
+            }
 
-        if (dev_num > 0)
+            display_device.access([&](EMU_DISPLAY_DEVICEW& dev) {
+                dev.StateFlags = 0x5; // DISPLAY_DEVICE_PRIMARY_DEVICE | DISPLAY_DEVICE_ATTACHED_TO_DESKTOP
+                utils::string::copy(dev.DeviceName, u"\\\\.\\DISPLAY1");
+                utils::string::copy(dev.DeviceString, u"Emulated Virtual Adapter");
+                utils::string::copy(dev.DeviceID, u"PCI\\VEN_10DE&DEV_0000&SUBSYS_00000000&REV_A1");
+                utils::string::copy(dev.DeviceKey, u"\\Registry\\Machine\\System\\CurrentControlSet\\Control\\Video\\{00000001-"
+                                                   u"0002-0003-0004-000000000005}\\0000");
+            });
+        }
+        else
         {
-            return STATUS_UNSUCCESSFUL;
-        }
+            const auto dev_name = read_unicode_string(c.emu, str_device);
 
-        display_device.access([&](EMU_DISPLAY_DEVICEW& dev) {
-            dev.StateFlags = 0x80005; // DISPLAY_DEVICE_UNSAFE_MODES_ON | DISPLAY_DEVICE_PRIMARY_DEVICE | DISPLAY_DEVICE_ATTACHED_TO_DESKTOP
-            utils::string::copy(dev.DeviceName, u"\\\\.\\DISPLAY1");
-            utils::string::copy(dev.DeviceID, u"PCI\\VEN_10DE&DEV_0000&SUBSYS_00000000&REV_A1");
-            utils::string::copy(dev.DeviceString, u"Emulator Display");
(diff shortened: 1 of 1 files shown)


src/windows-emulator/module/module_manager.hpp
```diff
@@ -83,9 +83,15 @@ class module_manager
     callbacks* callbacks_{};
 
     module_map modules_{};
+    mutable module_map::iterator last_module_cache_{modules_.end()};
 
     module_map::iterator get_module(const uint64_t address)
     {
+        if (last_module_cache_ != this->modules_.end() && last_module_cache_->second.contains(address))
```
momo5502, line 90: this can break when the last module is unmapped. this leaves a dangling iterator.
why not just fetch the last_module iterator here on the fly?

src/windows-emulator/syscalls/file.cpp
```diff
@@ -182,8 +182,18 @@ namespace syscalls
             return handle_query<FILE_FS_VOLUME_INFORMATION>(c.emu, fs_information, length, io_status_block,
                                                             [&](FILE_FS_VOLUME_INFORMATION&) {});
 
+        case FileFsAttributeInformation:
+            return handle_query<_FILE_FS_ATTRIBUTE_INFORMATION>(
+                c.emu, fs_information, length, io_status_block, [&](_FILE_FS_ATTRIBUTE_INFORMATION& info) {
+                    info.FileSystemAttributes = 0x40006; // FILE_CASE_PRESERVED_NAMES | FILE_UNICODE_ON_DISK | FILE_NAMED_STREAMS
+                    info.MaximumComponentNameLength = 255;
+                    const auto name = u"NTFS"s;
```
momo5502:
```suggestion
                    constexpr auto name = u"NTFS"sv;
```

src/windows-emulator/module/module_manager.hpp
Igoorx: That won't break or leave dangling iterators because the cache is invalidated whenever the map changes:

https://github.com/momo5502/sogen/blob/83aa3b663a867ce97fcc87480086a045d3766ac7/src/windows-emulator/module/module_manager.cpp#L158-L159

https://github.com/momo5502/sogen/blob/83aa3b663a867ce97fcc87480086a045d3766ac7/src/windows-emulator/module/module_manager.cpp#L213-L214

Also, I tried your idea, but fetching on the fly is slower than storing the iterator. The only other alternative I can think of is to store a pointer to the module, but that would also require cache invalidation when the module is unmapped.

momo5502: is it really that slow? i mean it would just be this:
```suggestion
        if (!this->modules_.empty() && this->modules_.rbegin()->second.contains(address))
```

Igoorx: Oh, got it. I think you misunderstood it, the last_module_cache_ stores the last module that was returned by `get_module`, not the literal last module in the module map. But anyway, I just tried this change, it didn't improve the performance.

# rbegin
```cpp
if (!this->modules_.empty() && this->modules_.rbegin()->second.contains(address))//if (last_module_cache_ != this->modules_.end() && last_module_cache_->second.contains(address))
{
    return --this->modules_.rbegin().base();
}
```
```console
[timeit] 6 minutes 5.008s (365.008s), Usr 349.953s, Sys 2.421s (14:41:07-14:47:12)
```

# cache
```cpp
if (last_module_cache_ != this->modules_.end() && last_module_cache_->second.contains(address))
{
    return last_module_cache_;
}
```
```console
[timeit] 4 minutes 53.976s (293.976s), Usr 284.343s, Sys 2.015s (14:48:31-14:53:25)
```

momo5502: i see. aight, let's keep it this way then. sanitizer hopefully reveals any invalid iterators, in case there ever happen to be any :P

momo5502 (approved):

PR #605, opened 2025-11-27
test: add Windows API threads test

0xeb:
## Summary

Adds `test_threads_winapi()` to complement the existing `test_threads()` which uses `std::thread`. This new test verifies Windows threading API emulation using:

- `CreateThread` to spawn multiple threads
- `WaitForMultipleObjects` to synchronize
- `CloseHandle` for cleanup

## Test plan

- [ ] Verify test passes on Windows builds
- [ ] Verify test passes on cross-platform builds

### DIFF
--- a/src/samples/test-sample/test.cpp
+++ b/src/samples/test-sample/test.cpp
@@ -111,8 +111,8 @@ namespace
 
         for (int i = 0; i < thread_count; i++)
         {
-            ctx_t* ctx = ctxs[i] = new ctx_t{5 * (i + 1), i - 2};
-            threads[i] = CreateThread(nullptr, 0, thread_proc, ctx, 0, nullptr);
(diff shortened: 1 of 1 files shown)


src/samples/test-sample/test.cpp
```diff
@@ -87,6 +87,54 @@ namespace
...
+            for (int i = 1; i <= c.iterations; i++)
+            {
+                ++c.result;
+            }
+            return 0;
+        };
+
+        constexpr int thread_count = 5;
+        HANDLE threads[thread_count] = {nullptr};
+        ctx_t* ctxs[thread_count] = {nullptr};
```
momo5502: Why not just do this?
```suggestion
        ctx_t ctxs[thread_count] = {};
```

Then you can get rid of the delete below

momo5502 (approved):

PR #606, opened 2025-11-27
cmake: add SOGEN_STATIC_CRT option for static runtime linking

0xeb:
## Summary

Adds a CMake option to use static CRT (`/MT`) instead of the default dynamic CRT (`/MD`).

**Use case:** Projects that require static linking (e.g., IDA Pro plugins) cannot use the default `/MD` runtime because they must link with `/MT`.

## Changes

- Add `SOGEN_STATIC_CRT` option (default: `OFF`)
- When `ON`, sets `CMAKE_MSVC_RUNTIME_LIBRARY` to static (`/MT` or `/MTd`)
- Also respects parent project's `CMAKE_MSVC_RUNTIME_LIBRARY` if already defined
- Includes documentation warning about heap allocation boundaries

## Backwards Compatibility

- Default behavior unchanged (`/MD`)
- Existing projects unaffected

## Usage

```bash
cmake -B build -DSOGEN_STATIC_CRT=ON
```

Or in parent CMakeLists.txt:
```cmake
set(SOGEN_STATIC_CRT ON CACHE BOOL "")
add_subdirectory(sogen)
```

## Test plan

- [x] Verified default build still uses `/MD`
- [x] Verified `-DSOGEN_STATIC_CRT=ON` produces `/MT` build
- [x] Tested with IDA Pro plugin project (emudbg) - links successfully

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 
 ##########################################
 
-option(MOMO_BUILD_AS_LIBRARY "Configure and Build the sogen as a shared library (without the samples and tests)" ${MOMO_IS_SUBPROJECT})
+option(SOGEN_BUILD_STATIC "Build sogen as static libraries for embedding (e.g., IDA plugins)" ${MOMO_IS_SUBPROJECT})
 
 ##########################################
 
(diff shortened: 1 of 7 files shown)


0xeb:
@momo5502 ping? really, this PR does not change much and is off by default for all users. Except for cases when full static linking is needed , etc.

PR #637, opened 2025-12-24
Fix unicorn hooking

ssvine:
This PR fixes incorrect size when using `uc_hook_add`.

`uc_hook_add`'s memory range is inclusive on both sides. So we need to adjust an ending address to keep correct size.

This is manifested, for example, when using gdb. If a breakpoint is set on a 1-byte instruction, you won't be able to single-step the next instruction because that instruction also triggers the breakpoint, which the debugger doesn't expect.

### DIFF
--- a/src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
+++ b/src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
@@ -551,7 +551,7 @@ namespace unicorn
                 unicorn_hook hook{*this};
 
                 uce(uc_hook_add(*this, hook.make_reference(), UC_HOOK_CODE, wrapper.get_function(), wrapper.get_user_data(), address,
-                                address + size));
+                                calc_end_address(address, size)));
 
                 auto* container = this->create_hook_container();
                 container->add(std::move(wrapper), std::move(hook));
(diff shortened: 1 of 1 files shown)


momo5502 (changes requested):
Thank you for the contribution :D
Please make sure to not overflow if size is std::numerical_limits<uint64_t>::max()

src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
```diff
@@ -551,7 +551,7 @@ namespace unicorn
                 unicorn_hook hook{*this};
 
                 uce(uc_hook_add(*this, hook.make_reference(), UC_HOOK_CODE, wrapper.get_function(), wrapper.get_user_data(), address,
-                                address + size));
+                                this->calc_end_address(address, size)));
```
momo5502:
```suggestion
                                calc_end_address(address, size)));
```

src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
```diff
@@ -582,8 +582,9 @@ namespace unicorn
                 function_wrapper<void, uc_engine*, uc_mem_type, uint64_t, int, int64_t> wrapper(std::move(read_wrapper));
 
                 unicorn_hook hook{*this};
+
                 uce(uc_hook_add(*this, hook.make_reference(), UC_HOOK_MEM_READ_AFTER, wrapper.get_function(), wrapper.get_user_data(),
-                                address, address + size));
+                                address, this->calc_end_address(address, size)));
```
momo5502:
```suggestion
                                address, calc_end_address(address, size)));
```

src/backends/unicorn-emulator/unicorn_x86_64_emulator.cpp
```diff
@@ -606,7 +607,7 @@ namespace unicorn
                 unicorn_hook hook{*this};
 
                 uce(uc_hook_add(*this, hook.make_reference(), UC_HOOK_MEM_WRITE, wrapper.get_function(), wrapper.get_user_data(), address,
-                                address + size));
+                                this->calc_end_address(address, size)));
```
momo5502:
```suggestion
                                calc_end_address(address, size)));
```

momo5502 (changes requested):

PR #640, opened 2025-12-24
Fix thread support

ssvine:
This PR fixes several things related to threads emulation:

1. Support `SameTebFlags.InitialThread`
This flag is needed to support emulation of .net executables (not yet fully supported) that don't have an entry point set in `PE` header. This applies to both `PE32` and `PE64` executables. If `InitialThread` is set the loader substitutes an entry point of the .net executable with `mscoree.dll!_CorExeMain`.

2. Fix static thread local storage for `WOW64`
This fix resolves `shell32.dll` initialization on `WOW64`. This fix also uses correct structure and field names that are obtained from the corresponding `.pdb` files.

3. Fix dynamic thread local storage for `WOW64`

4. Fix setting argument of a `WOW64` thread start proc

5. Fix creating suspended thread and parse create_flags
Currently creating suspended thread doesn't work because `NtCreateThreadEx` handler uses invalid flag `CREATE_SUSPENDED`. This PR fixes that, and moreover it carefully parses create_flags of the `NtCreateThreadEx` call.

6. Fix `FS` and `GS` handling
This PR fixes several problems with `GS` and `FS` segments:

    * Wrong GDT descriptor for selector 0x53
    * Update GDT descriptor for selector 0x53 for a `WOW64` process every context switch like Windows does
    * Set `GS` base when `GS` segment register is updated in 64-bit code (code selector is `0x33`). When `GS` segment register is loaded with correct selector (`0x2b`) `GS` base is set to 0. So, when the code accesses something like `gs:[0]`, a page fault occurs. `KiPageFault` handles this situation and sets correct `GS` base.

Also, take into account that `teb64.ExceptionList` initially contains `teb32` address for `WOW64` process. This is used to setup `FS` base when `wrfsbase` instruction is available. We can enable this instruction using `kusd.ProcessorFeatures.arr[PF_RDWRFSGSBASE_AVAILABLE] = 1;` and this work perfectly with `unicorn` backend. Unfortunately `icicle` backend does not support `wrfsbase`, so I don't enable this instruction by default.

### DIFF
--- a/src/windows-emulator/emulator_thread.cpp
+++ b/src/windows-emulator/emulator_thread.cpp
@@ -19,7 +19,7 @@ namespace
         descriptor |= ((base & 0xFF0000) << 16);                              // Base[23:16]
         descriptor |= (0xF3ULL << 40);                                        // P=1, DPL=3, S=1, Type=3 (Data RW Accessed)
         descriptor |= (static_cast<uint64_t>((limit & 0xF0000) >> 16) << 48); // Limit[19:16]
-        descriptor |= (0x40ULL << 52);                                        // G=0 (byte), D=1 (32-bit), L=0, AVL=0
+        descriptor |= (0x40ULL << 48);                                        // G=0 (byte), D=1 (32-bit), L=0, AVL=0
         descriptor |= ((base & 0xFF000000) << 32);                            // Base[31:24]
 
         // Write the updated descriptor to GDT index 10 (selector 0x53)
@@ -283,14 +283,10 @@ emulator_thread::emulator_thread(memory_manager& memory, const process_context& 
         // Note: CurrentLocale and other fields will be initialized by WOW64 runtime
     });
 
-    // CRITICAL: Setup FS segment (0x53) to point to 32-bit TEB for accurate WOW64 emulation
-    // This mimics what Windows kernel does during NtCreateUserProcess for WOW64 processes
-    // Without this, FS:0 won't correctly access the 32-bit TEB
-    //
-    // NOTE: We cannot use set_segment_base() here because that sets the FS_BASE MSR
-    // which is for 64-bit flat addressing. 32-bit code uses actual GDT-based segmentation
-    // with selector 0x53, so we must modify the GDT entry directly.
-    setup_wow64_fs_segment(memory, teb32_addr);
+    this->teb64->access([&](TEB64& teb_obj) {
+        // teb64.ExceptionList initially points to teb32
+        teb_obj.NtTib.ExceptionList = teb32_addr;
+    });
 
     // Use the allocator to reserve memory for CONTEXT64
     this->wow64_cpu_reserved = emulator_object<WOW64_CPURESERVED>{memory, wow64_cpureserved_base};
@@ -488,4 +484,15 @@ void emulator_thread::setup_registers(x86_64_emulator& emu, const process_contex
     emu.reg(x86_register::rcx, ctx_obj.value());
     emu.reg(x86_register::rdx, context.ntdll_image_base);
     emu.reg(x86_register::rip, context.ldr_initialize_thunk);
+}
+
+void emulator_thread::refresh_execution_context(x86_64_emulator& emu) const
+{
+    (void)emu;
+
+    if (this->teb32.has_value())
+    {
+        // Refresh GDT entry for FS selector on context switch
+        setup_wow64_fs_segment(*this->memory_ptr, this->teb32->value());
+    }
 }
--- a/src/windows-emulator/emulator_thread.hpp
+++ b/src/windows-emulator/emulator_thread.hpp
@@ -122,6 +122,7 @@ class emulator_thread : public ref_counted_object
     void restore(x86_64_emulator& emu) const
     {
         emu.restore_registers(this->last_registers);
+        this->refresh_execution_context(emu);
     }
 
     void setup_if_necessary(x86_64_emulator& emu, const process_context& context)
@@ -242,6 +243,7 @@ class emulator_thread : public ref_counted_object
 
   private:
     void setup_registers(x86_64_emulator& emu, const process_context& context) const;
+    void refresh_execution_context(x86_64_emulator& emu) const;
 
     void release()
     {
--- a/src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
+++ b/src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
@@ -178,6 +178,21 @@ namespace icicle
             }
         }
 
+        pointer_type get_segment_base(const x86_register base) override
+        {
+            switch (base)
+            {
+            case x86_register::fs:
+            case x86_register::fs_base:
+                return this->reg(x86_register::fs_base);
+            case x86_register::gs:
+            case x86_register::gs_base:
+                return this->reg(x86_register::gs_base);
+            default:
+                return 0;
+            }
+        }
+
         size_t write_raw_register(const int reg, const void* value, const size_t size) override
         {
             return icicle_write_register(this->emu_, reg, value, size);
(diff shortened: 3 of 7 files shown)


src/windows-emulator/emulator_thread.hpp
```diff
@@ -121,6 +122,14 @@ class emulator_thread : public ref_counted_object
     void restore(x86_64_emulator& emu) const
     {
         emu.restore_registers(this->last_registers);
+
+        // For WOW64, also set FS segment base to point to 32-bit TEB
+        // Windows kernel sets both GDT descriptor and FS_BASE MSR during thread creation
+        if (this->teb32.has_value())
```
momo5502: I don't think that's necessary, saving and restoring of the registers should already do that

ssvine: If I setup fs and gs registers once (like it was before the fix), and then comment this line, I get

```
Mapping violation: 0x0 (4) - r-- at 0x40103e (threads-2.exe)
```

ssvine: Test application (x86):

```c++
#include <windows.h>
#include <atomic>
#include <iostream>

struct Task {
    int task_id;
};

Task g_tasks[] = { {-1}, {101}, {102}, {103} };
std::atomic<int> g_task_index = 0;
thread_local int g_thread_task_id = g_tasks[g_task_index++].task_id;

DWORD WINAPI WorkerThread(LPVOID) {
    std::cout << "Working on task " << g_thread_task_id << "\n";
    return 0;
}

using NtCreateThreadEx_t = NTSTATUS(WINAPI*)(
    _Out_ PHANDLE ThreadHandle,
    _In_ ACCESS_MASK DesiredAccess,
    _In_opt_ LPVOID ObjectAttributes,  // POBJECT_ATTRIBUTES
    _In_ HANDLE ProcessHandle,
    _In_ PVOID StartRoutine,  // PUSER_THREAD_START_ROUTINE
    _In_opt_ PVOID Argument,
    _In_ ULONG CreateFlags,  // THREAD_CREATE_FLAGS_*
    _In_ SIZE_T ZeroBits,
    _In_ SIZE_T StackSize,
    _In_ SIZE_T MaximumStackSize,
    _In_opt_ LPVOID AttributeList  // PPS_ATTRIBUTE_LIST
    );

int main() {
    //CreateThread(nullptr, 0, [](LPVOID) -> DWORD { return 0; }, nullptr, 0, nullptr);

    NtCreateThreadEx_t pNtCreateThreadEx =
        (NtCreateThreadEx_t)GetProcAddress(GetModuleHandle(L"ntdll.dll"), "NtCreateThreadEx");
    const ULONG SkipThreadAttach = 0x2;
    HANDLE hThread;
    NTSTATUS result = pNtCreateThreadEx(&hThread, THREAD_ALL_ACCESS, nullptr,
        GetCurrentProcess(), (PTHREAD_START_ROUTINE)([](LPVOID) -> DWORD {
            std::cout << "Thread created\n";
            return 0;
            }), nullptr,
        SkipThreadAttach, 0, 0, 0, nullptr);

    // Run tasks.
    for (int i = 0; i < 3; i++) {
        CreateThread(nullptr, 0, WorkerThread, nullptr, 0, nullptr);
        Sleep(1000);
    }
    std::cout << "Exiting\n";
}
```

This can also be used to test TLS and creation flags.

Output should be:
```
Thread created
Working on task 101
Working on task 102
Working on task 103
Exiting
```

momo5502: Interesting. I hope you don't mind if I delay the review of this PR, it's slightly bigger and I want to review it thoroughly and I'm not at home during the christmas holidays.
I really appreciate all your recent contributions though, thank your very much :)

ssvine: No problem)

src/windows-emulator/emulator_thread.cpp
```diff
@@ -152,6 +155,10 @@ emulator_thread::emulator_thread(memory_manager& memory, const process_context&
             teb_obj.NtTib.Self = this->teb64->value();
             teb_obj.CurrentLocale = 0x409;
             teb_obj.ProcessEnvironmentBlock = context.peb64.value();
+            teb_obj.SameTebFlags.InitialThread = initial_thread;
+            teb_obj.SameTebFlags.SkipThreadAttach = static_cast<bool>(create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH);
```
momo5502: you don't need the static cast :D

```suggestion
            teb_obj.SameTebFlags.SkipThreadAttach = create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH;
```

src/windows-emulator/emulator_thread.cpp
```diff
@@ -209,6 +216,10 @@ emulator_thread::emulator_thread(memory_manager& memory, const process_context&
         teb_obj.CurrentLocale = 0x409;
 
         teb_obj.ProcessEnvironmentBlock = context.peb64.value();
+        teb_obj.SameTebFlags.InitialThread = initial_thread;
+        teb_obj.SameTebFlags.SkipThreadAttach = static_cast<bool>(create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH);
```
momo5502: same here

src/windows-emulator/emulator_thread.cpp
```diff
@@ -264,6 +275,10 @@ emulator_thread::emulator_thread(memory_manager& memory, const process_context&
         }
 
         teb32_obj.WowTebOffset = -0x2000;
+        teb32_obj.InitialThread = initial_thread;
+        teb32_obj.SkipThreadAttach = static_cast<bool>(create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH);
```
momo5502: same here

src/windows-emulator/emulator_thread.hpp
```diff
@@ -86,6 +86,7 @@ class emulator_thread : public ref_counted_object
     bool await_any{false};
     bool waiting_for_alert{false};
     bool alerted{false};
+    uint32_t create_flags{0};
```
momo5502, line 89: please make sure this flag is serialized/deserialized

here: https://github.com/momo5502/sogen/blob/main/src/windows-emulator/emulator_thread.hpp#L167
and here: https://github.com/momo5502/sogen/blob/main/src/windows-emulator/emulator_thread.hpp#L212

momo5502 (commented):
PR looks really good, a few small comments, but nothing big. thank you very much for your contribution :)

src/windows-emulator/emulator_thread.cpp
ssvine: It's very easy to overlook here...

`SkipThreadAttach` is an unsigned bit-field of size 1-bit (it holds values from range [0,1]). So, the value to store in a bit-field `SkipThreadAttach` is `result % 2`.
At the same time `create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH` equals `0` or `THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH` (which is 2).
But `(0 % 2) == (2 % 2) == 0`, so without casting to bool this will store 0 to the field.

So, casting to bool is needed for all the fields

src/windows-emulator/emulator_thread.hpp
ssvine: Yes, I missed it. I'll update it tomorrow

src/windows-emulator/emulator_thread.cpp
momo5502: interesting, good catch, but then i would even be more explicit, afaik the standard doesn't specify the binary representation of bool.

```suggestion
            teb_obj.SameTebFlags.SkipThreadAttach = (create_flags & THREAD_CREATE_FLAGS_SKIP_THREAD_ATTACH) ? 1 : 0;
```

that's also shorter and more readable

momo5502:
PR looks really good, thank you for that. I love that you could remove many workarounds :)

PR #643, opened 2025-12-27
VMP program simulation patches

66hh:
This PR fixes some issues with VMP program emulation. It has successfully allowed programs protected by VMP 3.5 to run normally, and it modifies the following functions:

1. Correct implementation of ProcessInstrumentationCallback
2. Add missing path check in handle_NtCreateFile
3. Fix the check in handle_NtOpenSection
4. Fix the behavioral differences between the emulator and the kernel when the return value is an invalid pointer
5. Add printing of the content of the NtRaiseHardError message box
6. Bypass NtClose detection for VMP

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -131,9 +131,9 @@ namespace syscalls
     NTSTATUS handle_NtReadVirtualMemory(const syscall_context& c, handle process_handle, emulator_pointer base_address,
                                         emulator_pointer buffer, ULONG number_of_bytes_to_read,
                                         emulator_object<ULONG> number_of_bytes_read);
-    NTSTATUS handle_NtWriteVirtualMemory(const syscall_context& c, const handle process_handle, const emulator_pointer base_address,
-                                         const emulator_pointer buffer, const ULONG number_of_bytes_to_write,
-                                         const emulator_object<ULONG> number_of_bytes_write);
+    NTSTATUS handle_NtWriteVirtualMemory(const syscall_context& c, handle process_handle, emulator_pointer base_address,
+                                         emulator_pointer buffer, ULONG number_of_bytes_to_write,
+                                         emulator_object<ULONG> number_of_bytes_write);
     NTSTATUS handle_NtSetInformationVirtualMemory();
     BOOL handle_NtLockVirtualMemory();
 


src/common/platform/process.hpp
```diff
@@ -1124,4 +1124,10 @@ struct PROCESS_PRIORITY_CLASS
     UCHAR PriorityClass;
 };
 
+struct PROCESS_INSTRUMENTATION_CALLBACK_INFORMATION
+{
+    ULONG Version;
+    ULONG Reserved;
+    PVOID Callback;
```
momo5502: this needs to be uint64_t, otherwise it will break on 32 bit hosts

66hh: Okay, I have fixed this issue.

src/windows-emulator/syscalls/object.cpp
```diff
@@ -6,7 +6,14 @@ namespace syscalls
 {
     NTSTATUS handle_NtClose(const syscall_context& c, const handle h)
     {
+
         const auto value = h.value;
+
+        if (h.h == 0xDEADC0DE)
```
momo5502, line 11: why is that even needed? it should already result in STATUS_INVALID_HANDLE, as this won't resolve to a valid handle store below?

src/windows-emulator/syscall_dispatcher.cpp
```diff
@@ -101,6 +101,19 @@ void syscall_dispatcher::dispatch(windows_emulator& win_emu)
...
+            uint64_t rip_old = emu.reg<uint64_t>(x86_register::rip);
+
+            // The increase in RIP caused by executing the syscall here has not yet occurred.
+            // If RIP is set directly, it will lead to an incorrect address, so the length of
+            // the syscall instruction needs to be subtracted.
+            emu.reg<uint64_t>(x86_register::rip, context.instrumentation_callback - 2);
+
+            emu.reg<uint64_t>(x86_register::r10, rip_old);
+        }
+
```
momo5502: extract that into a function with a meaningful name pls

src/windows-emulator/syscall_utils.hpp
```diff
@@ -232,7 +232,17 @@ NTSTATUS handle_query(x86_64_emulator& emu, const uint64_t buffer, const uint32_
     const auto length_setter = [&](const size_t required_size) {
         if (return_length)
         {
-            return_length.write(static_cast<LengthType>(required_size));
+            try
+            {
+                // VMProtect is trying to pass an incorrect return address.
+                // This can run on the original version of Windows, but in
+                // the emulator, this will cause an exception.
+                return_length.write(static_cast<LengthType>(required_size));
```
momo5502: might make sense to introduce a try_write here on the emulator_object

src/windows-emulator/syscalls/object.cpp
66hh: Because this value could incorrectly be treated as a fake handle and enter the following branch, I had to specifically filter this handle.
if (value.is_pseudo)
{
    return STATUS_SUCCESS;
}

66hh: > Because this value could incorrectly be treated as a fake handle and enter the following branch, I had to specifically filter this handle. if (value.is_pseudo) { return STATUS_SUCCESS; }

I verified this through a simple piece of code.
	uint64_t val = 0xDEADC0DE;

	handle_value* val2 = (handle_value*)&val;

	std::cout << val2->is_pseudo;
This will output 1

66hh:
In addition, I discovered some new issues while testing certain samples, and it seems that address alignment has not been properly implemented. I think I need to convert this PR to a draft first.

src/windows-emulator/syscalls/object.cpp
66hh: > why is that even needed? it should already result in STATUS_INVALID_HANDLE, as this won't resolve to a valid handle store below?

Should I try modifying handle_value to fix the conflict?

src/windows-emulator/syscall_utils.hpp
66hh: Resolved

src/windows-emulator/syscall_dispatcher.cpp
66hh: Resolved

src/windows-emulator/syscalls/object.cpp
momo5502: nah, it's fine then

src/windows-emulator/syscalls/memory.cpp
```diff
@@ -280,23 +292,57 @@ namespace syscalls
...
+        number_of_bytes_read.try_write(0);
 
         if (process_handle != CURRENT_PROCESS)
         {
             return STATUS_NOT_SUPPORTED;
         }
 
-        std::vector<uint8_t> memory{};
-        memory.resize(number_of_bytes_to_read);
+        uint8_t* memory = new uint8_t[number_of_bytes_to_read];
```
momo5502:
```suggestion
        std::vector<uint8_t> memory(number_of_bytes_to_read, 0);
```

that way, you can get rid of the delete below. new/delete is evil 😂

src/windows-emulator/syscalls/memory.cpp
```diff
@@ -280,23 +292,57 @@ namespace syscalls
...
+                                         const emulator_object<ULONG> number_of_bytes_write)
+    {
+        number_of_bytes_write.try_write(0);
+
+        if (process_handle != CURRENT_PROCESS)
+        {
+            return STATUS_NOT_SUPPORTED;
+        }
+
+        uint8_t* memory = new uint8_t[number_of_bytes_to_write];
```
momo5502: same here

66hh:
Additionally, I found another issue: if VMP enables memory protection, this patch is still not enough to make it run. If I can figure out what causes this problem, I might open a new PR later to address it.
image
It seems that the code in ntdll does not meet VMP's expectations?

momo5502 (approved):

66hh:
> Before merging, CI obv. has to pass

I think I have fixed the problem. Can we test it again?

66hh:
The format should be correct this time as well.

PR #682, opened 2026-01-04
Set PEB64/32 HeapSegmentReserve & HeapSegmentCommit from Executable.

ahm3dgg:


ahm3dgg:
Okay Hopefully that will work fine

Issue #383, opened 2025-06-16
Issues regarding sections


Redcommander:
Not sure how to best describe this problem, which is split into a few issues I've noticed

1) NtOpenSection does not work if the filename (ObjectName) starts with \KnownDlls\ (such as \KnownDlls\ntdll.dll)
2) When calling NtMapViewOfSection and the section cannot be mapped to the address specified in the image file STATUS_IMAGE_NOT_AT_BASE should be returned (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55)
3) When a view of ntdll is mapped, it should not share the same address as the originally loaded ntdll view (as far as I know, I might be wrong here)

I'm not sure how these changes should be best implemented (if at all) into sogen which is why I created an issue instead of a PR.

momogflovesbbc:
<img width="998" height="1321" alt="Image" src="https://github.com/user-attachments/assets/6ec0649c-9379-4d35-b331-e5f5f9988019" />

ahm3dgg:
there is a little mistake in `create_known_dlls_section_objects` should have maintained two different vectors KnownDlls and KnownDlls32, since there is a few differences between the two.

PR #685, opened 2026-01-06
Add Support for KnownDLLs

ahm3dgg:
This is my attempt at solving #383

- For KnownDLLs I create the section objects at process setup, later in NtOpenSection, I check if the root directory is KnownDLL or if it the path starts with "\\KnownDll", and create a handle to the section object.

- For handling STATUS_IMAGE_NOT_AT_BASE, I maintain a map of module filepath -> module load count, at first load the module will get either the preferred image or the whatever the memory manager gives, and if load count is greater than 1, I return STATUS_IMAGE_NOT_AT_BASE, and for that I had to allow loading multiple copies of the same DLL.

I have also tried the user test, and it does pass.
I am unsure if its the best way though, so if there anything I have done wrong, or can be suggested I will be happy to listen.


momo5502:
> > > but the problem is that this list, is not complete
> > 
> > 
> > Found this topic [What are Known DLLs anyway?](https://web.archive.org/web/20090619141433/http://blogs.msdn.com/larryosterman/archive/2004/07/19/187752.aspx)
> > There is an interesting observation:
> > If you compare the `HKLM\System\CCS\Control\Session Manager\KnownDLLs` registry key with the sections under `\KnownDlls` (using a viewer like `winobj`, you’ll notice that the `\KnownDlls` object container always has more entries in it than the registry key. This is because the `\KnownDlls` sections are computed as the transitive closure of the DLLs listed in `KnownDLLs`. So if a DLL’s listed in `KnownDLLs`, all of the DLL’s that are statically linked with the DLL are ALSO listed in the `\KnownDlls` section.
> > P.S. I'm not the maintainer anyway. It's up to @momo5502 to decide :)
> 
> Yes exactly, and yes I would like to hear from @momo5502 and his opinion on my approach in general

If I get it correctly, the list of known dlls consists of registry entries + their transitive dependencies. Should be easy to build that list, shouldn't it? we have code to parse the registry and we have code to parse PE files and their imports (iirc), so we can just compute that list, can't we? However, if that's too complicated, I'm also fine with smaller milestones, hardcoded lists are fine, but also parsing the registry is fine. I'm happy about every step in the right direction.

PR #688, opened 2026-01-06
process: allocate valid GDI state

redthing1:


momo5502:
> I will fix the test issues. Also I have more GDI syscall implementations in progress, I can roll them into this PR or split lmk if you have a preference

I honestly prefer separate PRs, makes it easier to review :P

PR #691, opened 2026-01-06
Add support for user callbacks

Igoorx:
This PR is my attempt to add support for user callbacks in the emulator. User callbacks allow the emulator to call guest callbacks from syscalls, and when the callback finishes running, control returns to the syscall through the completion method. I've also added a test and implemented the NtUserEnumDisplayMonitors syscall.

One thing to note is that this implementation isn't faithful to how the Windows kernel does it, since the kernel uses the KernelCallbackTable and the `ntdll!KiUserCallbackDispatch` method, and this implementation currently just calls the callback directly.

### DIFF
--- a/src/windows-emulator/emulator_utils.hpp
+++ b/src/windows-emulator/emulator_utils.hpp
@@ -408,8 +408,8 @@ inline std::u16string read_unicode_string(emulator& emu, const uint64_t uc_strin
     return read_unicode_string(emu, emulator_object<UNICODE_STRING<EmulatorTraits<Emu64>>>{emu, uc_string});
 }
 
-/// Retrieves function arguments from registers or stack memory. This function assumes the caller has already allocated
-/// stack space and that RSP currently points to the return address.
+/// Retrieves function arguments from registers or stack memory. This function assumes the stack pointer currently points to the
+/// return address.
 inline uint64_t get_function_argument(x86_64_emulator& emu, const size_t index, const bool is_syscall = false)
 {
     bool use_32bit_stack = false;
--- a/src/windows-emulator/process_context.cpp
+++ b/src/windows-emulator/process_context.cpp
@@ -454,14 +454,16 @@ void process_context::setup_callback_hook(windows_emulator& win_emu, memory_mana
     uint64_t sentinel_addr = this->callback_sentinel_addr;
     if (!sentinel_addr)
     {
-        auto sentinel_obj = this->base_allocator.reserve_page_aligned<uint8_t>();
+        using sentinel_type = std::array<uint8_t, 2>;
+        constexpr sentinel_type sentinel_opcodes{0x90, 0xC3}; // NOP, RET
+
+        auto sentinel_obj = this->base_allocator.reserve_page_aligned<sentinel_type>();
         sentinel_addr = sentinel_obj.value();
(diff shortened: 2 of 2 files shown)


src/windows-emulator/emulator_utils.hpp
```diff
@@ -440,3 +440,48 @@ inline uint64_t get_function_argument(x86_64_emulator& emu, const size_t index,
...
+        emu.reg(x86_register::rdx, value);
+        break;
+    case 2:
+        emu.reg(x86_register::r8, value);
+        break;
+    case 3:
+        emu.reg(x86_register::r9, value);
+        break;
+    default:
+        emu.write_stack(index + 1, value);
```
momo5502, line 481: should this adjust rsp?

src/windows-emulator/user_callback_dispatch.hpp
```diff
@@ -0,0 +1,46 @@
...
+    // Push the return address onto the stack (Simulating CALL)
+    stack_ptr -= sizeof(emulator_pointer);
+    c.emu.write_memory(stack_ptr, &c.proc.callback_sentinel_addr, sizeof(c.proc.callback_sentinel_addr));
+
+    c.proc.active_thread->callback_stack.push_back(frame);
+
+    c.emu.reg(x86_register::rsp, stack_ptr);
+
+    size_t index = 0;
+    (set_function_argument(c.emu, index++, static_cast<uint64_t>(args)), ...);
```
momo5502: if argument count exceeds the amount of arguments in registers and requires using the stack, then the order must be in reverse (last argument pushed first), if they are pushed, unless you want it to write using the index. then consumers of the set_function_argument api would need to know that they have to adjust rsp beforehand.

src/windows-emulator/emulator_utils.hpp
Igoorx: It currently doesn't adjust the RSP because the RSP is pre-adjusted beforehand

src/windows-emulator/user_callback_dispatch.hpp
Igoorx: Yes, my intention was to do the latter. I'll try to make that more explicit, but I can also change it to push the arguments instead if you think that would be better.

src/windows-emulator/user_callback_dispatch.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "syscall_utils.hpp"
+
+// TODO: Here we are calling guest functions directly, but this is not how it works in the real Windows kernel.
+//       In the real implementation, the kernel invokes ntdll!KiUserCallbackDispatcher and passes a callback
+//       index that refers to an entry in PEB->KernelCallbackTable. The dispatcher then looks up the function
+//       pointer in that table and invokes the corresponding user-mode callback.
```
momo5502, line 8: just out of curiosity, why aren't we doing it the intended way?

Igoorx: Because, from what I understand, we'd need to know the correct callback index in the KernelCallbackTable for each syscall, and that index changes with every Windows build. So we'd probably have to do something similar to what you did for syscalls. Since that wouldn't be exactly trivial, and it already works fine from what I can tell, I just didn't want to do that work right now.

PR #694, opened 2026-01-07
Support for KnownDLLs

ahm3dgg:
This is my attempt at solving #383

- For KnownDLLs I create the section objects at process setup, later in NtOpenSection, I check if the root directory is KnownDLL or if it the path starts with "\\KnownDll", and create a handle to the section object.

- For handling STATUS_IMAGE_NOT_AT_BASE, I maintain a map of module filepath -> module load count, at first load the module will get either the preferred image or the whatever the memory manager gives, and if load count is greater than 1, I return STATUS_IMAGE_NOT_AT_BASE, and for that I had to allow loading multiple copies of the same DLL.

- Refactored some stuff, some functions that were used for pe file parsing are now under winpe namespace.

- Added Dummy handler for NtFlushInstructionCache, WOW64 seems to need it while loading. 

- Remapping of win32u.dll and returning STATUS_IMAGE_NOT_AT_BASE is problematic because ntdll.dll checks for that, and eventually crashes, so as a workaround I don't allow remapping of it.


ahm3dgg:
> Cannot successfully run 32-bit apps. Don't know why, but In WOW64 you should maintain both `KnownDlls` and `KnownDlls32`.
> 
> Update: I cannot successfully run 32-bit app because of a regression introduced in other PR...

Can you tell me which one ? because I have the same issue

ahm3dgg:
> is_wow64_process

but I think we should also have \\KnownDLL32 for 64-bit, from a test I made, a 64-bit executable can obtain a section handle given \\KnownDLL32 as root directory, however it can't map the section due to machine type mismatch.

ahm3dgg:
> Do you mean an error `Failed to detect PE architecture from file`?

No this error is weird, it only occurs on the tests here.

ssvine:
Maybe you have another error. Can you provide a log or something?

ahm3dgg:
oh I see it now its because NtFlushInstructionCache is unimplemented, but idk if its the real issue or is it something else, but I will try to add it and see

ahm3dgg:
> > No this error is weird, it only occurs on the tests here.
> 
> No, no, this is not weird, it is pretty understandable. You set section object attribute `s.file_name` with wrong path. We have two kind of paths:
> 
> 1. A path that emulated program sees, like `C:\Windows\System32\kernel32.dll`
> 2. A path in the emulated environment like `D:\sogen-root\root\filesys\C\Windows\System32\kernel32.dll`
> 
> You use path 2, but you have to use path 1.
> 
> You can construct path 1 like this `auto known_dll_win_path = system_root_path / known_dll_name;`. There are two places in the code to fix setting `s.file_name`.

You are right, I missed that. Thank you very much, I added a dummy stub for NtFlushInstructionCache and now wow64 works fine !

ahm3dgg:
> Do you still need a workaround for `win32u.dll`?

Yes :p, here is the trace for that https://pastebin.com/isVLiJK4

ssvine:
I don't see obvious reasons why this tests failed. Something with serialization/deserialization.

Once I noticed (not related to this PR), that creating a snapshot of the running process and then continuing the process from the snapshot gives me different results from what I see when I run the process without snapshots. I didn't dig deep though. I hope this is not the case in this PR.

ahm3dgg:
> > I don't see obvious reasons why this tests failed. Something with serialization/deserialization.
> > Once I noticed (not related to this PR), that creating a snapshot of the running process and then continuing the process from the snapshot gives me different results from what I see when I run the process without snapshots. I didn't dig deep though. I hope this is not the case in this PR.
> 
> I did try to fix all things I can think about related to serialization and deserialization :p I don't know what am I missing, to be honest.

I got it, the serialzation tests work by comparing byte by byte, I have used std::unordered_map, which is actually *unordered*, the actual serialized bytes can vary between runs, using std::map solves that issue because it maintains some sort of order. 

What's interesting is, I have thought about that before and for some reason I thought that nah this is dumb can't be, but its actually is.

src/common/platform/win_pefile.hpp
```diff
@@ -593,6 +659,93 @@ namespace winpe
...
+        info.subsystem_major_version = optional_header.MajorSubsystemVersion;
+        info.subsystem_minor_version = optional_header.MinorSubsystemVersion;
+        info.dll_characteristics = optional_header.DllCharacteristics;
+        info.size_of_stack_reserve = optional_header.SizeOfStackReserve;
+        info.size_of_stack_commit = optional_header.SizeOfStackCommit;
+        info.size_of_code = optional_header.SizeOfCode;
+        info.loader_flags = optional_header.LoaderFlags;
+        info.checksum = optional_header.CheckSum;
+
+        // Check if image contains code
```
momo5502: image

almost all comments in your PR can be removed as the code speaks for itself

src/emulator/address_utils.hpp
```diff
@@ -52,3 +52,8 @@ constexpr uint64_t page_align_up(const uint64_t value, const uint64_t page_size
 {
     return align_up(value, page_size);
 }
+
+constexpr uint64_t rva_to_raw(uint64_t va_base, uint64_t raw_base, uint64_t rva)
```
momo5502: what is 'raw'?

ahm3dgg: Raw file offset

src/common/platform/win_pefile.hpp
ahm3dgg: > image
> almost all comments in your PR can be removed as the code speaks for itself

I am not the one who wrote this comment, I was just moving it to another file

ahm3dgg: > image
> almost all comments in your PR can be removed as the code speaks for itself

This code was lying in section.cpp used for parsing image file info then setting cached_file_info, the struct was also embedded in section, all what I did was just moving it

ahm3dgg: > image
> almost all comments in your PR can be removed as the code speaks for itself

I dont also remember writing that much comments, I think I wrote one or two since I am not used to anyways 😅

src/emulator/address_utils.hpp
ahm3dgg: > what is 'raw'?

I have renamed this to `rva_to_file_offset`, and removed the comments you pointed out (though it wasn't me),
also removed redundant checks for `win32u.dll`, that I used while testing, now I just check it in `module_manager::map_module`, and just return the original copy, that problem only occurs because of the ntdll loader, I am thinking if I can check if we finished the ntdll initialization stuff, and though after allow remapping.

src/common/platform/win_pefile.hpp
momo5502: lol, i just noticed :D

src/windows-emulator/process_context.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include "emulator_utils.hpp"
 #include "windows_emulator.hpp"
+#include "utils/io.hpp"
+#include "utils/buffer_accessor.hpp"
```
momo5502:
```suggestion
#include <utils/io.hpp>
#include <utils/buffer_accessor.hpp>
```

src/windows-emulator/syscalls/section.cpp
```diff
@@ -187,9 +72,10 @@ namespace syscalls
         const auto attributes = object_attributes.read();
 
         auto filename = read_unicode_string(c.emu, attributes.ObjectName);
+        auto filename_sv = std::u16string_view(filename);
```
momo5502, line 75: i don't think that's needed, std::u16string can implicitly be converted to std::u16string_view

src/windows-emulator/process_context.cpp
```diff
@@ -411,6 +441,10 @@ void process_context::setup(x86_64_emulator& emu, memory_manager& memory, regist
         }
     }
 
+    this->apiset = get_apiset_namespace_table(reinterpret_cast<const API_SET_NAMESPACE*>(apiset_container.data.data()));
```
momo5502: does this not need serialization?

src/windows-emulator/module/module_manager.hpp
```diff
@@ -157,6 +159,7 @@ class module_manager
     mapped_module* executable{};
     mapped_module* ntdll{};
     mapped_module* win32u{};
+    std::map<std::filesystem::path, uint64_t> module_load_count;
```
momo5502: if a dll is unloaded, this needs to be decremented and eventually be removed from the map

src/windows-emulator/process_context.cpp
```diff
@@ -753,3 +791,182 @@ const std::u16string* process_context::get_atom_name(const uint16_t atom_id) con
...
+        const auto char_count = value.data.size() / sizeof(char16_t);
+        const auto* data_ptr = reinterpret_cast<const char16_t*>(value.data.data());
+        if (data_ptr[char_count - 1] != u'\0')
+        {
+            continue;
+        }
+
+        auto known_dll_name = std::u16string(data_ptr, char_count - 1);
+        auto known_dll_path = system_root_path / known_dll_name;
+        auto local_known_dll_path = local_system_root_path / known_dll_name;
```
momo5502: please build the entire virtual path and then translate it entirely every time using file_system.translate(...). there can be mappings for specifiy subpaths that would be missed that way

src/windows-emulator/module/module_manager.hpp
ahm3dgg: > if a dll is unloaded, this needs to be decremented and eventually be removed from the map

Yes you are right, I forgot about that, I will fix these issues.

src/windows-emulator/syscalls/section.cpp
ahm3dgg: > i don't think that's needed, std::u16string can implicitly be converted to std::u16string_view

Yeah I thought so, but I was getting this error, that made me do that

Its defined as 

```cpp
    template <class Elem, class Traits>
    bool equals_ignore_case(const std::basic_string_view<Elem, Traits>& lhs, const std::basic_string_view<Elem, Traits>& rhs)
```

and The Error is:

```
could not deduce template argument for 'const std::basic_string_view<_Elem,_Traits> &' from 'std::u16string'```
```

I am probably going to add something like, This solves the problem, if you have better solutions tell me

```cpp
template <class Elem, class Traits, class Alloc> bool equals_ignore_case(const std::basic_string<Elem, Traits, Alloc>& lhs, const std::basic_string_view<Elem, Traits> rhs)
```

src/windows-emulator/process_context.cpp
ahm3dgg: > does this not need serialization?

well, maybe. my intension to put them in the process context for maybe in case it was needed in the future, for my case I was just using it while resolving api-sets for known dlls, and I do serialize them.

src/windows-emulator/process_context.cpp
ahm3dgg: > please build the entire virtual path and then translate it entirely every time using file_system.translate(...). there can be mappings for specifiy subpaths that would be missed that way

Its fine, that will be cleaner. however I am not really seeing the issue you are pointing out, the paths in this case is simple its just the root + filename. maybe you can give me an example

src/windows-emulator/process_context.cpp
momo5502: so it's a yes? it does need serialization?

src/windows-emulator/process_context.cpp
momo5502: the virtual filesystem supports mappings (pretty much like symlinks), if you translate some path and then just concatenate other subpaths, you can miss the resolving of those mappings.

e.g. you have a root like this: `C:\someroot\` which would translate to `/root/filesystem/c/someroot/`
and imagine there was a mapping (aka. symlink) registered at `C:\someroot\dir\` that points to `/blub/` on the host filesystem
and you're resolving a file `C:\someroot\dir\myfile.exe`, you would get `/root/filesystem/c/someroot/dir/myfile.exe`,
although it should resolve it to `/blub/myfile.exe`

src/windows-emulator/process_context.cpp
```diff
@@ -753,3 +794,178 @@ const std::u16string* process_context::get_atom_name(const uint16_t atom_id) con
...
+        if (data_ptr[char_count - 1] != u'\0')
+        {
+            continue;
+        }
+
+        auto known_dll_name = std::u16string(data_ptr, char_count - 1);
+        auto known_dll_path = system_root_path / known_dll_name;
+        auto local_known_dll_path = file_system.translate(system_root_path / known_dll_name);
+
+        if (!std::filesystem::exists(local_known_dll_path))
```
momo5502: remove that and instead use the read_file overload that returns a bool. otherwise you can run into races.

src/windows-emulator/process_context.hpp
```diff
@@ -87,6 +89,13 @@ struct process_context
     bool delete_atom(uint16_t atom_id);
     const std::u16string* get_atom_name(uint16_t atom_id) const;
 
+    template <typename T>
+    void build_knowndlls_section_table(registry_manager& registry, const file_system& file_system, bool is_32bit);
+
+    std::optional<section> get_knowndll_section_by_name(const std::u16string& name, bool is_32bit) const;
+    void add_knowndll_section(const std::u16string& name, const section& section, bool is_32bit);
+    bool is_knowndll_section_exists(const std::u16string& name, bool is_32bit) const;
```
momo5502:
```suggestion
    bool has_knowndll_section(const std::u16string& name, bool is_32bit) const;
```

src/windows-emulator/syscalls/section.cpp
momo5502: ah i see. honestly, then keep it the way it was. i don't think the overload you added is that helpful, cause there can be other variants that cause conflicts (e.g. if you flip the arguments). 
theoretically, you could also specify the arguments, then it doesn't have to deduce them, but specifying the traits is a bit ugly, so honestly, just keep your initial solution with the filename_sv, its fine

ssvine:
Microsoft defines [naming rules](https://learn.microsoft.com/en-us/windows/win32/apiindex/windows-apisets#api-set-contract-names) for api set names. I think it's better to check name requirements accordingly. For example, like this:

```c++
#include <regex>

static std::wregex apiset_pattern(LR"((api|ext)-[0-9A-Za-z-]+l\d+-\d+-\d+\.dll)", std::regex::icase);

static bool is_apiset_name(const std::u16string_view name)
{
    const std::wstring_view wname{reinterpret_cast<const wchar_t*>(name.data()), name.length()};
    return std::regex_match(wname.begin(), wname.end(), apiset_pattern);
}
```

ssvine:
And it seems, `build_knowndlls_section_table` doesn't build fully transitive library set?

For example if `a.dll` imports `b.dll` and `b.dll` imports `c.dll` then `c.dll` should also be included in the set.

ssvine:
> I think it's better to check name requirements accordingly. For example, like this:

I forgot about platforms with sizeof(wchar_t) == 4.

So, my example should be like this:

```c++
#include <regex>

static std::regex apiset_pattern(R"((api|ext)-[0-9A-Za-z-]+l\d+-\d+-\d+\.dll)", std::regex::icase);

// ...

            const auto known_dll_dep_name_str = (buffer.as_string(
                static_cast<size_t>(rva_to_file_offset(import_directory_vbase, import_directory_rbase, descriptor.Name))));
            auto known_dll_dep_name = u8_to_u16(known_dll_dep_name_str);

            if (std::regex_match(known_dll_dep_name_str.begin(), known_dll_dep_name_str.end(), apiset_pattern))
            {
                // ...
```

ahm3dgg:
> > I think it's better to check name requirements accordingly. For example, like this:
> 
> I forgot about platforms with sizeof(wchar_t) == 4.
> 
> So, my example should be like this:
> 
> ```c++
> #include <regex>
> 
> static std::regex apiset_pattern(R"((api|ext)-[0-9A-Za-z-]+l\d+-\d+-\d+\.dll)", std::regex::icase);
> 
> // ...
> 
>             const auto known_dll_dep_name_str = (buffer.as_string(
>                 static_cast<size_t>(rva_to_file_offset(import_directory_vbase, import_directory_rbase, descriptor.Name))));
>             auto known_dll_dep_name = u8_to_u16(known_dll_dep_name_str);
> 
>             if (std::regex_match(known_dll_dep_name_str.begin(), known_dll_dep_name_str.end(), apiset_pattern))
>             {
>                 // ...
> ```

I am not really familiar with regex or std::regex, and I am getting a bug now when building with sanitizer, I am not really sure what is the issue with it, I have tried your regex 

```cpp
static std::regex apiset_pattern(R"((api|ext)-[0-9A-Za-z-]+l\d+-\d+-\d+\.dll)", std::regex::icase);
```

and also tried a suggestion (by AI, since I honestly dont know regex :D) which is to escape "-" that didn't fix it as well.

```cpp
static std::regex apiset_pattern(R"((api|ext)-[0-9A-Za-z\-]+l\d+-\d+-\d+\.dll)", std::regex::icase);
```

I mean I honestly don't know what is the issue, at first I tried using your example as it is, and passing begin, end iterators. then I thought maybe passing a copy will work, I am not sure what is causing the problem

momo5502:
> And it seems, `build_knowndlls_section_table` doesn't build fully transitive library set?
> 
> For example if `a.dll` imports `b.dll` and `b.dll` imports `c.dll` then `c.dll` should also be included in the set.

i think it might make sense to solve that in a follow up PR. makes it easier to review for me 😂

ahm3dgg:
> > And it seems, `build_knowndlls_section_table` doesn't build fully transitive library set?
> > For example if `a.dll` imports `b.dll` and `b.dll` imports `c.dll` then `c.dll` should also be included in the set.
> 
> i think it might make sense to solve that in a follow up PR. makes it easier to review for me 😂

I think I have fixed it, though I am having an issue with matching the regex when building with sanitizer. I honestly think that matching with "api-" and "ext-" because really these are just standard system dlls, I am not expecting anything weird with them

ahm3dgg:
@momo5502

I will stay with matching with "api-" and "ext-" for now, I think its enough. I don't really know the cause of the issue, it only occurs with GCC with santizier enabled, I will try to see and maybe make a follow-up PR, since this PR is already very long 

so can we get this merged ?

src/windows-emulator/syscalls/section.cpp
ahm3dgg: > ah i see. honestly, then keep it the way it was. i don't think the overload you added is that helpful, cause there can be other variants that cause conflicts (e.g. if you flip the arguments). theoretically, you could also specify the arguments, then it doesn't have to deduce them, but specifying the traits is a bit ugly, so honestly, just keep your initial solution with the filename_sv, its fine

I did returned them back

ssvine:
> an issue with matching the regex when building with sanitizer

It is a bug in [address sanitizer](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117457) and it is not yet fixed.

Anyway, @ahm3dgg great work!

momo5502:
Even fix commits can be helpful. Seeing a bunch of fix commits in the history can reveal that this particular place in the code was tricky to develop. If I see this, I'm especially careful when making changes, because this place in the code was tuned to work the way it does.

ahm3dgg:
> Anyways, thank you, @ahm3dgg for your contribution :) Feel free to create follow up PRs, and don't worry about your history 😂

Thank you very much

PR #696, opened 2026-01-08
Fix user_handle_table memory handling

ssvine:
This PR fixes `user_handle_table` for WOW64. Otherwise we get AV during 32-bit `user32.dll` initialization.

### DIFF
--- a/src/windows-emulator/user_handle_table.hpp
+++ b/src/windows-emulator/user_handle_table.hpp
@@ -12,15 +12,17 @@ class user_handle_table
     {
     }
 
-    void setup()
+    void setup(const bool is_wow64_process)
     {
+        this->is_wow64_process_ = is_wow64_process;
+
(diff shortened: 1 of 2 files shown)


src/windows-emulator/user_handle_table.hpp
```diff
@@ -129,11 +131,19 @@ class user_handle_table
...
+        const auto base = memory_->find_free_allocation_base(size, allocation_base);
+        return memory_->allocate_memory(size, permissions, false, base);
+    }
+
     uint64_t server_info_addr_{};
     uint64_t handle_table_addr_{};
     uint64_t display_info_addr_{};
     std::vector<bool> used_indices_{};
     memory_manager* memory_{};
+    bool is_wow64_process_{};
```
momo5502, line 148: this needs to be serialized

ssvine: Done

PR #697, opened 2026-01-08
Improve Windows version detection and LdrSystemDllInitBlock initialization

brian763:
- Add WINDOWS_VERSION enum and PS_SYSTEM_DLL_INIT_BLOCK sizes for WOW64 support across different Windows builds.
- Read system information (SystemRoot, BuildNumber, UpdateBuildRevision) from registry instead of hardcoded paths.
- Add build comparison helpers in process_context for precise build checks.

### DIFF
--- a/src/windows-emulator/kusd_mmio.cpp
+++ b/src/windows-emulator/kusd_mmio.cpp
@@ -1,6 +1,7 @@
 #include "std_include.hpp"
 #include "kusd_mmio.hpp"
 #include <utils/time.hpp>
+#include <utils/string.hpp>
 #include "windows_emulator.hpp"
 #include "version/windows_version_manager.hpp"
 
@@ -83,9 +84,7 @@ namespace
         kusd.ProcessorFeatures.arr[PF_RDPID_INSTRUCTION_AVAILABLE] = 0;
 
         const auto& system_root = version.get_system_root();
-        const auto& root_str = system_root.u16string();
-        const auto copy_size = std::min(root_str.size() * sizeof(char16_t), sizeof(kusd.NtSystemRoot.arr) - sizeof(char16_t));
-        memcpy(&kusd.NtSystemRoot.arr[0], root_str.data(), copy_size);
+        utils::string::copy(kusd.NtSystemRoot.arr, std::u16string_view{system_root.u16string()});
 
         kusd.ImageNumberLow = IMAGE_FILE_MACHINE_AMD64;
         kusd.ImageNumberHigh = IMAGE_FILE_MACHINE_AMD64;
--- a/src/common/utils/string.hpp
+++ b/src/common/utils/string.hpp
@@ -30,6 +30,21 @@ namespace utils::string
         array[std::min(Size - 1, size)] = {};
     }
 
+    template <typename T, size_t Size, class Traits = std::char_traits<T>>
+        requires(std::is_trivially_copyable_v<T>)
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays)
+    void copy(T (&array)[Size], const std::basic_string_view<T, Traits> str)
+    {
+        if constexpr (Size == 0)
+        {
+            return;
+        }
+
+        const auto size = std::min(Size - 1, str.size());
+        memcpy(array, str.data(), size * sizeof(T));
+        array[size] = {};
+    }
+
     inline char char_to_lower(const char val)
     {
         return static_cast<char>(std::tolower(static_cast<unsigned char>(val)));


ahm3dgg:
> Hello, since you are already working on version info, I wanted to issue a PR to set PEB OS version fields with what's in the registry, same for peb32
> 
> ```c++
>     this->peb64.access([&](PEB64& p) {
>        ....
>         p.OSPlatformId = 2;
>         p.OSMajorVersion = 0x0000000a;
>         p.OSBuildNumber = 0x00006c51;
>       ....
>     });
> ```
> 
> can be read from `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, `OSBuildNumber`, `OSMajorVersion`, `OSMinorVersion`.
> 
> Also KUSD will need to be updated, Just for consistency, maybe its also better to just seperate version info related code into separate files.

Just to be more clear, what I meant that it would be better if you tried implementing that in your PR. because I didn't want to issue a PR and maybe both things gets merged resulting into so inconsistency in the design.

momo5502 (approved):

momo5502:
> @ahm3dgg Thanks for the suggestion! I will add PEB/PEB32 OS version fields and KUSD updates in this PR, and refactor the version info code into separate files for better consistency.

PR looks good, however, i will wait for those changes before merging

ahm3dgg:
> @ahm3dgg hi, could you please take a look at this PR?

btw couldn't `get_system_dll_init_block_size` be a function that uses the version manager instead of being a method of it ?

ahm3dgg:
> > > > @ahm3dgg hi, could you please take a look at this PR?您好，能否请您看一下这个 PR？
> > > 
> > > 
> > > btw couldn't `get_system_dll_init_block_size` be a function that uses the version manager instead of being a method of it ?顺便问一下， `get_system_dll_init_block_size` 能不能是一个使用版本管理器的函数，而不是它的一个方法？
> > 
> > 
> > Also was thinking if _WINDOWS_VERSION could be part of windows_version_manager.hpp, there is also this check我还考虑过 _WINDOWS_VERSION 是否可以作为 windows_version_manager.hpp 的一部分，其中也包含这个检查。
> > ```c++
> >         if (version.is_build_before(26040))
> >         {
> >             monitor.b20.monitorDpi = 96;
> >             monitor.b20.nativeDpi = monitor.b20.monitorDpi;
> >             monitor.b20.cachedDpi = monitor.b20.monitorDpi;
> >             monitor.b20.rcMonitorDpiAware = monitor.rcMonitor;
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > you could also put this version in the enum and not hardcode it.你也可以将此版本放在枚举中，而不是硬编码。
> 
> @ahm3dgg Regarding _WINDOWS_VERSION and the hardcoded 26040: Build 26040 does not map to any official Windows release name like 21H2 or 24H2. I am not sure why this specific build number was chosen for the DPI check. Adding it to the enum without a proper official version name would be misleading, so I'd prefer to keep it as is for now.

Oh interesting, I honestly not familiar with that DPI stuff, I was just pointing out the style, can you point me where these checks are being made, so I can take a look at them ?

ahm3dgg:
> @ahm3dgg it's in process_context::setup in src/windows-emulator/process_context.cpp.

Oh okay, I did thought this was reverse engineered or sth so I was asking about where that stuff gets set in windows, but I think here you mean that it was chosen by the author

brian763:
@momo5502 Do you have any other questions regarding this PR? If everything looks good to you, could we proceed with merging this? Thank you!

src/windows-emulator/kusd_mmio.cpp
```diff
@@ -80,8 +82,10 @@ namespace
         kusd.ProcessorFeatures.arr[PF_RDTSCP_INSTRUCTION_AVAILABLE] = 1;
         kusd.ProcessorFeatures.arr[PF_RDPID_INSTRUCTION_AVAILABLE] = 0;
 
-        constexpr std::u16string_view root_dir{u"C:\\Windows"};
-        memcpy(&kusd.NtSystemRoot.arr[0], root_dir.data(), root_dir.size() * 2);
+        const auto& system_root = version.get_system_root();
+        const auto& root_str = system_root.u16string();
+        const auto copy_size = std::min(root_str.size() * sizeof(char16_t), sizeof(kusd.NtSystemRoot.arr) - sizeof(char16_t));
+        memcpy(&kusd.NtSystemRoot.arr[0], root_str.data(), copy_size);
```
momo5502: this could be extracted in a utils::string, something that copies a string of arbitrary char time into a fixed size array and guaranteey 0 termination

momo5502: something similar already exists, it could be extended to support basic_string_view:

https://github.com/momo5502/sogen/blob/main/src/common/utils/string.hpp#L21

brian763: Done. I have added basic_string_view overload for utils::string::copy and updated the code accordingly.

PR #698, opened 2026-01-09
Making a New Type for Serializable Map Data structures

ahm3dgg:
I thought this would be better, as to avoid failing serialization tests, most maps are std::map and that's correct, however as newcomer to the project I didn't thought about it and I am used to use std::unordered_map more.

### DIFF
--- a/src/emulator/serialization.hpp
+++ b/src/emulator/serialization.hpp
@@ -13,7 +13,7 @@
 #include <map>
 
 template <typename... Args>
-using serialized_map = std::map<Args...>;
+using serializable_map = std::map<Args...>;
 
 namespace utils
 {
--- a/src/windows-emulator/handles.hpp
+++ b/src/windows-emulator/handles.hpp
@@ -167,7 +167,7 @@ class handle_store : public generic_handle_store
 {
   public:
     using index_type = uint32_t;
-    using value_map = serialized_map<index_type, T>;
+    using value_map = serializable_map<index_type, T>;
 
     bool block_mutation(bool blocked)
     {
--- a/src/windows-emulator/memory_manager.hpp
+++ b/src/windows-emulator/memory_manager.hpp
@@ -61,7 +61,7 @@ class memory_manager : public memory_interface
         nt_memory_permission permissions{};
     };
 
-    using committed_region_map = serialized_map<uint64_t, committed_region>;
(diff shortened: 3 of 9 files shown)


ahm3dgg:
> Maybe something like this will count:
> 
> ```c++
> #include <ranges>
> #include <algorithm>
> 
> template <typename T>
> concept MapLike = requires(T m, const typename T::key_type& key) {
>     typename T::key_type;
>     typename T::mapped_type;
>     typename T::value_type;
> 
>     { m[key] } -> std::same_as<typename T::mapped_type&>;
>     { m.at(key) } -> std::same_as<typename T::mapped_type&>;
>     { m.find(key) } -> std::same_as<typename T::iterator>;
>     { m.size() } -> std::same_as<typename T::size_type>;
> };
> 
> template <typename T>
> concept OrderedMap = requires {
>     MapLike<T>;
>     typename T::key_compare;
>     typename T::reverse_iterator;
> };
> 
> template <typename T>
> concept UnorderedMapWithSortableKeys = requires {
>     MapLike<T>;
>     typename T::key_equal;
>     typename T::hasher;
>     std::totally_ordered<typename T::key_type>;
> };
> 
> template <MapLike Map>
> void write_map(const Map& map)
> {
>     this->write<uint64_t>(map.size());
> 
>     if constexpr (OrderedMap<Map>)
>     {
>         for (const auto& entry : map)
>         {
>             this->write(entry.first);
>             this->write(entry.second);
>         }
>     }
>     else if constexpr (UnorderedMapWithSortableKeys<Map>)
>     {
>         std::vector<typename Map::key_type> keys{};
>         for (const auto& key : map | std::views::keys)
>         {
>             keys.push_back(key);
>         }
> 
>         std::sort(keys.begin(), keys.end());
> 
>         for (const auto& key : keys)
>         {
>             this->write(key);
>             this->write(map.at(key));
>         }
>     }
>     else
>     {
>         static_assert(false, "write_map parameter should be a map or unordered_map with comparable keys");
>     }
> }
> ```
> 
> Unordered maps with comparable keys can be sorted. If we sort the keys before serialization, then, I think, serialization tests will be passed. `read_map` does not need any changes.

That looks good

ssvine:
Ok, let's just keep it simple at the moment.

PR #873, opened 2026-04-10
Fix private namespace

ssvine:
This is a follow-up to one of my previous pull requests. I missed that the system disables passing the name through object attributes.

I also fixed getting `ObjectNameInformation` in `NtQueryObject`. When an object (of any type) is unnamed, the system returns `UNICODE_STRING` with `{.Length=0, .MaximumLength=0, .Buffer=nullptr}`


momo5502:
@codex review

momo5502 (approved):

PR #880, opened 2026-04-13
Match windows syscall id wraparound behaviour

0xeeffoc:
Mask syscall ids with 0x3FFF instead of 0xFFFF
Windows wraps syscall ids every 0x4000, so higher bits are ignored
This aligns syscall handling with actual Windows behavior and avoids incorrect id resolution.

Fixes a emulator crash when higher syscall ids are used (e.g. EAX=0x4025). The emulator previously used the full value as an index, leading to out of bounds lookup since Windows wraps this to 0x0025.


momo5502 (approved):

momo5502:
Thank you for your contribution :) Once formatting is fixed, this can be merged.
You will need clang-format for that

PR #888, opened 2026-04-24
Route named pipe I/O through shared per-name queues

momo5502:
## Summary
- route named-pipe reads and writes through shared per-name queues in `process_context`
- stop acknowledging named-pipe writes while discarding the payload
- serialize the shared named-pipe message state with the rest of process state

## Notes
- This PR keeps the fix scoped to truthful named-pipe data flow.
- It does not attempt to fully model native named-pipe server/client semantics beyond shared same-name message queues.


momo5502:
@codex review

PR #913, opened 2026-05-03
Fix regression caused by NtQuerySecurityAttributesToken

Igoorx:
The latest version of sogen fails to run some apps that used to work, and I managed to bisect the issue back to PR https://github.com/momo5502/sogen/pull/764. It seems like returning anything besides NOT_SUPPORTED for `WIN://SYSAPPID` causes adverse effects, so this PR adds a special case for that.


momo5502 (approved):

momo5502:
ideally, a test would be nice

Igoorx:
Ouch, looks like this breaks `SHGetKnownFolderPath`. I guess this is a more annoying issue than I first expected.

> ideally, a test would be nice

Agreed, but I don't know exactly why this issue is happening in the first place. `NtQuerySecurityAttributesToken` is a syscall called at the start of the app, and then, after lots of instructions, when the app is loading some specific dependencies, it exits with status code `C0000145`. I guess I will have to investigate more to find out what the correct fix is.

Igoorx:
Okay, this should be good now. The core issue was that the activation context was failing to be created, which caused dynamic libraries with SxS manifests to fail to load. I've adjusted the fix so it now works without breaking `SHGetKnownFolderPath`, and I've added a test for the activation context.

momo5502:
awesome thank you

PR #916, opened 2026-05-04
Decouple WHP GPAs from guest virtual addresses

Igoorx:
This PR changes the WHP backend so guest virtual allocations are backed by physical addresses that WHP can accept. The WHP backend previously used the guest virtual page address directly as the GPA passed to `WHvMapGpaRange`. That works for low mappings, but fails for high user-mode allocations such as `0x00000bf010000000` because WHP rejects those values as physical addresses.

Disclaimer: This was done with the help of AI. I have zero experience with WHP 😅


momo5502:
@codex review

momo5502 (approved):

PR #918, opened 2026-05-05
Fix MMIO updates on the WHP backend

Igoorx:
This PR fixes MMIO updates on the WHP backend. Before this change, updates on read were only triggered once and then never again.

However, this works a bit differently from the other backends: updating on every single read causes a significant performance hit, so as a compromise, updates are limited to once every 4ms and are armed on VM exits. This keeps the performance impact negligible.

### DIFF
--- a/src/backends/whp-emulator/whp_x86_64_emulator.cpp
+++ b/src/backends/whp-emulator/whp_x86_64_emulator.cpp
@@ -816,6 +816,7 @@ namespace whp
 
                 while (!this->stop_requested_)
                 {
+                    this->expire_mmio_read_grace_pages();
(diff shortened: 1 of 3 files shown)


Igoorx:
Looks like CI is failing in something unrelated, the same thing is happening in the other PR 🤔

momo5502:
github runners updated to visual studio 2026. the compiler there has a bug :(

momo5502:
Should be "fixed" now, if you rebase your PRs

Igoorx:
Sleep seems to be non-deterministic in the emulator 😨. But since that is unrelated to this change, I've updated the test to use a different time-hogging method for measuring timeGetTime.

momo5502 (approved):

PR #922, opened 2026-05-09
Fix AVX instructions in the WHP backend

Igoorx:
This PR fixes AVX instructions in the WHP backend. The issue was caused by the XCR0 register being initialized to 0x3, which only enables x87 and SSE state. AVX instructions also require the YMM state bit to be enabled, so XCR0 needs to be initialized to 0x7 instead.

### DIFF
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -598,6 +598,7 @@ namespace
         watch_system_objects(context, options.modules, options.verbose_logging, options.concise_logging);
 
         const auto& exe = *win_emu->mod_manager.executable;
+        const auto is_whp = win_emu->emu().get_name() == "Windows Hypervisor Platform";
 
         win_emu->emu().hook_instruction(x86_hookable_instructions::cpuid, [&] {
             auto& emu = win_emu->emu();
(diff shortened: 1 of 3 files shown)


src/backends/whp-emulator/whp_x86_64_emulator.cpp
```diff
@@ -1608,7 +1608,7 @@ namespace whp
                 {
                     names.push_back(WHvX64RegisterXCr0);
                     WHV_REGISTER_VALUE xcr0{};
-                    xcr0.Reg64 = 0x3ull;
```
momo5502, line 1611: what if the CPU of the user doesn't support AVX?

src/analyzer/main.cpp
```diff
@@ -599,6 +599,9 @@ namespace
 
         const auto& exe = *win_emu->mod_manager.executable;
 
+        const auto* env = getenv("EMULATOR_WHP");
```
momo5502: that environment variable is an implementation detail of the backend selection. feels unclean to duplicate it here. maybe expose the criterion you need to check on below from the emulator interface, or expose the WHP selection from the backend selection

src/backends/whp-emulator/whp_x86_64_emulator.cpp
Igoorx: I guess in this case CPUID wouldn’t advertise AVX support, and the program would crash anyway if it encountered AVX instructions. But I see your point, making things consistent should be safer.

src/analyzer/main.cpp
Igoorx: Oh right, I guess I could just find that out using `get_name`, although exposing a `get_backend_type` would be cleaner... But this CPUID check is already a dirty hack, so I don’t think it’s worth overthinking. Let me know if you think it should be done differently.

momo5502 (approved):

PR #923, opened 2026-05-10
Implement static_socket TCP ops, ProcessVmCounters, and tidy fixes

0xeeffoc:
Stub NtQueryInformationProcess ProcessVmCounters
Fixes K32GetProcessMemoryInfo aborting emulation
Implement TCP ops (listen/accept/send/recv/poll_sockets) in static_socket_factory, replacing the previous runtime_error("Not implemented") paths that surfaced as NtDeviceIoControlFile exceptions whenever an emulated app touched AFD.

### DIFF
--- a/src/windows-emulator/network/static_socket_factory.cpp
+++ b/src/windows-emulator/network/static_socket_factory.cpp
@@ -1,33 +1,99 @@
 #include "static_socket_factory.hpp"
+#include "../logger.hpp"
 
+#include <cstdarg>
+#include <cstdio>
+#include <cstring>
+#include <deque>
 #include <queue>
 #include <stdexcept>
 #include <unordered_map>
 
 #include <network/socket.hpp>
 
+#ifndef POLLIN
+#define POLLIN 0x0001
+#endif
+#ifndef POLLOUT
+#define POLLOUT 0x0004
+#endif
+#ifndef POLLHUP
+#define POLLHUP 0x0010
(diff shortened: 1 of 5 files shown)


momo5502:
@codex review

src/windows-emulator/network/static_socket_factory.cpp
```diff
@@ -132,24 +318,97 @@ namespace network
...
 
     std::unique_ptr<socket_factory> create_static_socket_factory()
     {
-        return std::make_unique<static_socket_factory>();
+        return std::make_unique<static_socket_factory_impl>();
+    }
+
+    void set_static_socket_factory_logger(socket_factory& factory, logger* log)
+    {
+        if (auto* impl = dynamic_cast<static_socket_factory_impl*>(&factory))
```
momo5502: i'm not a big fan of how logging was done here. a lot of string operations that perform allocations that are not needed, when logging is not used. ofc the impact of those is negligible, yet it feels a bit unclean.
care to elaborate your design decisions on that?

i feel like the the socket factory interface provides everything to be able to create a logging_socket_factory, that just wraps the existing socket_factory.
like a proxy object, that wraps the real objects, and just adds logging ontop and forwards all virtual calls.

0xeeffoc: Artifacts from other branch, sorry about that.

PR #924, opened 2026-05-10
Add debugging support to the WHP backend

Igoorx:
This PR adds support for debugging using the WHP backend.

There is one thing to keep in mind, though. As far as I understand, breakpoints in WHP can only be either stealthy (page-fault-based and slower in some circumstances) or open (INT3/hardware-based and always fast). Because of that, I added a new parameter to choose between the two specifically for the WHP backend.

Disclaimer: This PR was mostly done with the help of AI. I reviewed and tested it in a basic debugging workflow and didn’t notice any issues.


momo5502:
@codex review

momo5502 (approved):

PR #944, opened 2026-05-16
Support fast fail

ssvine:
This PR adds support for [`__fastfail`](https://learn.microsoft.com/en-us/cpp/intrinsics/fastfail?view=msvc-170), which is essentially an `int 29h` interrupt instruction.


momo5502:
@codex review

momo5502 (approved):

PR #945, opened 2026-05-17
Memory view tab

0xeeffoc:
Browse memory regions when the emulation is paused and inspecting with hex bytes and ASCII visualization


momo5502:
@codex review

momo5502:
really interesting contribution, thank you :D
I checked it out and it looks great. i have a few ideas for improvement, e.g. listing modules in the memory regions separately (maybe even their sections).
this opens up possibilities for experiences i have never really thought about. maybe disassembly could be added, or even graph views. obviously none of this should be done in this PR, but i like the new territory we are exploring here :)

0xeeffoc:
Thanks!
Wanted to take it in a direction where everything could be done on the site itself
Of course I have other things to add like copying or maybe editing actual bytes instead of just viewing, and maybe transforming tab almost like cyberchef
Also your ideas for disasm and graph sound awesome!

src/debugger/event_handler.cpp
```diff
@@ -85,6 +89,173 @@ namespace debugger
...
+            case memory_region_kind::section_image:
+                return "image";
+            case memory_region_kind::mmio:
+                return "mmio";
+            default:
+                return "unknown";
+            }
+        }
+
+        std::string to_hex(uint64_t value)
```
momo5502: feels like we already have something similar: https://github.com/momo5502/sogen/blob/main/src/common/utils/string.hpp#L124
except that it doesn't prepend 0x. maybe this can be reused

src/debugger/event_handler.cpp
```diff
@@ -85,6 +89,173 @@ namespace debugger
             send_event(response);
         }
 
+        std::string permission_string(const memory_permission perm)
```
momo5502: this already exists: https://github.com/momo5502/sogen/blob/main/src/windows-emulator/memory_utils.hpp#L7
maybe we can reuse this as well.
it uses lowercase letters, but you can also change them to upper case in the existing implementation, if you prefer that. i don't have any strong preference

momo5502 (approved):

momo5502:
> Thanks! Wanted to take it in a direction where everything could be done on the site itself Of course I have other things to add like copying or maybe editing actual bytes instead of just viewing, and maybe transforming tab almost like cyberchef Also your ideas for disasm and graph sound awesome!

love it :D

PR #735, opened 2026-02-10
feat: Linux x86-64 user-mode emulation with ELF loading, syscall support, web integration, and differential test suite

19h:
Introduces a complete Linux x86-64 user-mode emulation subsystem, enabling execution and debugging of statically and dynamically linked ELF binaries (glibc/musl) within the existing emulator framework—both native and in-browser via Emscripten/WASM.

## Core Emulation (`src/linux-emulator/`)

- **CPU & Memory**: x86-64 long mode initialization (GDT, CS=0x33), Linux-style `mmap`-based memory model, `brk` heap, full 48-bit user address space.
- **Syscalls**: ~100 implemented syscalls spanning file I/O (`openat`, `getdents64`, `pipe2`), memory (`mmap`, `mprotect`, `munmap` with strict alignment enforcement), process control (`exit`, `arch_prctl`, `clone`, `futex`), scheduling stubs, time, and IO multiplexing (`poll`, `select`, `epoll`).
- **ELF Loader**: `PT_LOAD` segment mapping, `PT_INTERP` dynamic linker resolution with sysroot path search (`/lib/x86_64-linux-gnu/`, `/usr/lib64/`), `R_X86_64_RELATIVE`/`GLOB_DAT`/`IRELATIVE` relocation handling, `DT_RPATH`/`DT_RUNPATH` with `$ORIGIN` substitution.
- **IFUNC Resolution**: Eager execution of `R_X86_64_IRELATIVE` resolvers pre-entry via scratch stack + `hlt` sentinel trapping, patching GOT in-place.
- **TLS**: x86-64 Variant II layout—correctly sized TLS block from `PT_TLS`, TCB at block end, `fs:0x00` self-pointer, `fs:0x10` pthread pointer, `.tdata` initialization.
- **Signals**: `rt_sigframe` construction on user stack, synthetic `sigreturn` trampoline page, `rt_sigaction`/`rt_sigprocmask` support.
- **Virtual Filesystems**: `/proc/self/{maps,cmdline,auxv,status}` synthesis, minimal vDSO ELF generation (`__vdso_clock_gettime`, `__vdso_getcpu`).
- **Filesystem**: Host passthrough prefix registration for out-of-root binaries, relative path fallback through passthrough prefixes, auto-creation of `/tmp` and `/var/tmp`.

## Web Integration (`page/`)

- Linux/Windows mode toggle in settings UI with mode-aware control visibility.
- Dedicated `linux-emulator-worker.js` with IDBFS mount at `/root`.
[...]

PR #949, opened 2026-05-17
Linux x86-64 user-mode emulation

momo5502:
This is a continuation of #735


cmake/compiler-env.cmake
```diff
@@ -137,6 +137,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Emscripten")
 
   add_link_options(
     -fexceptions
+    --minify=0
```
momo5502: hm

momo5502:
@codex review

PR #959, opened 2026-05-19
Fix stat implementation in the Linux emulator

Igoorx:
This PR fixes a issue in the Linux emulator’s stat implementation on Windows. Previously, it relied on Windows stat fields being populated correctly, but Windows does not fill in dev/ino, which caused issues with code that checks these values.

### DIFF
--- a/src/linux-emulator/syscalls/file.cpp
+++ b/src/linux-emulator/syscalls/file.cpp
@@ -19,6 +19,10 @@
 
 using namespace linux_errno; // NOLINT(google-build-using-namespace)
 
+#ifdef _WIN32
+#define lstat stat
+#endif
+
 namespace
 {
     constexpr int LINUX_AT_FDCWD = -100;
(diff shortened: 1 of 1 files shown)


momo5502:
Nice one :D Thanks :)

momo5502 (approved):

momo5502:
once clang-tidy is fixed, this can be merged

Igoorx:
You’re welcome! :D

By the way, the Linux emulator playground at sogen.dev doesn't seem to be working because `https://sogen.dev/linux32/linux-analyzer.js` returns a 404. I can open an issue about it if you want.

momo5502:
> You’re welcome! :D
> 
> By the way, the Linux emulator playground at sogen.dev doesn't seem to be working because `https://sogen.dev/linux32/linux-analyzer.js` returns a 404. I can open an issue about it if you want.

Good catch, thanks. #961 should fix this.

PR #963, opened 2026-05-20
Optimize file reading and module mapping for large binaries

AdvDebug:
This pull request significantly refactors and improves the Windows emulator's PE module mapping logic for large binaries, by reading large files directly instead of using chunks (for normal, non-special files) and optimize module mapping for large binaries by mapping each section individually (unicorn hates mapping a large section in a bulk), and reduce the amount of calls needed for protect_memory since it is expensive. in addition to making relocations, export, import, etc read directly from mapped memory instead of using temporary buffers or building a full mapped byte vector.


momo5502:
@codex review

src/windows-emulator/module/module_mapping.cpp
momo5502, line 1: seems like line endings are goofed on this file

src/windows-emulator/module/module_mapping.hpp
momo5502, line 1: same here, line endings should be reverted to what they were

src/windows-emulator/module/module_mapping.cpp
```diff
@@ -265,6 +472,92 @@ namespace
...
+                                        const uint64_t nt_headers_offset, const PEOptionalHeader_t<T>& optional_header)
+    {
+        binary.sections.clear();
+        binary.exports.clear();
+        binary.imports.clear();
+        binary.imported_modules.clear();
+        binary.address_names.clear();
+
+        const auto image_size = static_cast<size_t>(binary.size_of_image);
+        if (!memory.allocate_memory(binary.image_base, image_size, memory_permission::read_write, true))
```
momo5502: this should fix the tests:

```suggestion
        if (!memory.allocate_memory(binary.image_base, image_size, memory_permission::all, true))
```

the reason is that the memory manager tracks allocation patterns, as they are required for certain syscalls. this changes the initial protection of the pages

momo5502 (approved):

PR #966, opened 2026-05-20
Rework section_first_execute trigger for WHP support

Igoorx:
This PR moves the trigger for the `section_first_execute` event to a new callback called `on_section_first_execution`, which can be triggered by the `on_instruction` callback or by a memory range hook (in WHP). This makes the `section_first_execute` event work on the WHP backend, along with concise logging mode.

This PR also includes the following changes:
- Fixes a crash when removing hooks inside a hook callback in Icicle.
- Adds support for hooking execution of a memory range in Icicle and WHP, and exposes the existing Unicorn implementation.

### DIFF
--- a/src/analyzer/analysis.cpp
+++ b/src/analyzer/analysis.cpp
@@ -441,32 +441,26 @@ namespace sogen
             ++c.instructions[instructions[0].id];
         }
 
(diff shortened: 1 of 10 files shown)


momo5502:
@codex review

momo5502 (approved):

PR #990, opened 2026-05-23
Implement --call-count and --break-call arguments

Igoorx:
This PR adds two new arguments to the analyzer CLI: `--call-count` and `--break-call`.

When debugging with sogen, I often want to break at a precise point of execution.
My previous workaround was to manually pause at the right moment and then pause in
GDB, which works fine most of the time but gets annoying when a function or syscall
is called more than once. These two arguments make that easier:

- `--call-count`: Prefixes each function/syscall execution with an incrementing counter.
- `--break-call <n>`: Breaks in GDB when the specified call count is reached.

### DIFF
--- a/src/windows-analyzer/analysis.cpp
+++ b/src/windows-analyzer/analysis.cpp
@@ -18,6 +18,7 @@ namespace sogen
     namespace
     {
         constexpr size_t MAX_INSTRUCTION_BYTES = 15;
+        constexpr uint64_t SYSCALL_INSTRUCTION_SIZE = 2;
 
         template <typename Return, typename... Args>
         std::function<Return(Args...)> make_callback(analysis_context& c, Return (*callback)(analysis_context&, Args...))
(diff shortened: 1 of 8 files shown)


momo5502:
@codex review

src/windows-analyzer/analysis.cpp
```diff
@@ -441,6 +442,35 @@ namespace sogen
...
+
+        bool break_before_traced_syscall(analysis_context& c, const uint64_t call_count, const uint64_t address)
+        {
+            if (!break_before_traced_call(c, call_count))
+            {
+                return false;
+            }
+
+            c.syscall_to_resume_after_break = address;
+            c.win_emu->emu().reg<uint64_t>(x86_register::rip, address - SYSCALL_INSTRUCTION_SIZE);
```
momo5502, line 470: is this needed? :O

momo5502 (approved):

src/windows-analyzer/analysis.cpp
Igoorx: Yeah, it is needed to get RIP to before the syscall runs. Otherwise, the break leaves us after the syscall. Technically, it doesn't matter for GDB debugging, but I wanted to keep it consistent with the break on a function call.

PR #993, opened 2026-05-25
Debugger break before first instruction

momo5502:
Allow debugger to be armed before emulation starts and pause at first instruction.

### DIFF
--- a/src/windows-analyzer/main.cpp
+++ b/src/windows-analyzer/main.cpp
@@ -54,6 +54,7 @@ namespace sogen
             bool log_foreign_module_access{false};
             bool tenet_trace{false};
             bool prepend_call_count{false};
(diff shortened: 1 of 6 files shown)


src/windows-analyzer/main.cpp
```diff
@@ -847,6 +854,7 @@ namespace sogen
             printf("  -ss, --skip-syscalls       Skip the logging of regular syscalls\n");
             printf("  -rep, --reproducible       Stub clocks and other mechanisms to make executions reproducible\n");
             printf("  --env <var> <value>        Set environment variable\n");
+            printf("  --break-start             Pause before executing the first instruction\n");
```
momo5502, line 860: the --break-start option should only be availbale on emscripten builds

PR #1000, opened 2026-05-29
Fix pseudo-handle issues

Igoorx:
This PR fixes an issue where several syscalls failed to recognize the current process and current thread pseudo-handles, causing these valid handles to be rejected.


momo5502:
@codex review

PR #1004, opened 2026-06-02
Integrate UI panel to playground

momo5502:


src/windows-emulator/windows_emulator.cpp
```diff
@@ -3,6 +3,10 @@
 
 #include "cpu_context.hpp"
 
+#ifdef OS_EMSCRIPTEN
+#include <emscripten.h>
+#endif
+
```
momo5502: this include is not needed anymore

src/windows-emulator/windows_emulator.cpp
```diff
@@ -425,7 +429,9 @@ namespace sogen
             }
             else
             {
+#ifndef OS_EMSCRIPTEN
                 std::this_thread::sleep_for(1ms);
+#endif
```
momo5502: is this ifdef needed?

PR #1006, opened 2026-06-02
Fix window message callbacks and WoW64 fixes

Igoorx:
This PR includes the following changes:

* Fixes the window message callbacks, which were set up slightly incorrectly.
* Fixes some WoW64 crashes by restoring additional registers, improving the implementation of the `ThreadWow64Context` set-information class, and implementing the `ProcessWow64Information` information class.
* Fixes argument building to correctly quote arguments that contain spaces.
* Enables the ~~Working Directory,~~ Interrupts, and Message Queue tests for x86, since they appear to be working in the current repo state.

### DIFF
--- a/src/emulator-platform/platform/window.hpp
+++ b/src/emulator-platform/platform/window.hpp
@@ -126,10 +126,10 @@ namespace sogen
         uint8_t rgbReserved[32];
     };
 
+#define EMU_HWND_MESSAGE ((hwnd) - 3)
+
 #ifndef OS_WINDOWS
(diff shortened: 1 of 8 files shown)


Igoorx:
Hmm... The message queue test is somehow failing on CI, and looking at the log, it seems to fail before dispatching messages. My best guess is that something is going wrong with the addresses.

Igoorx:
Fixed, it was just an issue with the HWND_MESSAGE macro.

momo5502 (approved):

PR #1022, opened 2026-06-05
Deliver pending user APCs before an alertable wait times out

momo5502:
## Problem

An alertable wait that has user APCs queued (e.g. `SleepEx(_, TRUE)` or `WaitForSingleObjectEx`) must run those APCs (returning `WAIT_IO_COMPLETION`) **before** the wait completes by timeout.

`is_thread_ready` latched `STATUS_SUCCESS`/`STATUS_TIMEOUT` into `pending_status` as soon as the await time elapsed. `switch_to_thread` only dispatches APCs when no status is already pending, so the queued APCs were silently dropped and the wait returned without running them.

This is timing/clock dependent:

- Under the **relative-time clock** (`tick_clock(1000)` ⇒ 1 instruction = 1 ms), the 1 ms wait expires while the *first* APC's dispatcher is still executing, so the **second** APC is deterministically skipped. The `test-sample` "APC" self-test then fails (`executions != 2`) and the process exits non-zero — this is the failure seen in the Emscripten build and the Linux Python-wheel test.
- On a real wall clock it is a race the host can lose under load.

## Fix

Don't latch the timeout completion while an alertable user APC is still pending, so `switch_to_thread`'s APC-dispatch path drains the queue first.

- `emulator_thread.hpp`: add `has_pending_alertable_apc()` (`apc_alertable && !pending_apcs.empty()`).
- `emulator_thread.cpp`: guard the two timeout-completion branches in `is_thread_ready` with `&& !this->has_pending_alertable_apc()`.

## Verification

Compiles clean on Windows. The deterministic relative-time path is exercised by the gtest `EmulationTest.BasicEmulationWorks` and the Python-wheel/Emscripten CI jobs — those should confirm the APC self-test now passes.

### DIFF
--- a/src/windows-emulator/emulator_thread.hpp
+++ b/src/windows-emulator/emulator_thread.hpp
@@ -282,10 +282,6 @@ namespace sogen
             return this->await_time.has_value() && this->await_time.value() != infinite && this->await_time.value() < clock.steady_now();
         }
 
-        // A user APC queued for an alertable wait must be delivered (WAIT_IO_COMPLETION)
-        // before the wait completes by timeout. While such an APC is pending, the wait
-        // must not latch a timeout status, otherwise it suppresses APC dispatch in
-        // switch_to_thread (which skips dispatch when a status is already pending).
         bool has_pending_alertable_apc() const
         {
             return this->apc_alertable && !this->pending_apcs.empty();
--- a/src/windows-emulator/emulator_thread.hpp
+++ b/src/windows-emulator/emulator_thread.hpp
@@ -282,6 +282,15 @@ namespace sogen
             return this->await_time.has_value() && this->await_time.value() != infinite && this->await_time.value() < clock.steady_now();
         }
 
+        // A user APC queued for an alertable wait must be delivered (WAIT_IO_COMPLETION)
+        // before the wait completes by timeout. While such an APC is pending, the wait
+        // must not latch a timeout status, otherwise it suppresses APC dispatch in
+        // switch_to_thread (which skips dispatch when a status is already pending).
+        bool has_pending_alertable_apc() const
+        {
+            return this->apc_alertable && !this->pending_apcs.empty();
+        }
+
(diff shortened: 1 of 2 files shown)


momo5502:
@codex review

src/windows-emulator/emulator_thread.hpp
```diff
@@ -282,6 +282,15 @@ namespace sogen
             return this->await_time.has_value() && this->await_time.value() != infinite && this->await_time.value() < clock.steady_now();
         }
 
+        // A user APC queued for an alertable wait must be delivered (WAIT_IO_COMPLETION)
+        // before the wait completes by timeout. While such an APC is pending, the wait
+        // must not latch a timeout status, otherwise it suppresses APC dispatch in
+        // switch_to_thread (which skips dispatch when a status is already pending).
+        bool has_pending_alertable_apc() const
+        {
+            return this->apc_alertable && !this->pending_apcs.empty();
```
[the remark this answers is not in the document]
momo5502: Good catch — addressed in 4ab42912. The remove_multiple path yields with the caller's `alertable` flag, so it genuinely had the bug; guarded both I/O-completion timeout branches with `has_pending_alertable_apc()`. (remove_single currently yields non-alertable, so the guard is a no-op there today but added for consistency.)

src/windows-emulator/emulator_thread.hpp
```diff
@@ -282,6 +282,15 @@ namespace sogen
             return this->await_time.has_value() && this->await_time.value() != infinite && this->await_time.value() < clock.steady_now();
         }
 
+        // A user APC queued for an alertable wait must be delivered (WAIT_IO_COMPLETION)
+        // before the wait completes by timeout. While such an APC is pending, the wait
+        // must not latch a timeout status, otherwise it suppresses APC dispatch in
+        // switch_to_thread (which skips dispatch when a status is already pending).
```
momo5502: comment can be removed

momo5502: Done — removed in 9a4b7f9c.

PR #1023, opened 2026-06-06
Render builtin MessageBox: controls, button clicks, IDs and captions

momo5502:
## Summary

Makes the builtin `MessageBox` actually render and behave through the guest USER/GDI path. Starting from an empty window, `messagebox-sample` (`MB_YESNO | MB_ICONQUESTION`) now:

- renders the dialog with **Yes/No buttons**, their **labels**, and the **message text**
- responds to **mouse clicks** — clicking **Yes** returns `IDYES` (the sample prints `clicked: yes`)

## Background

The real user32 control wndprocs run (layout, `BeginPaint`), but produced no pixels and the dialog returned `IDOK` regardless of which button was clicked. Investigation (incl. Ghidra on the host user32.dll) traced this to three independent gaps in the emulated USER/GDI layer.

## Changes

**Control painting** — MessageBox creates its child controls during `WM_INITDIALOG` and they paint while the dialog is still hidden; user32 gates control painting on the whole parent chain being visible, so they correctly skip drawing. `ShowWindow` then only invalidated the dialog itself, never its children, so they never repainted once the dialog became visible.
- Add `invalidate_window_tree()` and invalidate the visible child subtree from `NtUserShowWindow` / `NtUserSetWindowPos(SWP_SHOWWINDOW)`.

**Button clicks** — `is_button_window` only matched the literal `"Button"` class, but builtin controls store the raw ordinal-atom class (`#1`), so click hit-testing never found the buttons. Now matches `#1` too, so clicking dismisses the dialog.

**Button IDs + captions** — user32's `SoftModalMessageBox` builds the dialog template by reading each button's `{caption, control id}` from `SERVERINFO.MBStrings` (`gpsi + 0x3A4`). Our SERVERINFO left that table zeroed, so every button got control id 0 (wrong return value) and an empty caption (blank label).
- Seed `SERVERINFO.MBStrings` with the 11 standard message-box buttons (OK, Cancel, Abort, Retry, Ignore, Yes, No, Close, Help, Try Again, Continue) after the client pfn arrays are copied.
- Removed the previous fragile caption→ID heuristic.

**New/!missing syscalls**
- Implement `NtGdiGetDCDword` (gdi32 calls it for `GdiGetIsMemDc` during control paint; was unimplemented and halted emulation).
- Implement `NtGdiPolyPatBlt` (button press/focus repaint on click).
- Stub `NtUserDrawIconEx` so the icon static finishes painting.

## Known remaining gaps

- The message-box **icon** is not drawn (`NtUserDrawIconEx` is a stub).
- The dialog **background** may stay white rather than the gray dialog face.

## Test

```
build\release\artifacts\analyzer.exe -s build\release\artifacts\messagebox-sample.exe
```
Renders the dialog with Yes/No + text; clicking Yes prints `clicked: yes`.

### DIFF
--- a/docs/windows-ui-emulation.md
+++ b/docs/windows-ui-emulation.md
@@ -33,7 +33,7 @@ Real emulation goal is stronger: builtin controls, dialogs, and custom `WM_PAINT
 - SDL is now native default backend when SDL3 is available
 - guest dialog flow works much better than before
 - `MessageBox` no longer depends on export hook hack
-- builtin control callback dispatch reaches real user32 code
+- builtin control callback dispatch reaches ntdll client thunks, but those thunks still need correct user32 client initialization before real builtin control paint can work
 - builtin `Button` / `Static` no longer crash during paint because correct extra bytes now allocated:
   - `Button` -> `cbWndExtra = 8`
   - `Static` -> `cbWndExtra = 8`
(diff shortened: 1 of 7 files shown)


momo5502:
@codex review

PR #1024, opened 2026-06-06
Resolve builtin control-class atoms

momo5502:
## Summary

Makes builtin control-class atom resolution portable across Windows builds, so dialogs (e.g. `MessageBox`) create their `Button`/`Static`/etc. child controls correctly regardless of which host `user32.dll` the emulator runs against.

## Problem

`normalize_builtin_window_class_name` hardcoded build-specific class-atom values for the static control (`#2032`/`#2160`/`#12192` → `Static`). Those atom values are assigned **per Windows build**, so a sample that resolved the static class as `#2032` on one system failed with `#12192` ("missing class") on another — the hardcoded list could never be complete.

## Root cause

From the host `user32.dll` dialog-creation worker (`FUN_18002a070`): dialog templates store controls by **system class ordinal** (`0x80`=Button, `0x81`=Edit, `0x82`=Static, `0x83`=ListBox, `0x84`=ScrollBar, `0x85`=ComboBox, written as `0xFFFF,<ord>`). user32 turns the ordinal into a class atom by indexing the `SERVERINFO.atomSysClass[ICLS]` WORD table at **`gpsi + 0x364`** with `(ordinal & 0x7F)`, then passes that atom to `NtUserCreateWindowEx`:

```c
class_atom = *(WORD *)(gpsi + (ordinal & 0x7F) * 2 + 0x364);
```

Since the table holds the build's own atom values, no static list of atom numbers can be portable.

## Fix

`resolve_builtin_class_atom` (in `syscalls/user.cpp`) reverse-maps an incoming integer class atom back to its canonical builtin class name by reading that **same** `gpsi + 0x364` table and matching the atom against the ICLS-indexed entries (0=Button, 1=Edit, 2=Static, 3=ListBox, 4=ScrollBar, 5=ComboBox). It is wired into `read_large_string_or_atom` / `read_unicode_string_or_atom` via `resolve_atom_name`. Because the lookup round-trips through the exact memory user32 used for the forward mapping, it always agrees with user32 regardless of the build's atom values — no hardcoded atoms. The build-specific `#NNNN` aliases are dropped from `normalize_builtin_window_class_name`.

Verified with a runtime probe at `NtUserCreateWindowEx`: incoming Button atom `0x1` matched `atomSysClass[0]` and incoming Static atom `0x7F0` matched `atomSysClass[2]`. The message-box dialog now creates `Button`/`Static` controls with no "missing class" errors and no `#NNNN` fallbacks.

## Also in this PR

- Remove a leftover `std::printf` debug trace from the GDI batch-flush validation path (`gdi.cpp`); the error path now bails silently like the other guards in that function.
- Documentation: `docs/windows-ui-emulation.md` updated to record the atom-resolution mechanism/fix and the earlier MessageBox rendering/click findings.

### DIFF
--- a/docs/windows-ui-emulation.md
+++ b/docs/windows-ui-emulation.md
@@ -141,37 +141,36 @@ so always analyze the DLL the emulator actually maps):
 
 - The message-box **icon** is not drawn (`NtUserDrawIconEx` stub).
 - The dialog **background** may stay white instead of the gray dialog face.
-- **System builtin-class atom resolution is not portable** (see next section).
 
-## System builtin-class atom resolution (open, build-specific)
+## System builtin-class atom resolution (resolved, portable)
 
-`normalize_builtin_window_class_name` hardcodes specific class-atom values (`#1`→Button,
-`#2032`/`#2160`/`#12192`→Static, `#32770`→dialog). This breaks across Windows builds: the same
-sample works with one system's user32 (static = `#2032`) but fails with another's (static =
-`#12192`, "missing class"). Investigation findings:
+**Fixed** in `resolve_builtin_class_atom` (`syscalls/user.cpp`). The earlier stopgap had
+`normalize_builtin_window_class_name` hardcoding build-specific class-atom values
+(`#2032`/`#2160`/`#12192`→Static); this broke across Windows builds because the same sample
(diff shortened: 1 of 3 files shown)


momo5502:
@codex review

PR #1026, opened 2026-06-06
Keep Python hook handles alive internally

### LINKED ISSUE #1025: The memory_execution_at / memory_execution functions in the Python API need to have their return values held at the Python level in order to take effect.
The memory_execution_at / memory_execution functions in the Python API need to have their return values held at the Python level in order to take effect.

As shown below:

Hook working normally:

```python
hooks_list = []

def on_module_load(module):
    if module.name.lower() == "test-sample.exe":
        h = emu.hooks.memory_execution_at(module.entry_point, lambda address: print(f"hit entry point: 0x{address:x}"))
        hooks_list.append(h)

emu.callbacks.on_module_load = on_module_load
```

Hook failed:

```python
def on_module_load(module):
    if module.name.lower() == "test-sample.exe":
        emu.hooks.memory_execution_at(module.entry_point, lambda address: print(f"hit entry point: 0x{address:x}"))

emu.callbacks.on_module_load = on_module_load
```

momo5502:
Fixes #1025.

Low-level Python hook APIs now retain hook handles inside hook_registry, so temporary return values no longer get destroyed at end of statement.

### DIFF
--- a/src/python-bindings/sogen_api_hooks.cpp
+++ b/src/python-bindings/sogen_api_hooks.cpp
@@ -92,7 +92,7 @@ namespace sogen::py
     {
     }
 
-    void hook_handle::remove()
+    void hook_handle::remove() const
     {
         if (this->shared_state)
         {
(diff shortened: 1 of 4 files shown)


momo5502:
@codex review

PR #1027, opened 2026-06-06
Improve Windows UI control emulation

momo5502:
## Summary

Improves the Windows UI emulation path for built-in controls and manual UI samples.

- Adds USER/GDI support needed for checkbox and radio control painting, including OEM button glyph drawing, control text updates, mouse capture, and brush-origin handling.
- Routes SDL mouse down/up events through child-window hit testing so controls receive input more like real child windows.
- Adds and wires a manual controls sample to exercise checkboxes, radio buttons, status text updates, and button clicks.
- Documents the background-brush/window-surface behavior used by the UI paint path.

## Validation

- `cmake --build --preset=tidy`
- `git diff --check`
- `cmd /c "cd build\release\artifacts\ && analyzer.exe -s test-sample.exe"`

Note: the same smoke command from `build\tidy\artifacts` failed before emulation because that artifacts directory did not contain `registry\SYSTEM`; release artifacts had the registry data and the smoke test passed.

### DIFF
--- a/docs/windows-ui-emulation.md
+++ b/docs/windows-ui-emulation.md
@@ -268,6 +268,49 @@ with the class background brush — gray dialog face instead of white). Still op
 - `NtGdiGetDCDword` only returns the default (0); other indices (MapMode, GraphicsMode, …) would need
   real values if a control relies on them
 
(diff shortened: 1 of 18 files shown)


momo5502:
@codex review

PR #1029, opened 2026-06-07
Re-encode WM_SETTEXT to the target wndproc's encoding

momo5502:
## Summary

Follow-up to #1027. A wide builtin control wndproc (e.g. `StaticWndProcW`) received the **ANSI** `WM_SETTEXT` payload from `SetWindowTextA` unchanged, then called `_DefSetText` with `bAnsi=0` over ANSI bytes → corrupted control text. The byte-sniffing heuristic in `read_def_set_text_string` was only masking it.

real win32k delivers a message's text to a wndproc in the proc's own encoding. This change:

- tracks each window's wndproc encoding (`window.unicode_proc` — wide for builtin controls, the registered proc otherwise);
- in `handle_NtUserMessageCall`, re-encodes the `WM_SETTEXT` string into a scratch guest buffer when the caller's encoding (`ansi`) differs from the target proc's (freed in `completion_NtUserMessageCall`);
- drops the byte-sniffing special case in `read_def_set_text_string` — `bAnsi` is now consistent at `NtUserDefSetText`, so it is trusted.

App classes registered with `RegisterClassExA` keep an ANSI proc, so an ANSI `WM_SETTEXT` matches and is not converted — no behavior change for the common case.

## Investigation

`_DefSetText(hwnd, text, bAnsi)` builds the `LARGE_STRING` from the **wndproc's own encoding**, so `bAnsi` is authoritative only if the buffer already matches the proc. Tracing `SetWindowTextA` → `NtUserMessageCall(WM_SETTEXT)` showed `ansi=1` + ANSI buffer (consistent), but the emulator forwarded the raw ANSI `l_param` to the wide `StaticWndProcW` without re-encoding — the actual bug. Full write-up in `docs/windows-ui-emulation.md` (UPDATE 2026-06-07 (3)), which also documents the capture-routing-across-top-levels fix from #1027.

## Testing

- `[defsettext]` trace for the status label flips to `bAnsi=0` + UTF-16; control text renders correctly.
- Serialization tests pass (added serialized fields to `window` and `message_call_state`).
- `analyzer -s test-sample.exe` smoke test green.

### DIFF
--- a/src/windows-emulator/windows_objects.hpp
+++ b/src/windows-emulator/windows_objects.hpp
@@ -94,7 +94,7 @@ namespace sogen
         uint64_t dialog_result{};
         emulator_pointer system_menu_ptr{};
         bool host_surface_window{};
-        bool unicode_proc{}; // wndproc expects UTF-16 text (builtin controls); false => ANSI proc
+        bool unicode_proc{};
 
         window(memory_interface& memory)
             : user_object(memory)
--- a/src/windows-emulator/windows_objects.hpp
+++ b/src/windows-emulator/windows_objects.hpp
@@ -94,6 +94,7 @@ namespace sogen
         uint64_t dialog_result{};
         emulator_pointer system_menu_ptr{};
         bool host_surface_window{};
+        bool unicode_proc{}; // wndproc expects UTF-16 text (builtin controls); false => ANSI proc
 
         window(memory_interface& memory)
             : user_object(memory)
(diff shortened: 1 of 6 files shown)


src/windows-emulator/windows_objects.hpp
```diff
@@ -94,6 +94,7 @@ namespace sogen
         uint64_t dialog_result{};
         emulator_pointer system_menu_ptr{};
         bool host_surface_window{};
+        bool unicode_proc{}; // wndproc expects UTF-16 text (builtin controls); false => ANSI proc
```
momo5502: comment is useless

PR #1044, opened 2026-06-13
win32: implement Win32 cursor/foreground input so game menus respond

momo5502:
The IW engine (MW3 stock iw5, MW2/iw4x) drives its menu cursor and gates its
per-frame input loop on the classic Win32 cursor APIs (GetCursorPos /
GetForegroundWindow), not on window messages, raw input, or DirectInput
(confirmed from iw5mp.exe's import table). Those were stubbed:
NtUserGetCursorPos returned STATUS_NOT_SUPPORTED and NtUserGetForegroundWindow
returned 0, so the engine never activated mouse input. Keyboard navigation was
also limited to the few keys whose SDL keycode happens to equal the Win32
virtual-key code, so arrow keys (menu up/down) did nothing.

- Track the cursor position (screen coordinates) and the foreground window in
  process_context, updated from pointer events in handle_ui_event.
- NtUserGetCursorPos now reports the tracked position; NtUserGetForegroundWindow
  returns the interacted-with window, falling back to a visible top-level window
  so a freshly-created game window is foreground before the first mouse event.
- Map the arrow / navigation / editing keys (SDL keycode -> Win32 VK) so they
  produce WM_KEYDOWN; menu up/down now work.


momo5502:
@codex review

src/windows-emulator/syscalls/object.cpp
```diff
@@ -596,7 +621,7 @@ namespace sogen
 
             for (ULONG i = 0; i < count; ++i)
             {
-                const auto h = handles.read(i);
+                const auto h = resolve_wait_handle(c, handles.read(i));
```
Igoorx: From what I found when looking at this code before, we aren't actually supposed to resolve pseudo-handles in WaitForMultipleObjects: https://doxygen.reactos.org/d0/daf/modules_2rostests_2winetests_2kernel32_2sync_8c_source.html
```
    /* in contrast to WaitForSingleObject, pseudo handles are not allowed in
     * WaitForMultipleObjects and NtWaitForMultipleObjects */
```

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
```diff
@@ -54,7 +54,55 @@ namespace sogen
                 return static_cast<uint64_t>(key);
             }
 
-            return 0;
+            // Non-ASCII keys whose SDL keycode does not coincide with the Win32 virtual-key code: map them
+            // explicitly so menu navigation (arrows) and editing keys produce the expected WM_KEYDOWN.
+            switch (key)
```
momo5502, line 57: can we clean this up a bit? define the VK_ constants in window.hpp, along the other VK_ definitions?

momo5502: also maybe add support for a few more keys? e.g. tilde, etc.?

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
```diff
@@ -540,6 +625,39 @@ namespace sogen
                 const auto it = this->guest_by_window_id_.find(window_id);
                 return it == this->guest_by_window_id_.end() ? 0 : it->second;
             }
+
+            // Tell the guest a window became (in)active. SDL only reports focus transitions on real OS focus
+            // changes, which need not happen for the embedded game window, so activation is also asserted on
+            // the first mouse interaction. Latched so a held/moving cursor doesn't re-post every frame.
+            // WM_ACTIVATE's wParam is WA_ACTIVE/WA_INACTIVE with the minimized flag (HIWORD) left at 0.
```
momo5502: remove the comment

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
```diff
@@ -469,6 +549,11 @@ namespace sogen
                                                   surface.width, surface.height);
                 if (state.texture)
                 {
+                    // The presented frame is opaque; its alpha channel holds whatever the guest's backbuffer
+                    // happened to leave there (D3D9/Vulkan ignore backbuffer alpha at present time). Without
+                    // disabling blending, SDL composites the texture over the cleared background using that
+                    // alpha and darkens the whole image.
```
momo5502: remove the comment. many of the comments in this PR are useless

src/windows-emulator/memory_manager.cpp
```diff
@@ -422,7 +422,13 @@ namespace sogen
...
     bool memory_manager::decommit_memory(const uint64_t address, const size_t size)
     {
-        const auto entry = this->find_reserved_region(address);
+        // NtFreeVirtualMemory operates on whole pages: the base rounds down and the end rounds up. Without
+        // this the committed regions get split at an unaligned offset and unmap_memory is handed a
+        // non-page-aligned range (which the WHP backend rejects).
+        const auto aligned_start = page_align_down(address);
+        const auto end = page_align_up(address + size);
+
+        const auto entry = this->find_reserved_region(aligned_start);
```
momo5502: i don't feel like this is a good change. if the exact base is not passed, there is a bug somewhere else

src/windows-emulator/syscalls/object.cpp
momo5502: Good catch — fixed in 20c9f8eb. The previous `is_object_pseudo_handle` check ran *after* `resolve_wait_handle` had already expanded the pseudo handle, so it never actually fired. The raw handle is now checked before resolving and a pseudo handle returns `STATUS_INVALID_HANDLE` here, while `NtWaitForSingleObject` keeps resolving them as Windows does.

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
momo5502: Done in 20c9f8eb — defined the `VK_*` codes in window.hpp next to the existing ones and `map_sdl_keycode` now uses the symbolic names instead of bare hex.

momo5502: Added in 20c9f8eb — grave/tilde (`VK_OEM_3`) plus the printable OEM keys (`- = [ ] \ ; ' , . /`).

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
momo5502: Removed in 20c9f8eb.

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
momo5502: Removed in 20c9f8eb.

src/windows-emulator/memory_manager.cpp
momo5502: Agreed. Reverted the rounding inside `decommit_memory` so it gets the exact, already page-aligned range. The NT page rounding (base down, end up — which `NtFreeVirtualMemory` does for `MEM_DECOMMIT`) now lives in the syscall handler instead, so the exact base reaches `decommit_memory`. 20c9f8eb.

PR #1047, opened 2026-06-14
windows-emulator: make instruction precision configurable

momo5502:
## Summary

Makes per-instruction precision configurable in the Windows emulator.

A new `emulator_settings::use_instruction_precision` flag (enabled by default) gates the fine-grained, per-instruction behavior:

- instruction-quantum thread scheduling (the `MAX_INSTRUCTIONS_PER_TIME_SLICE` round-robin quantum), and
- exact instruction-pointer tracking via the cached `current_ip`.

The **effective** precision is `use_instruction_precision && emu().supports_instruction_counting()`, exposed via `windows_emulator::uses_instruction_precision()`. So precision is automatically disabled when either the option is turned off **or** the backend cannot count instructions (e.g. WHP) — without the caller having to know which backend is in use.

## Behavior when precision is off

This routes counting-capable backends (Unicorn/Icicle) through the same paths WHP already relied on:

- the global per-instruction hook (`on_instruction_execution`) is **not** registered;
- thread scheduling falls back to the wall-clock interrupt thread in `start()` (~1s preemption) instead of the instruction quantum;
- the instruction pointer is read live via `read_instruction_pointer()` instead of the cached `current_ip` (DbgPrint IP fixup and `dispatch_exception`).

## Notes

- All previous `supports_instruction_counting()` call sites in the windows-emulator layer now go through `uses_instruction_precision()`.
- Snapshot serialization is unchanged: the flag is derived from settings + backend at construction time.
- Built locally (`windows-emulator` target) — compiles and links cleanly.

### DIFF
--- a/src/windows-emulator/windows_emulator.hpp
+++ b/src/windows-emulator/windows_emulator.hpp
@@ -98,6 +98,12 @@ namespace sogen
         bool disable_logging{false};
         bool use_relative_time{false};
 
+        // Enables fine-grained, per-instruction precision: instruction-quantum thread
+        // scheduling and exact instruction-pointer tracking. Enabled by default, but
+        // automatically disabled when the backend does not support instruction counting
+        // (e.g. the WHP backend). See windows_emulator::uses_instruction_precision().
+        bool use_instruction_precision{true};
+
         std::filesystem::path emulation_root{};
         std::filesystem::path registry_directory{"./registry"};
 
@@ -218,6 +224,15 @@ namespace sogen
             return this->executed_instructions_;
         }
 
+        // True when per-instruction precision is active: the configured
+        // use_instruction_precision setting AND the backend supports instruction
+        // counting. When false, thread scheduling falls back to coarse, wall-clock
+        // based preemption and the instruction pointer is read live from the backend.
+        bool uses_instruction_precision() const
+        {
+            return this->instruction_precision_;
+        }
+
         stop_reason last_stop_reason() const
         {
             return this->last_stop_reason_;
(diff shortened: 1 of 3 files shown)


src/windows-emulator/windows_emulator.hpp
```diff
@@ -218,6 +224,15 @@ namespace sogen
             return this->executed_instructions_;
         }
 
+        // True when per-instruction precision is active: the configured
+        // use_instruction_precision setting AND the backend supports instruction
+        // counting. When false, thread scheduling falls back to coarse, wall-clock
+        // based preemption and the instruction pointer is read live from the backend.
```
momo5502: remove the comment

src/windows-emulator/windows_emulator.hpp
```diff
@@ -98,6 +98,12 @@ namespace sogen
         bool disable_logging{false};
         bool use_relative_time{false};
 
+        // Enables fine-grained, per-instruction precision: instruction-quantum thread
+        // scheduling and exact instruction-pointer tracking. Enabled by default, but
+        // automatically disabled when the backend does not support instruction counting
+        // (e.g. the WHP backend). See windows_emulator::uses_instruction_precision().
```
momo5502: remove the comment

src/windows-emulator/windows_emulator.hpp
```diff
@@ -98,6 +98,12 @@ namespace sogen
         bool disable_logging{false};
         bool use_relative_time{false};
```
momo5502: remove that space

PR #1052, opened 2026-06-15
Fix some Unicode filesystem issues

Igoorx:
This PR fixes issues related to Unicode characters in filenames and file paths:

* Ensure Unicode characters aren’t lost in arguments or file paths on Windows.
* Correctly account for Unicode filenames in `compat_stat`.
* Ensure `translate_path` doesn’t corrupt Unicode characters.

### DIFF
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -6,7 +6,7 @@
 #include <windows.h>
 #endif
 
-#include "emulator-platform/platform/unicode.hpp"
+#include <emulator-platform/platform/platform.hpp>
 
 namespace sogen
 {
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -4,6 +4,7 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#include <shellapi.h>
 #endif
 
 #include <emulator-platform/platform/platform.hpp>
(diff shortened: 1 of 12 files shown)


src/analyzer/main.cpp
```diff
@@ -19,6 +21,59 @@ namespace
     }
 }
 
+#ifdef _WIN32
+std::string wide_to_utf8(const std::wstring& w)
```
momo5502: might make sense to port that to src/tools/sandbox/main.cpp as well

momo5502: and maybe extract that function into common

momo5502: actually, i don't think this is needed, setlocale(LC_UTF8) or so should work, afaik?

Igoorx: I think `setlocale` would only affect the console, since it would happen after we are already given the argument array. The issue here is that when Unicode goes through the ANSI entry point, it becomes `????`.

src/analyzer/main.cpp
```diff
@@ -1,9 +1,13 @@
 #include <cstdlib>
+#include <string>
+#include <vector>
 
 #ifdef _WIN32
 #include <windows.h>
 #endif
 
+#include "emulator-platform/platform/unicode.hpp"
```
momo5502:
```suggestion
#include <emulator-platform/platform/unicode.hpp>
```

momo5502 (approved):

Igoorx:
~~Huh, that’s a curious CI failure, since that code was supposedly compiling before. I guess I will just fix it.~~

Ah, got it. I guess I should include platform.hpp instead, since it has the directives to suppress the warnings.

PR #1058, opened 2026-06-15
gpu-bridge: forward VK_EXT_memory_budget per-heap budget/usage

momo5502:
## What

`vkGetPhysicalDeviceMemoryProperties2` in the guest Vulkan shim ignored its `pNext` chain, so a chained `VkPhysicalDeviceMemoryBudgetPropertiesEXT` was left zero-initialized. As a result `IDXGIAdapter3::QueryVideoMemoryInfo` reported **0 bytes** of available video memory, which DXVK surfaces as `Budget: 0 MiB` for every heap.

## Change

Add a `get_physical_device_memory_budget` bridge command:

- **host** (`vulkan_host`): queries the real device via `vkGetPhysicalDeviceMemoryProperties2` chained with `VkPhysicalDeviceMemoryBudgetPropertiesEXT` and returns per-heap budget/usage.
- **shim**: `vkGetPhysicalDeviceMemoryProperties2` now walks the `pNext` chain and fills any `VkPhysicalDeviceMemoryBudgetPropertiesEXT` from the bridge response.
- **protocol**: new request/response structs (fixed `VK_MAX_MEMORY_HEAPS`-sized arrays).

DXVK now reports the real per-heap VRAM budget instead of `0 MiB`.

## Testing

- `cmake --build --preset=tidy` (clang-tidy clean) and clang-format applied.
- Release smoke test (`analyzer -s test-sample.exe`) passes.
- Verified against The Witcher 3 (DXVK): budget now reads e.g. `3.32 GiB / 23.12 GiB / 213 MiB` instead of `0 MiB`.

### DIFF
--- a/src/analyzer/CMakeLists.txt
+++ b/src/analyzer/CMakeLists.txt
@@ -2,6 +2,7 @@ add_executable(analyzer main.cpp)
 
 if(WIN)
   target_sources(analyzer PRIVATE resource.rc)
+  target_link_libraries(analyzer PRIVATE shell32)
 endif()
 
 sogen_assign_source_group(main.cpp)
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -1,8 +1,13 @@
 #include <cstdlib>
+#include <string>
+#include <vector>
 
 #ifdef _WIN32
 #include <windows.h>
+#include <shellapi.h>
 #endif
+
+#include <emulator-platform/platform/platform.hpp>
 
 namespace sogen
 {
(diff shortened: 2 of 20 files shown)


one-eyed-queen:
A niche question, so sogen now uses proxy/shim dll kind of setup for emulation which translates things for the emulated process, so is this method better than doing:-
Direct Memory Thunking (Same Process Translation)if the guest application and the host system share the same CPU architecture (eg- x86_64 windows app running on x86_64 linux), a separate host process is not required.
Lockless Shared Memory Ring Buffers (Zero Copy IPC)if isolation requiremen mandate a separate host process for stability or security, standard pipes or sockets should be avoided.
Asynchronous Command Batching and lastly, Adopting an Established Industry Protocol (Mesa Venus)Instead of inventing a custom serialization protocol, the project could adopt open source standards designed exactly for this use case?

momo5502:
There is no separeate host process. Sogen fully emulates the applications within it's own address space.
So regarding your points:

1. Direct thunking: the host already reads/writes guest memory directly, there is no IPC to skip.
2. Shared-memory ring buffer: there's no IPC to replace, it's already in-process.
3. Async batching: already implemented.
4. Mesa Venus: I actually started out modelling based on Venus, there is even codegen that parses vk.xml the same way. However, handgenerating the marshalling code allowed make progress faster and get COD working, without supporting what would have been needed for Venus. So, hand-marshalling was just faster short-term, Venus would probably be better on the long run.

Considering that most of the code was AI generated in a short amount of time, it's not a big deal to switch at any point. However, at the moment, there are other features that are more important.

one-eyed-queen:
Oh okay cool, thanks for the information, and yeah nice work, I actually am using sogen way to emulate a minimal kernel system instead of a user space for my educational research on denuvo DRM(which is in my repo and i made it a web page like work) so i was interested in this, thanks for information. Also would be cool if u had discrod too, take care and happy working

momo5502:
Sure I'm momo5502 on discord

one-eyed-queen:
Cool added ya, thanks you

PR #1059, opened 2026-06-16
Emulator fixes surfaced by CoD MW3 / open-iw5 (IOCP, LSA, win32k, thread enumeration)

momo5502:
Four independent emulator-side fixes surfaced while bringing up Call of Duty: Modern Warfare 3 (via open-iw5) under the emulator.

## io-completion: consume target signal when delivering a wait packet
Delivering a wait-completion-packet completion is a satisfied wait, so it must consume the target's signal exactly like `NtWaitForSingleObject` (auto-reset event resets, semaphore decrements). Without this an edge-triggered target stays signaled and floods the completion port with spurious packets — the game's async scheduler busy-spins forever.

## lsa-policy-lookup: answer opnum 5 instead of failing the RPC
The shell's account-lookup path (`SHGetKnownFolderPath`, hit while the engine sets up its crash-report queue) issues an additional LsaLookup call (opnum 5). We can't resolve it, but returning an RPC error makes the runtime raise `RPC_S_CALL_FAILED`, whose vectored handler re-enters a half-initialized singleton and deadlocks. Answer with a well-formed "none mapped" reply instead.

## win32k: stub NtUserDisableProcessWindowsGhosting
Window ghosting is a desktop-compositor feature with no meaning in the emulator; accept the request so engine startup proceeds.

## system-process-info: implement SystemProcessInformation for thread enumeration
`NtQuerySystemInformation(SystemProcessInformation)` returned `STATUS_NOT_SUPPORTED`, which breaks `CreateToolhelp32Snapshot(TH32CS_SNAPTHREAD)`. MinHook freezes all other threads before patching and enumerates them through that snapshot; when it fails, `MH_EnableHook` returns `MH_ERROR_MEMORY_ALLOC` and the detour is never written — so **every MinHook-based detour silently no-ops** under the emulator. For open-iw5 this left the GSC builtin dispatch table empty, crashing on the first script builtin (`spawnstruct`) during map load. Return the current process and its live threads so the snapshot succeeds and MinHook can install its detours.

## Testing
- `analyzer.exe -s test-sample.exe` passes.
- MW3 (open-iw5) now loads multiplayer maps in the emulator.

### DIFF
--- a/src/emulator-platform/platform/process.hpp
+++ b/src/emulator-platform/platform/process.hpp
@@ -1152,6 +1152,65 @@ namespace sogen
         typename Traits::PVOID DefaultBase;
     };
 
+    template <typename Traits>
+    struct SYSTEM_THREAD_INFORMATION
+    {
+        LARGE_INTEGER KernelTime;
+        LARGE_INTEGER UserTime;
+        LARGE_INTEGER CreateTime;
+        ULONG WaitTime;
+        typename Traits::PVOID StartAddress;
+        CLIENT_ID<Traits> ClientId;
+        LONG Priority;
+        LONG BasePriority;
+        ULONG ContextSwitches;
+        ULONG ThreadState;
+        ULONG WaitReason;
+    };
+
+    template <typename Traits>
+    struct SYSTEM_PROCESS_INFORMATION
+    {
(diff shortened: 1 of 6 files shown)


momo5502:
@codex review

PR #1075, opened 2026-06-19
Shadow Vulkan mappings to avoid slow WHP memory access

Igoorx:
While testing Sogen’s new GPU bridge, I got a game running but found that directly mapped Vulkan memory caused extremely poor performance under WHP. This PR instead shadows it with regular guest RAM and synchronizes it when needed, making the game substantially more playable, though not yet completely smooth.

*Note:* The only test done after this change was running the aforementioned game.

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -4,6 +4,7 @@
 #include "../windows_emulator.hpp"
 
 #include <gpu_bridge_protocol.hpp>
+#include <new>
 
 namespace sogen
 {
@@ -228,6 +229,8 @@ namespace sogen
             }
 
           private:
+            static constexpr size_t direct_mapping_alignment = 0x1000;
+
             vulkan_host vulkan_{};
 
             // Guest-visible RAM shadows mapped VkDeviceMemory so CPU accesses remain normal WHP memory.
@@ -1369,9 +1372,8 @@ namespace sogen
                 uint64_t host_size = 0;
                 response.vk_result = this->vulkan_.map_memory(request.device, request.memory, host_ptr, host_size);
 
-                constexpr uint64_t page = 0x1000;
-                if (response.vk_result != 0 || !host_ptr || host_size == 0 || (reinterpret_cast<uintptr_t>(host_ptr) % page) != 0 ||
-                    (host_size % page) != 0)
+                if (response.vk_result != 0 || !host_ptr || host_size == 0 ||
+                    (reinterpret_cast<uintptr_t>(host_ptr) % direct_mapping_alignment) != 0 || (host_size % direct_mapping_alignment) != 0)
                 {
                     if (host_ptr)
                     {
@@ -1380,8 +1382,9 @@ namespace sogen
                     return write_output(win_emu, context, response);
                 }
 
-                const uint64_t mapped_size = (host_size + page - 1) & ~(page - 1);
-                void* shadow_ptr = VirtualAlloc(nullptr, static_cast<size_t>(mapped_size), MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE);
+                const uint64_t mapped_size = (host_size + direct_mapping_alignment - 1) & ~(direct_mapping_alignment - 1);
+                void* shadow_ptr =
+                    ::operator new(static_cast<size_t>(mapped_size), std::align_val_t{direct_mapping_alignment}, std::nothrow);
                 if (!shadow_ptr)
                 {
                     this->vulkan_.unmap_memory(request.device, request.memory);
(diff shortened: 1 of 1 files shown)


momo5502:
interesting. i need to check :D

momo5502:
i quickly tested with MW2, it crashes in NtDeviceIoControlFile when loading a map, but the main menu feels extremely laggy now, compared to main

Igoorx:
Darn. I will get MW2 to see if I can figure out what could be causing the crash and whether the lag is just something like `sync_all_direct_mappings_to_host` being too heavy to call on every queue submission. Otherwise, I might just give up on submitting this patch and leave it as an open issue instead.

momo5502:
managed to fix the ioctl crash, MW2 is now at about 3-5 FPS inage, vs 20-30 FPS on main

momo5502:
> Darn. I will get MW2 to see if I can figure out what could be causing the crash and whether the lag is just something like `sync_all_direct_mappings_to_host` being too heavy to call on every queue submission. Otherwise, I might just give up on submitting this patch and leave it as an open issue instead.

I'm using IW4x btw. makes it easier to work with, as much of the steam stuff has been patched. if you want i can send you the files.

momo5502:
Found a consistent host-side crash with this PR while bringing up iw4x/MW2 (32-bit, WoW64) under the WHP backend + GPU bridge. The game dies deterministically at the **first `vkQueueSubmit2`** during renderer init:

```
Syscall NtDeviceIoControlFile threw an unknown exception: 0x7 (raw: 0x1B0007)
```

(`0x7` is the syscall id; the build is `/EHa`, so the `catch(...)` is actually catching a **host access violation** inside our own C++, not a guest fault.)

### Root cause

The new shadow mechanism's `sync_all_direct_mappings_to_host()` `memcpy`s each mapping's shadow into its `host_ptr` on every submit. But `handle_free_memory()` calls `vulkan_.free_memory()` (→ `vkFreeMemory`, which **implicitly unmaps** the memory and invalidates `host_ptr`) **without removing the `direct_mappings_` entry**:

```cpp
NTSTATUS handle_free_memory(...) {
    ...
    this->vulkan_.free_memory(request.device, request.memory); // host mapping invalidated
    return STATUS_SUCCESS;                                      // ...but direct_mappings_ entry kept!
}
```

So once a guest frees a `VkDeviceMemory` that was directly mapped (legal Vulkan — freeing implicitly unmaps, no explicit `vkUnmapMemory` required), its `direct_mappings_` entry lingers with a **dangling `host_ptr`**, and the next `queue_submit2` → `sync_all_direct_mappings_to_host()` `memcpy`s into freed host memory → AV.

Smoking gun: `direct_mappings_` only ever **grew** during the run (8 → 10 → 30, never shrank on free), unlike `handle_unmap_memory_direct()` which cleans up correctly.

Note: the missing cleanup in `handle_free_memory` predates this PR, but it was **latent/benign on `main`** because nothing dereferenced the leaked `host_ptr`. This PR makes it load-bearing — `sync_all_direct_mappings_to_host()` is the first thing to write into every mapping's `host_ptr` per submit. So this crash is specific to this branch; `main` doesn't have it.

### Fix

Tear down the direct mapping in `handle_free_memory` before `vkFreeMemory`, mirroring `handle_unmap_memory_direct` (minus the copy-to-host, since the memory is being discarded):

```cpp
// Freeing a VkDeviceMemory implicitly unmaps it, so drop any direct mapping first: otherwise
// its host pointer dangles and the next sync_all_direct_mappings_to_host() copies into freed memory.
if (const auto it = this->direct_mappings_.find(request.memory); it != this->direct_mappings_.end())
{
    win_emu.memory.release_memory(it->second.guest_address, static_cast<size_t>(it->second.size));
    if (it->second.shadow_ptr)
    {
        VirtualFree(it->second.shadow_ptr, 0, MEM_RELEASE);
    }
    this->direct_mappings_.erase(it);
}

this->vulkan_.free_memory(request.device, request.memory);
```

With this, iw4x/MW2 boots cleanly through renderer init to a normal shutdown (`Emulation terminated with status: 0`) under WHP + GPU bridge, where it previously died at the first submit every time.

Igoorx:
> I'm using IW4x btw. makes it easier to work with, as much of the steam stuff has been patched. if you want i can send you the files.

Yeah, that would help, thanks.

momo5502:
> > I'm using IW4x btw. makes it easier to work with, as much of the steam stuff has been patched. if you want i can send you the files.
> 
> Yeah, that would help, thanks.

Got discord or so? So I can send you the stuff?

I'm momo5502 on discord

Igoorx:
>Got discord or so? So I can send you the stuff?

I had one, but I forgot the password and couldn’t find it anywhere 😭. So I made a new one. I sent you a friend request.

PR #1079, opened 2026-06-20
Fix open-iw5 (MW3) single-player shutdown freeze

momo5502:
Closing open-iw5 (MW3) single-player froze the emulator deterministically — a guest thread waited on a wakeup that never fired, so the cooperative scheduler idled forever. Two fixes, found by dumping the guest thread wait-states at the hang.

## 1. Wait on a reaped thread handle never completes (`767c6f28`)
The shutdown thread does `WaitForSingleObject` on a worker thread's handle (a join). Once that worker had terminated **and been reaped** from the thread store, `observe_object_signal` / `consume_object_signal` looked the thread up, found nothing, and reported **not-signaled** — so the join hung forever.

On Windows a terminated thread's handle stays **signaled**, even after the object is gone. Treat a missing (reaped) thread the same as a terminated one in both `observe_object_signal` and `consume_object_signal`. With this, the join completes and the shutdown thread proceeds to `NtTerminateProcess(NULL)` to tear down the rest.

## 2. `NtUserGetObjectInformation` stub (`34efaf21`)
After the join was fixed, shutdown reached `NtUserGetObjectInformation` (called during cleanup), which was a known-but-unimplemented syscall and halted the emulator. Stubbed to fail (`FALSE`) for the USER objects the emulator does not model, consistent with the other UI stubs.

## Result
open-iw5 SP now shuts down cleanly on close (`Emulation terminated with status: 3`, the game's own exit code) instead of hanging. Verified end-to-end: launch → load a map → close → prompt clean exit.

Diagnostic instrumentation used to find this (a deadlock watchdog that dumps thread wait-states) is intentionally **not** included.

### DIFF
--- a/src/windows-emulator/emulator_thread.cpp
+++ b/src/windows-emulator/emulator_thread.cpp
@@ -148,7 +148,9 @@ namespace sogen
 
             case handle_types::thread: {
                 const auto* t = c.threads.get(h);
-                if (t && t->is_terminated())
+                // A reaped thread is gone from the store, but its handle stays signaled (it terminated),
+                // so a missing thread counts the same as a terminated one.
+                if (!t || t->is_terminated())
                 {
                     return wait_state::signaled;
                 }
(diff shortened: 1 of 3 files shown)


momo5502:
@claude review

PR #1084, opened 2026-06-20
Fix async presentation of static frames in GPU-bridge

Igoorx:
This PR fixes asynchronous presentation by polling for completed GPU readbacks in the device work loop and presenting them immediately, so completed static frames appear without requiring another guest present. This was a particularly noticeable issue in apps that did not re-render until something on the screen changed.

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -18,6 +18,18 @@ namespace sogen
         // in snapshots yet; restoring with an open GPU handle is an experimental limitation.
         struct gpu_bridge_device : io_device
         {
+            void work(windows_emulator& win_emu) override
+            {
+                for (auto& frame : this->vulkan_.poll_presented_frames())
+                {
+                    win_emu.ui().present_surface(static_cast<hwnd>(frame.hwnd), ui_surface_desc{.width = static_cast<int>(frame.width),
+                                                                                                .height = static_cast<int>(frame.height),
+                                                                                                .stride = static_cast<int>(frame.width * 4),
+                                                                                                .format = ui_surface_format::bgra8,
+                                                                                                .pixels = frame.pixels.data()});
+                }
+            }
+
             NTSTATUS io_control(windows_emulator& win_emu, const io_device_context& context) override
             {
                 switch (context.io_control_code)
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -22,11 +22,7 @@ namespace sogen
             {
                 for (auto& frame : this->vulkan_.poll_presented_frames())
                 {
-                    win_emu.ui().present_surface(static_cast<hwnd>(frame.hwnd), ui_surface_desc{.width = static_cast<int>(frame.width),
-                                                                                                .height = static_cast<int>(frame.height),
-                                                                                                .stride = static_cast<int>(frame.width * 4),
-                                                                                                .format = ui_surface_format::bgra8,
-                                                                                                .pixels = frame.pixels.data()});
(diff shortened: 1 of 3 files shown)


momo5502:
@claude review

momo5502:
Claude's dead 🥀

src/windows-emulator/devices/gpu_bridge.cpp
```diff
@@ -18,6 +18,18 @@ namespace sogen
...
         {
+            void work(windows_emulator& win_emu) override
+            {
+                for (auto& frame : this->vulkan_.poll_presented_frames())
+                {
+                    win_emu.ui().present_surface(static_cast<hwnd>(frame.hwnd), ui_surface_desc{.width = static_cast<int>(frame.width),
+                                                                                                .height = static_cast<int>(frame.height),
+                                                                                                .stride = static_cast<int>(frame.width * 4),
+                                                                                                .format = ui_surface_format::bgra8,
+                                                                                                .pixels = frame.pixels.data()});
```
momo5502: `handle_queue_present` only presents when `hwnd_value != 0 && !pixels.empty()`, but this path presents every frame `poll_presented_frames` returns unconditionally. `poll_presented_frames` always copies `sc.hwnd` (which comes straight from the guest via `create_surface`) into `frame.hwnd`, so a guest that created a surface with a null `hwnd` would reach `present_surface` with a 0 window here. SDL no-ops on an unresolved window so it is harmless today, but it would be good to keep the guard consistent between the two present paths (either guard here or filter inside `poll_presented_frames`).

src/windows-emulator/devices/vulkan_host.cpp
```diff
@@ -3671,11 +3663,57 @@ namespace sogen
...
+            }
+
+            auto& frame = frames.emplace_back();
+            frame.pixels.resize(static_cast<size_t>(readback_size));
+            std::memcpy(frame.pixels.data(), mapped, frame.pixels.size());
+            dev.unmap_memory(dev.handle, sc.readback_memory);
+            frame.width = sc.width;
+            frame.height = sc.height;
+            frame.hwnd = sc.hwnd;
+            sc.present_in_flight = false;
```
momo5502: This readback-map-copy block is a near-exact duplicate of `queue_present` part 1 (lines ~3604-3614), and the `present_surface` call in `gpu_bridge_device::work` duplicates `handle_queue_present`. Consider extracting a small helper (e.g. `bool drain_readback(swapchain_data&, device_data&, presented_frame&)`) used by both `poll_presented_frames` and `queue_present`, so the layout/size/mapping logic lives in one place. Right now a change to the readback format would have to be made in two spots.

src/windows-emulator/devices/vulkan_host.cpp
```diff
@@ -3671,11 +3663,57 @@ namespace sogen
...
         return VK_SUCCESS;
     }
 
+    std::vector<vulkan_host::presented_frame> vulkan_host::poll_presented_frames()
+    {
+        std::vector<presented_frame> frames{};
+
+        for (auto& [id, sc] : this->impl_->swapchains)
+        {
+            (void)id;
```
momo5502: Nit: `for (auto& [id, sc] : ...)` followed by `(void)id;` can be replaced with `for (auto& sc : this->impl_->swapchains | std::views::values)`, which matches the idiom used elsewhere in the codebase (e.g. `perform_context_switch_work`, `get_thread_by_id`).

momo5502 (commented):
Reviewed the async-present change. The approach is sound: `work()` runs once per context-switch on the single emulation thread (`perform_context_switch_work` -> `dev.work`), so there is no data race with `io_control`, and `present_in_flight` gates each readback so a frame is published exactly once. The static-final-frame case is now handled (the last submitted copy gets drained by the poll loop instead of waiting for a present that never comes), and as a nice side effect the present is no longer a full frame behind. A few smaller points inline.

momo5502 (approved):

PR #1085, opened 2026-06-20
Fix GDI Handle Table and get LoadCursor working

Igoorx:
This PR implements several things required for the `LoadCursor` GDI API to succeed, fixes a regression in GDI APIs introduced by the process ID change in #1082, and some other general emulation improvements:

- **Fixes GDI handle ownership by storing the correct process ID in the GDI handle table.**
- **Improves memory DC and bitmap behavior in `NtGdiCreateCompatibleDC`, `NtGdiSelectBitmap`, and `NtGdiGetDCforBitmap`.**
- Stubs `NtUserCreateEmptyCursorObject`, `NtUserSetCursorIconDataEx`, and `NtUserGetRequiredCursorSizes`.
- Implements cursor state tracking through `NtUserSetCursor` and `NtUserGetCursor`.
- Adds support for `NtGdiSetPixel`, `NtGdiGetPixel`.
- Adds stubs for `NtGdiCreatePaletteInternal`, `NtGdiCreateHalftonePalette`, `NtGdiDoPalette`, and `NtGdiCreateDIBSection`.
- Improves `NtGdiGetGlyphOutline` so it atleast writes a visible dot glyph.
- Initializes the `processId`, `threadId`, and `dpiContext` fields for window objects, and sets the default DPI awareness context in `Win32ClientInfo`.
- Implements `NtUserGetKeyNameText` with scan-code-based key-name lookup.
- Adds stubs for `NtUserTranslateAccelerator`, `NtUserSelectPalette`, `NtUserSwapMouseButton`, and `NtUserWindowFromPoint`.


Igoorx:
Hm... I guess it could be a good idea to add `LoadCursorFromFile` to the tests as a GDI test, but I’ll need to make changes so at least one cursor file is copied to the emulation root.

momo5502:
@claude review

src/windows-emulator/syscalls/gdi.cpp
```diff
@@ -1753,18 +1812,20 @@ namespace sogen
             return handle_value;
         }
 
+        uint64_t handle_NtGdiCreateDIBSection()
+        {
+            return STATUS_SUCCESS;
```
momo5502: `CreateDIBSection` returns an `HBITMAP`, and `STATUS_SUCCESS == 0`, so this stub actually hands back a **NULL bitmap = failure**. If the intent is to fail loudly until it's implemented, `return 0;` with a short comment would be clearer; using `STATUS_SUCCESS` as a handle value reads as success but behaves as failure. (Same naming concern doesn't apply to `NtGdiDoPalette`, which legitimately returns NTSTATUS.)

src/windows-emulator/syscalls/user.cpp
```diff
@@ -1906,14 +1906,34 @@ namespace sogen
...
-        uint64_t handle_NtUserGetCursor()
+        hcursor handle_NtUserGetCursor(const syscall_context& c)
         {
-            return 0;
+            return c.proc.current_cursor;
+        }
+
+        hicon handle_NtUserCreateEmptyCursorObject()
+        {
+            return make_pseudo_handle(0x100, handle_types::reserved).bits;
```
momo5502, line 1921: Every call returns the same fixed pseudo-handle `0x100`, so all empty cursor objects are indistinguishable. Combined with the no-op `SetCursorIconData[Ex]`, that's fine for the LoadCursor path today, but will break anything tracking multiple distinct cursors. Worth a brief comment noting the limitation.

src/windows-emulator/syscalls/user.cpp
```diff
@@ -1906,14 +1906,34 @@ namespace sogen
...
+        }
+
+        BOOL handle_NtUserSetCursorIconDataEx()
+        {
+            return TRUE;
+        }
+
+        BOOL handle_NtUserGetRequiredCursorSizes()
+        {
+            return TRUE;
```
momo5502, line 1936: `NtUserGetRequiredCursorSizes` returns a size/count, not a BOOL, so a caller reading the result gets `1`. Harmless for now since LoadCursor doesn't seem to rely on it, just flagging the semantic mismatch.

src/windows-emulator/syscalls/gdi.cpp
```diff
@@ -3939,6 +4061,67 @@ namespace sogen
 
             return STATUS_SUCCESS;
         }
+
+        COLORREF handle_NtGdiSetPixel(const syscall_context& c, const hdc dc, const int x, const int y, const COLORREF color)
+        {
+            constexpr uint64_t clr_invalid = 0xFFFFFFFF;
+
+            gdi_dc_state* dc_state = nullptr;
```
momo5502, line 4144: `dc_state` is written by `get_dc_state_and_surface` but never read here (only `surface` is checked), unlike the other call sites that check `!dc_state`. Likely a clang-tidy nudge candidate.

src/windows-emulator/syscalls/gdi.cpp
```diff
@@ -3939,6 +4061,67 @@ namespace sogen
...
+            if (!get_surface_pixel(*surface, surface_x, surface_y).has_value())
+            {
+                return clr_invalid;
+            }
+
+            const auto bgra = colorref_to_bgra(color);
+            set_surface_pixel(*surface, surface_x, surface_y, bgra);
+
+            uint32_t present_handle = 0;
+            resolve_dc_surface(c, dc, origin_x, origin_y, present_handle);
```
momo5502: `resolve_dc_surface` is called again purely to obtain `present_handle`, but `get_dc_state_and_surface` above already resolved the surface (it calls `resolve_dc_surface` internally). Minor redundant lookup; could thread `present_handle` out of the first call.

momo5502 (commented):
Reviewed the GDI/cursor bring-up. Handle-ownership fix and memory-DC/SelectBitmap semantics look correct, and the struct-layout static asserts are great discipline. A few notes left inline, all minor / stub-level. Nice work.

src/windows-emulator/syscalls/gdi.cpp
Igoorx: Instead of just fixing the return type, I’ll add a slightly more complete implementation, because further experiments showed that this stub can cause a null pointer dereference.

src/windows-emulator/syscalls/user.cpp
Igoorx: I don't think this matters, since at a glance it's clear that this is just a stub.

src/windows-emulator/syscalls/user.cpp
Igoorx: This is probably not true, the only use of `NtUserGetRequiredCursorSizes` is as a BOOL.
image

src/windows-emulator/syscalls/gdi.cpp
Igoorx: I don't think this matters.

momo5502 (approved):

PR #1086, opened 2026-06-21
Fix gaps in posted messages

Igoorx:
The other day, I was trying to fix an issue in a game running on Sogen, and one of my suspicions was that the missing values for the `pt` and `time` fields in messages that Sogen posts to the guest were the cause... Well, the issue was ultimately caused by #1084, but these were still good changes anyway, so I'm making this PR.

This PR also includes one other change: message coalescing, mostly for `WM_MOUSEMOVE` messages. According to the docs, this appears to be something Windows does as well.

### DIFF
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -29,12 +29,12 @@ namespace sogen
 
             virtual ~clock() = default;
 
-            virtual system_time_point system_now()
+            virtual system_time_point system_now() const
             {
                 return std::chrono::system_clock::now();
             }
 
-            virtual steady_time_point steady_now()
+            virtual steady_time_point steady_now() const
             {
                 return std::chrono::steady_clock::now();
             }
@@ -66,12 +66,12 @@ namespace sogen
                 }
             }
 
-            system_time_point system_now() override
+            system_time_point system_now() const override
             {
                 return this->now(this->system_start_);
             }
 
-            steady_time_point steady_now() override
+            steady_time_point steady_now() const override
             {
                 return this->now(this->steady_start_);
             }
@@ -81,7 +81,7 @@ namespace sogen
                 return this->ticks();
             }
 
-            virtual uint64_t ticks() = 0;
+            virtual uint64_t ticks() const = 0;
 
             uint64_t get_frequency() const
             {
@@ -94,7 +94,7 @@ namespace sogen
             steady_time_point steady_start_{};
 
             template <typename TimePoint>
-            TimePoint now(const TimePoint start)
+            TimePoint now(const TimePoint start) const
             {
                 using duration = typename TimePoint::duration;
 
--- a/src/common/utils/time.hpp
+++ b/src/common/utils/time.hpp
@@ -29,12 +29,12 @@ namespace sogen
 
             virtual ~clock() = default;
 
-            virtual system_time_point system_now() const
+            virtual system_time_point system_now()
             {
                 return std::chrono::system_clock::now();
             }
 
-            virtual steady_time_point steady_now() const
+            virtual steady_time_point steady_now()
             {
                 return std::chrono::steady_clock::now();
             }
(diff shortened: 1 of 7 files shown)


momo5502:
@claude review

momo5502:
@codex review

momo5502:
Fuck

momo5502 (commented):
## Review

Reviewed the branch locally: both affected translation units (`emulator_thread.cpp`, `user.cpp`, `windows_emulator.cpp`) compile cleanly. The build doesn't fully link in my environment, but only because of an unrelated icicle **Rust** dependency failure (`cargo` exit 101) — this PR touches no Rust.

### What's done well
- **Centralizing `time`/`pt` in `post_message` is the right call** — one source of truth instead of every caller filling them. The time base stays consistent with `GetTickCount`: both derive from `clock().steady_now()` milliseconds (`kusd_mmio.cpp`), so message-timing math remains coherent.
- **`current_message_time` is wired into every retrieval path**: synchronous `NtUserGetMessage`/`NtUserPeekMessage`, the blocking `await_msg` path in `is_thread_ready`, and `DispatchMessage`. No obvious gap.
- **Coalescing correctly excludes `WM_INPUT`** (`try_coalesce` defaults false). Since raw-input deltas are delivered via `deliver_raw_mouse_input` *before* the cosmetic `WM_MOUSEMOVE` coalesces, mouse-look games don't lose motion. Nice handling of that interaction.
- **`const`-correctness on `clock`/`ticks()`** is the proper enabler. Only one concrete `ticks()` override exists and it was updated.
- **Serialization updated** for `current_message_time` — snapshot compatibility preserved.

### Minor points (none blocking)
1. **Coalescing skips `queue_status_changed_bits`.** A coalesced message returns early without OR-ing the changed bit, and `NtUserGetQueueStatus` clears those on read — so the high-word "changed since last call" edge is lost for a coalesced move. Negligible in practice (the low word still reports `QS_MOUSEMOVE` and the thread is already awake), but worth a conscious note.
2. **`DispatchMessage` sets `current_message_time`.** Real Windows `GetMessageTime` reflects the last *Get/Peek*, not `DispatchMessage`. Harmless in the normal Get->Dispatch loop; only a peek(NOREMOVE)-then-dispatch-other sequence would briefly differ.
3. **`NtUserGetThreadState` routine `0x9`** is an RE-derived constant (the neighbor `dialog_state` is `0xA`). Plausible, but worth confirming against the targeted Windows build's win32k.
4. **Redundant caller-side fields:** `post_message` now overwrites `time`/`pt` unconditionally, so leftovers like `.time = 0, .pt = {}` (e.g. the `WM_PAINT` post) are now dead. Cosmetic.
5. **Style:** stray blank line inside the anonymous namespace after `get_current_message_time`, and the `now_ms` line is long — `clang-format` will tidy both.

### Design note
Swapping `(process_context&, clock&)` for `windows_emulator&` across these queue methods increases coupling — they now depend on the whole emulator rather than just what they use. Reasonable trade since `post_message` genuinely needs process (cursor) + clock together, just noting the direction.

Overall: solid, well-scoped fix with accurate Windows semantics. LGTM after optionally confirming the `0x9` routine value.

src/common/utils/time.hpp
```diff
@@ -81,7 +81,7 @@ namespace sogen
                 return this->ticks();
             }
 
-            virtual uint64_t ticks() = 0;
+            virtual uint64_t ticks() const = 0;
```
momo5502: i deliberately did not make these const, as an implementation may want to change state when ticks is called. it's not needed at the moment, but who knows, might be.

momo5502: any reason this is const now?

Igoorx: Makes sense, I can revert that. I made this change only because I didn’t want to pass windows_emulator as a mutable reference unnecessarily.

PR #1089, opened 2026-06-21
Skyrim bringup: child-window present, vkCmdClearAttachments, raw mouse/keyboard input

momo5502:
Three fixes from Skyrim (TESV.exe, 32-bit WoW64, DXVK D3D9) bringup.

## Changes

- **ui/sdl: composite child render windows onto their top-level ancestor.** Skyrim renders to a `WS_CHILD` render window whose parent is the visible top-level. The SDL backend only gives top-level windows a renderer, so `present_surface` on the child dropped every frame and the top-level showed its gray empty-surface fallback. Present now redirects to the top-level ancestor.

- **gpu-bridge: implement `vkCmdClearAttachments`.** Was the missing shim function (guest shim record → protocol command → host replay → `vulkan_host`), without which DXVK's render pass recording failed.

- **input: deliver raw mouse buttons and keyboard via `WM_INPUT`.** Extends the existing raw mouse-motion path: `RegisterRawInputDevices` now also accepts keyboard registrations (HID usage 0x06), and `GetRawInputData` returns a `RAWKEYBOARD` or `RAWMOUSE` body depending on the pending token — including mouse button transitions (`RI_MOUSE_*` in `usButtonFlags`) and key make/break with a best-effort US scan code. Games such as Skyrim read buttons/keystrokes through `WM_INPUT` rather than window messages.

## Testing

- Release build + smoke test (`analyzer -s test-sample.exe`) pass.
- With the present + clear-attachments fixes, Skyrim's main menu renders.
- The raw-input change is not yet exercisable in-game (a separate timing-induced deadlock stalls Skyrim before input is read); it is additive and gated on raw-input registration.

### DIFF
--- a/src/emulator-platform/platform/user.hpp
+++ b/src/emulator-platform/platform/user.hpp
@@ -169,6 +169,17 @@ namespace sogen
     };
     static_assert(sizeof(RAWMOUSE32) == 0x18);
 
+    struct RAWKEYBOARD32
+    {
+        uint16_t MakeCode;
+        uint16_t Flags;
+        uint16_t Reserved;
+        uint16_t VKey;
+        uint32_t Message;
+        uint32_t ExtraInformation;
+    };
+    static_assert(sizeof(RAWKEYBOARD32) == 0x10);
+
     enum USER_HANDLETYPE : uint8_t
     {
         TYPE_FREE = 0,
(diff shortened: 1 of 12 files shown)


momo5502:
@claude review

PR #1090, opened 2026-06-22
Fix GDI rendering regression

Igoorx:
This PR fixes a regression introduced by https://github.com/momo5502/sogen/pull/1085 that caused GDI windows to stop drawing. The issue was that the PR made `NtGdiCreateCompatibleDC` more compliant, but `NtUserGetDC` was relying on its previous non-compliant behavior.


momo5502:
@claude review

Issue #1045, opened 2026-06-14
Add FEX backend


momo5502:
https://fex-emu.com/
Would be nice to have, for better android support.

momo5502:
Box moved to here: #1094, as FEX is probably the more interesting one

PR #1091, opened 2026-06-22
Fixes for Sogen GPU paravirtualization on Android

Igoorx:
This PR fixes three small issues I encountered while trying to run games on Sogen on Android. I tested it with a lightweight 2D game called [Touhou 8](https://en.wikipedia.org/wiki/Imperishable_Night), and it works :D, albeit excruciatingly slowly (it took 10 minutes to load).

I think Sogen could even become a real alternative backend to projects like [Winlator](https://github.com/brunodev85/winlator) once https://github.com/momo5502/sogen/issues/1045 is resolved.

https://github.com/user-attachments/assets/0140c13d-c20d-48ee-860f-22c905f9de12


momo5502:
Holy shit. that's awesome. i gotta tweet about this :D you got twitter so i can tag you? 😂

Igoorx:
Nah, I don’t use Twitter. Feel free to mention me as just "Igoorx on GitHub".

momo5502:
@claude review

momo5502:
https://x.com/momo5502/status/2069283520183181330

Also, one important note regarding performance is that all backends, other than WHP, still have the per-instruction hook.
It's super expensive. I have started adding an option to get rid of it with unicorn (this should give enormous speedups).

https://github.com/momo5502/sogen/tree/fast-unicorn

however, it still causes issues, mainly due to preemptive scheduling being broken. seems like stopping unicorn from a separate thread causes races, despite the author claiming it being threadsafe :(

https://github.com/unicorn-engine/unicorn/issues/292#issuecomment-162004369
https://github.com/unicorn-engine/unicorn/issues/636

Igoorx:
>Also, one important note regarding performance is that all backends, other than WHP, still have the per-instruction hook.
It's super expensive. I have started adding an option to get rid of it with unicorn (this should give enormous speedups).

I noticed that too. I tried enabling the preemptive scheduling you added, and it indeed didn’t work. But disabling just the on_instruction callback already made the frame rate about twice as good, and that’s actually what we can see in the video :)

PR #1093, opened 2026-06-25
Add Linux KVM emulator backend

### LINKED ISSUE #1071: Add KVM backend

momo5502:
Adds a Linux KVM backend, alongside the existing Unicorn/WHP/Icicle backends. It is a standalone backend mirroring `whp-emulator` (WHP is untouched) and is selected at runtime via `EMULATOR_KVM=1` (Linux x86_64 only; gated out on Android).

## How it works

- Runs guest code natively in a KVM vCPU set up for long mode (4-level page tables, GDT/TSS, CPL3 user mode).
- **Syscalls**: `MSR_LSTAR` points at a one-`HLT` page; the `HLT` exit reconstructs the syscall context and dispatches it.
- **Exceptions**: KVM has no userspace exception-exit mechanism (unlike WHP's exit bitmap), so guest exceptions are trapped via a synthetic IDT whose gates point at one-`HLT` stubs running at CPL0 on an IST stack (via a TSS). The run loop derives the vector from the trapping RIP and rebuilds the faulting frame.
- **MMIO** (e.g. `KUSER_SHARED_DATA`): backed by a read-only memslot seeded from the read callback and refreshed before each guest entry, so reads execute natively (KVM's in-kernel instruction emulator lacks SSE) and only writes trap.
- **Preemption**: a wall-clock time slice that resets `immediate_exit` each slice and signal-kicks the vCPU thread to break an in-flight `KVM_RUN`.

## Testing

- `test-sample.exe` runs to completion with all 26 tests passing under KVM on both **Intel and AMD** hosts.
- Adds a `Smoke Test KVM x86_64` CI job (ubuntu-24.04 with `/dev/kvm`), mirroring the WHP smoke test.

The trickiest issues were AMD-specific and in the exception path: clearing pending `kvm_vcpu_events` after handling an exception (the KVM analog of the WHP AMD fix in #808, otherwise the stale event is re-injected into a triple fault on SVM), and rewinding RIP by one for the int3 breakpoint trap so the dispatcher reports the right `ExceptionAddress`. `src/backends/kvm-emulator/NOTES.md` has the full write-up.

Closes #1071


src/backend-selection/backend_selection.hpp
```diff
@@ -12,6 +12,7 @@ namespace sogen
         unicorn,
         icicle,
         whp,
+        kvm,
```
momo5502, line 15: this should be handled in the python bindings as well

momo5502: also landing page (in page folder; "Pluggable Backends") and readme should probably be adjusted to reflect KVM support

momo5502:
@claude review

docs/kvm-backend.md
momo5502, line 1: this document should be moved to docs

momo5502:
@claude review

PR #1095, opened 2026-06-26
KVM backend: faster memory mapping, host-memory support, and symlink mounts

momo5502:
A set of KVM-backend improvements (plus one backend-agnostic file-system fix) that make the KVM backend fast enough to run real applications, including a GPU game, on Linux.

## Performance: incremental memslot rebuild

`rebuild_mappings` used to tear down **every** KVM memslot and recreate them from scratch on each `map_memory`/`unmap`/`protect`/internal-page allocation. Every `KVM_SET_USER_MEMORY_REGION` forces an NPT/TLB rebuild, so this was O(total mappings) per memory operation — quadratic over a load. It now diffs the desired layout against the installed slots and only touches what changed. On a memory-heavy workload (loading a game's assets) this is **~385× faster** (6 → 2308 files loaded in a fixed 30s window).

## Performance: register caching

Single-register access did a full `KVM_GET_REGS`/`KVM_SET_REGS` ioctl each time. The general and segment register files are now cached per exit: reads fetch once, writes mark dirty, and the run loop flushes once before `KVM_RUN` and invalidates after.

## GPU paravirtualization: `map_host_memory`

Implemented `map_host_memory` (it previously threw "not supported"), so the GPU bridge can alias host buffers directly into the guest via KVM memslots. This unblocks the GPU path.

## File system: host symlinks in the emulation root

`file_system::translate` confined guest paths with `weakly_canonical` + `relative`, both of which follow symlinks — so a symlink inside the root pointing elsewhere (e.g. `c:/game` → a real install) resolved outside the root and collapsed to the drive root. The containment check is now lexical: `..` traversal is still blocked, but host-placed symlinks are followed by the OS at open time, like a bind-mount.

## Testing

`test-sample.exe` passes all 26 tests under KVM on Intel and AMD. With these changes a 32-bit game (`open-iw5`/MW3) loads its assets, reaches the GPU paravirtualization path, and renders.

@claude review

### DIFF
--- a/src/backends/kvm-emulator/kvm_x86_64_common.hpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_common.hpp
@@ -30,7 +30,6 @@ namespace sogen::kvm::detail
     struct mapped_page
     {
         void* host_page = nullptr;
-        uint32_t map_flags = 0;
         memory_permission permissions = memory_permission::none;
         std::shared_ptr<uint8_t> owned_page{};
     };
--- a/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
@@ -355,10 +355,12 @@ namespace sogen::kvm
                     }
 
                     this->refresh_mmio_pages();
+                    this->flush_register_cache();
 
                     this->run_active_ = true;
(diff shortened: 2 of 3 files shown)


momo5502:
@claude review

momo5502:
@claude review

PR #1097, opened 2026-06-27
kvm: fix GPU buffer cache coherency and shutdown segfault for host-aliased memory

momo5502:
Two KVM-backend fixes for the GPU paravirtualization path, found running Call of Duty: MW3 (open-iw5) under KVM on a Steam Deck (AMD Van Gogh / RADV).

## 1. Cache coherency for host-aliased GPU buffers

DXVK allocates `HOST_VISIBLE | HOST_COHERENT` memory that RADV backs with **write-combined** memory, while the KVM backend aliases those physical pages into the guest as **write-back cacheable**. The mismatched cache attributes mean the guest's cached vertex/texture writes are not guaranteed visible to the GPU until they happen to evict, which corrupted streamed geometry — intermittent jumbled glyphs and vertex noise in the menus, worse during asset streaming, fine once in-game. WHP is unaffected because Windows aliases the host memory coherently.

Fix: add a coherency capability to `memory_interface` — backends report whether host-memory aliasing is coherent with external devices and expose a cache-flush hook. The KVM backend reports non-coherent aliasing and flushes the aliased ranges via `clflush`; the GPU bridge flushes every directly-aliased buffer before each queue submit. Other backends keep the coherent default and skip the work entirely (no cost on WHP/unicorn).

An alternative considered was promoting the host allocation to a `HOST_CACHED` (write-back) sibling type, but DXVK's main streaming allocations are `DEVICE_LOCAL | HOST_VISIBLE` (heap with no cached sibling on the APU), so a flush is the only approach that covers all aliased memory regardless of type.

## 2. Use-after-free / segfault on shutdown

`handle_free_memory` freed the `VkDeviceMemory` without releasing the guest alias or erasing the `direct_mappings_` entry. After the guest freed a directly-aliased allocation it kept a stale alias of freed memory, and the bridge retained a dangling host pointer that the pre-submit cache flush then dereferenced — segfaulting while DXVK frees buffers during shutdown. Fix mirrors `handle_unmap_memory_direct`: tear the alias down and forget the host pointer before `vkFreeMemory`.

## Testing
- Verified on the Steam Deck: menu renders cleanly, closing the game no longer crashes.
- Windows release smoke suite (`analyzer -s test-sample.exe`): all 26 tests pass — confirms the shared `memory_interface` change doesn't affect the WHP/unicorn paths.

@claude review

### DIFF
--- a/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
@@ -11,7 +11,7 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
-#include <emmintrin.h>
+#include <immintrin.h>
 
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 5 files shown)


momo5502:
@claude review

PR #1102, opened 2026-06-27
Implement window restoration when loading snapshots

Igoorx:
This PR restores windows that were open when an emulator snapshot was created. Not many changes were necessary, since we already serialize everything :D

After these changes, apps that create windows are restored correctly, and you can even restore software-rendered games:

https://github.com/user-attachments/assets/e9e9b58f-5fa1-4769-ac52-d5b6809b0163

One future objective could be to make these "save states" work for games that use the GPU as well, though that would require saving and restoring GPU state.

### DIFF
--- a/page/src/web-ui-host.ts
+++ b/page/src/web-ui-host.ts
@@ -661,6 +661,15 @@ export function attachSogenUiHost(
       return;
     }
 
+    if (message.command === "reset") {
+      windows.clear();
+      focusedWindow = 0;
+      viewportPinnedByUser = false;
+      notifyWindowCount();
+      composite();
+      return;
+    }
+
     const hwnd = message.hwnd ?? 0;
     if (!hwnd) {
       return;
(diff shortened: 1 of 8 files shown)


momo5502:
@claude review

momo5502:
https://x.com/momo5502/status/2071088837057532158

Igoorx:
> `SDL_SetHint(SDL_HINT_NO_SIGNAL_HANDLERS, "1")` — unrelated and unexplained

This is related because the only way to generate a snapshot is by hitting Ctrl+C, but without that line, the signal is swallowed by SDL.

> Raw pointers into `process.windows` during restore loop
> These pointers are stable as long as `this->process.windows` is not mutated during the loop.

I doubt the UI backend will ever be able to create process-aware windows. And even if it does, as far as I know, `std::map` does not invalidate pointers on insertion, since it isn't a sequence container.

PR #1105, opened 2026-06-28
analyzer: parse the command line with CLI11

momo5502:
Replaces the hand-rolled argument parsing in the **analyzer**, **windows-analyzer**, **linux-analyzer** and **sandbox** with [CLI11](https://github.com/CLIUtils/CLI11) (added as a submodule at v2.5.0, header-only, BSD-3 → GPLv2-compatible).

## Why

- **Unicode/UTF-8 on Windows** — `App::ensure_utf8()` resolves the UTF-8 arguments from the wide command line directly, so the awkward `CommandLineToArgvW` + `wchar_t`→UTF-8 conversion in `analyzer/main.cpp` and `sandbox/main.cpp` is gone. The dispatcher is now a plain `int main(int, char**)`.
- **Trailing-argument passthrough** — `prefix_command()` stops parsing at the first positional (the application to analyze) and forwards everything after it verbatim to the emulated program, including guest flags like `-console -stdout`.
- **CMake** — one-line submodule integration via an interface target (`cli11`).

## Breaking change

The non-standard multi-character *single-dash* options are removed in favor of their long names. CLI11 (like every getopt-style parser) combines single-dash short flags, so `-vc` would mean `-v -c`; keeping both was impossible.

| Removed | Use instead |
|---|---|
| `-vc` | `--very-concise` |
| `-ss` | `--skip-syscalls` |
| `-rep` | `--reproducible` |
| `-fx` | `--first-exec` |
| `-is` | `--inst-summary` |
| `-bc` | `--break-call` |
| `-nip` | `--no-inst-precision` |

All single-character short options (`-d -s -v -b -x -f -c -m -e -a -t -i -p -r`) are unchanged. `page/src/settings.ts` now emits long forms for every switch.

## Verification

- Release build clean; net −390 lines.
- Smoke test (`analyzer -s test-sample.exe`): 26/26 pass.
- Passthrough: `analyzer --very-concise test-sample.exe -console -stdout` forwards `-console -stdout` to the guest with zero parse errors.
- Backend selection intact: `--backend whp`/`unicorn` select the expected engine; invalid value errors with `--backend: bogus not in {unicorn,icicle,whp,kvm}`.
- `-p src dst` / `--env name value` pairs parse without swallowing the target positional.
- `sandbox test-sample.exe` runs.

Builds on #1104 (merged).

https://claude.ai/code/session_01EK8LEh461LrzTUwgZctt84

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,3 +50,6 @@
 	path = deps/Vulkan-Headers
 	url = https://github.com/KhronosGroup/Vulkan-Headers.git
 	shallow = true
+[submodule "deps/CLI11"]
+	path = deps/CLI11
+	url = https://github.com/CLIUtils/CLI11.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -53,3 +53,4 @@
 [submodule "deps/CLI11"]
 	path = deps/CLI11
 	url = https://github.com/CLIUtils/CLI11.git
+	shallow = true
--- a/src/analyzer/main.cpp
+++ b/src/analyzer/main.cpp
@@ -1,13 +1,4 @@
 #include <cstdlib>
-#include <string>
-#include <vector>
-
-#ifdef _WIN32
-#include <windows.h>
-#include <shellapi.h>
-#endif
-
-#include <emulator-platform/platform/platform.hpp>
 
 namespace sogen
 {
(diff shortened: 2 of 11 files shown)


src/analyzer/main.cpp
```diff
@@ -24,52 +15,10 @@ namespace
...
-    LocalFree(argv);
-    return result;
-}
-#endif
-
-#ifndef _WIN32
-int dispatch_main(int argc, char** argv)
+// The analyzer is built as a console application, so main() receives the process arguments directly.
+// Unicode handling (including the Windows wide command line) is performed by each analyzer through
+// CLI11's ensure_utf8(), so no platform-specific argument conversion is needed here.
```
momo5502: the comment can be removed

src/tools/sandbox/main.cpp
```diff
@@ -94,29 +89,29 @@ namespace sogen::sandbox
...
+        int run_main(int argc, char** argv)
         {
-            try
-            {
-                std::vector<std::string> args{};
-                for (int i = 1; i < argc; ++i)
-                {
-                    args.emplace_back(w_to_u8(wargv[i]));
-                }
+            CLI::App app{"Sogen sandbox"};
```
momo5502:
```suggestion
            CLI::App app{"Sogen Sandbox"};
```

src/linux-analyzer/main.cpp
```diff
@@ -31,91 +33,39 @@ namespace sogen
...
-        }
+    }
 
-        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays)
-        cli_options parse_args(const int argc, char** argv)
-        {
-            cli_options opts{};
+    int run_main(int argc, char** argv)
+    {
+        CLI::App app{"Sogen Linux analyzer"};
```
momo5502:
```suggestion
        CLI::App app{"Sogen Linux Analyzer"};
```

.gitmodules
```diff
@@ -50,3 +50,6 @@
 	path = deps/Vulkan-Headers
 	url = https://github.com/KhronosGroup/Vulkan-Headers.git
 	shallow = true
+[submodule "deps/CLI11"]
+	path = deps/CLI11
+	url = https://github.com/CLIUtils/CLI11.git
```
momo5502, line 55: add shallow = true

momo5502:
@claude review

PR #1107, opened 2026-06-28
Assassin's Creed bring-up: NtQueryDirectoryFile FileId class, NtQuerySecurityPolicy, SystemCodeIntegrityInformation

momo5502:
A series of small syscall/feature gaps found while bringing up **Assassin's Creed** (32-bit, WHP backend). Each one was an `emu.stop()` on an unimplemented path; the game now progresses through startup well past all three.

### 1. `NtQueryDirectoryFile`: `FileIdBothDirectoryInformation` (0x25)
Only `FileDirectory`/`FileFullDirectory`/`FileBothDirectory` were handled. Added the `FILE_ID_BOTH_DIR_INFORMATION` struct and routed the class through the existing `handle_file_enumeration<T>` template (extra `ShortName`/`FileId` fields zero-fill, like the other classes).

### 2. `NtQuerySecurityPolicy` — stub
`wldp.dll` calls it during startup. Stubbed to `STATUS_NOT_SUPPORTED` (no lockdown policy configured), matching the existing `NtAdjustPrivilegesToken` stub.

### 3. `NtQuerySystemInformation`: `SystemCodeIntegrityInformation` (0x67)
Returns a `SYSTEM_CODEINTEGRITY_INFORMATION` reporting a normal retail configuration (code integrity enabled, test-signing off) so anti-tamper checks see a non-tampered system.

## Result on Assassin's Creed (WHP)
- Before: stopped at the directory `FileId` query.
- After all three: runs for 45s+ without a hard stop, loading the crypto/networking stack (`rsaenh`, `cryptnet`, `iphlpapi`, `nsi`, …). The next item is a soft, non-fatal `BAD PORT \RPC Control\keysvc` warning (unemulated RPC key-service port), not a crash.

## Verification
- Release build clean.
- Smoke test 26/26 (incl. `Dir I/O`) after each change — no regressions.

### DIFF
--- a/src/emulator-platform/platform/file_management.hpp
+++ b/src/emulator-platform/platform/file_management.hpp
@@ -526,6 +526,25 @@ namespace sogen
         char16_t FileName[1];
     } FILE_BOTH_DIR_INFORMATION, *PFILE_BOTH_DIR_INFORMATION;
 
+    typedef struct _FILE_ID_BOTH_DIR_INFORMATION
+    {
+        ULONG NextEntryOffset;
+        ULONG FileIndex;
+        LARGE_INTEGER CreationTime;
+        LARGE_INTEGER LastAccessTime;
+        LARGE_INTEGER LastWriteTime;
+        LARGE_INTEGER ChangeTime;
+        LARGE_INTEGER EndOfFile;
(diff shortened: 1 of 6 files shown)


momo5502:
@claude review

PR #1108, opened 2026-06-28
ci: ship DXVK + Vulkan shim in the emulation root

momo5502:
## What

Builds the guest `vulkan-1.dll` shim for **both 32- and 64-bit** in CI and provisions the emulation root with **DXVK + the shim**, so the uploaded root (and `root.zip`) can run Direct3D titles through the GPU bridge out of the box.

## Changes

- **`build-apiset-dumper`** now also builds `vulkan-shim`. To get both architectures from a single `release` tree, it builds x86 first, stashes the DLL, wipes the tree, then builds x64 (`dump-apiset` + `vulkan-shim`). Both are uploaded as `vulkan-shim-x64.dll` / `vulkan-shim-x86.dll`.
- **`src/tools/install-gpu-runtime.ps1`** (new) downloads the latest DXVK release, overwrites the bundled `d3d9/d3d10core/d3d11/dxgi` DLLs (`x64`→`system32`, `x32`→`syswow64`), and installs the shim as `vulkan-1.dll` in both directories.
- **`create-emulation-root`** runs that script right after the API-set dump, so every uploaded root ships the GPU runtime.
- **README** credits DXVK, since its binaries are redistributed in `root.zip` (it's not a submodule, so it isn't covered by any in-tree LICENSE).

## Note

This makes DXVK the default D3D path for **every** emulation root produced by CI — D3D9/10/11 now resolve to DXVK and Vulkan routes through the bridge shim. That matches the request ("overwrite the DirectX DLLs"), just flagging the blast radius. Easy to gate behind a flag if you'd rather opt in.

### DIFF
--- a/.github/workflows/ci-reusable.yml
+++ b/.github/workflows/ci-reusable.yml
@@ -131,8 +131,8 @@ jobs:
       - name: Verify Page Formatting
         run: cd page && npx --yes prettier . --check
 
-  build-apiset-dumper:
-    name: Build API Set Dumper
+  build-root-tools:
+    name: Build Root Tools
     runs-on: windows-latest
     steps:
       - name: Checkout Source
@@ -180,7 +180,7 @@ jobs:
   create-emulation-root:
     name: Create Emulation Root
     runs-on: ${{ matrix.runner }}
-    needs: [build-apiset-dumper]
+    needs: [build-root-tools]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


momo5502:
@claude review

.github/workflows/ci-reusable.yml
momo5502: maybe we should rename this?

momo5502: we should also rename the job

PR #1111, opened 2026-06-28
gdi: support 24bpp DIBs in NtGdiSetDIBitsToDeviceInternal

momo5502:
`NtGdiSetDIBitsToDeviceInternal` only accepted 32bpp `BI_RGB` DIBs and dropped everything else, logging `unsupported DIB`. 24bpp `BI_RGB` is common — e.g. game splash screens. Unreal Tournament III blits its splash as a 24bpp 572px bitmap, so the splash window stayed blank/gray.

This accepts 24bpp in addition to 32bpp:
- compute the DWORD-aligned scanline stride from `bit_count` instead of assuming 4 bytes/pixel,
- read 3 bytes (BGR) per pixel for 24bpp (4 for 32bpp), forcing opaque alpha.

## Verification
- Release build clean; smoke test 26/26 (incl. GDI test).
- UT3 (WHP): the `NtGdiSetDIBitsToDeviceInternal: unsupported DIB (bpp=24 …)` warning is gone — the splash now blits instead of being dropped.

### DIFF
--- a/src/windows-emulator/syscalls/gdi.cpp
+++ b/src/windows-emulator/syscalls/gdi.cpp
@@ -2084,7 +2084,7 @@ namespace sogen
             c.emu.read_memory(info + 16, &compression, sizeof(compression));
 
             constexpr uint32_t bi_rgb = 0;
-            if (bit_count != 32 || compression != bi_rgb || bi_width <= 0)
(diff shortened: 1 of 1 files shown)


momo5502:
Superseded by #1113 (combined with the clipboard stubs).

PR #1112, opened 2026-06-28
syscalls: stub clipboard open/close/empty/get/set

momo5502:
The clipboard NtUser syscalls were unimplemented, so any app touching the clipboard during startup stopped emulation. Unreal Tournament III (via wxWidgets) does this on launch — it hit `NtUserOpenClipboard`, then `NtUserConvertMemHandle` (the `SetClipboardData` write path).

Adds minimal stubs alongside the existing `NtUserGetClipboardSequenceNumber`:
- `NtUserOpenClipboard` / `NtUserCloseClipboard` / `NtUserEmptyClipboard` → success
- `NtUserGetClipboardData` → no data (empty clipboard)
- `NtUserConvertMemHandle` / `NtUserSetClipboardData` → non-null placeholder so `SetClipboardData` succeeds

Contents are not stored (reads report an empty clipboard), which is sufficient for apps that only probe or set the clipboard during init. A full implementation would back these with per-process clipboard storage.

## Verification
- Release build clean; smoke test 26/26.
- UT3 (WHP): no longer stops at the clipboard syscalls — it now proceeds well past them (loads OpenAL/PhysX/dbghelp) before hitting an unrelated later issue.

### DIFF
--- a/src/windows-emulator/syscalls.cpp
+++ b/src/windows-emulator/syscalls.cpp
@@ -638,6 +638,12 @@ namespace sogen
         uint32_t handle_NtUserGetDoubleClickTime();
         BOOL handle_NtUserModifyWindowTouchCapability();
         uint32_t handle_NtUserGetClipboardSequenceNumber();
+        BOOL handle_NtUserOpenClipboard();
+        BOOL handle_NtUserCloseClipboard();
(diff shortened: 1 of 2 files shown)


momo5502:
Superseded by #1113 (combined with the 24bpp DIB fix).

PR #1113, opened 2026-06-28
windows-emulator: 24bpp DIB blits + clipboard syscall stubs

momo5502:
Two small Windows-API gaps found bringing up Unreal Tournament III (combines #1111 and #1112).

### gdi: 24bpp DIBs in `NtGdiSetDIBitsToDeviceInternal`
The handler only accepted 32bpp `BI_RGB` DIBs, dropping everything else. 24bpp `BI_RGB` is common (e.g. game splash screens — UT3's splash is a 24bpp 572px bitmap), so the splash window stayed blank. Now accepts 24bpp too: DWORD-aligned stride and 3 bytes (BGR) per pixel.

### syscalls: clipboard open/close/empty/get/set stubs
The clipboard NtUser syscalls were unimplemented, so any app touching the clipboard at startup stopped emulation (UT3 via wxWidgets hits `NtUserOpenClipboard` then the `SetClipboardData` write path). Adds minimal stubs:
- `NtUserOpenClipboard`/`NtUserCloseClipboard`/`NtUserEmptyClipboard` → success
- `NtUserGetClipboardData` → no data
- `NtUserConvertMemHandle`/`NtUserSetClipboardData` → non-null placeholder so `SetClipboardData` succeeds

Contents aren't stored (reads report an empty clipboard) — sufficient for apps that only probe or set the clipboard.

## Verification
- Release build clean; smoke test 26/26 (incl. GDI + Dir I/O).
- UT3 (WHP): splash blits (no more `unsupported DIB` warning) and it no longer stops at the clipboard syscalls.

Supersedes #1111 and #1112.

### DIFF
--- a/src/emulator-platform/platform/window.hpp
+++ b/src/emulator-platform/platform/window.hpp
@@ -175,6 +175,13 @@ namespace sogen
 #define WS_DISABLED          0x08000000L
 #define WS_CLIPSIBLINGS      0x04000000L
 #define WS_CLIPCHILDREN      0x02000000L
+#define WS_CAPTION           0x00C00000L
+#define WS_BORDER            0x00800000L
+#define WS_DLGFRAME          0x00400000L
+#define WS_SYSMENU           0x00080000L
+#define WS_THICKFRAME        0x00040000L
+#define WS_MINIMIZEBOX       0x00020000L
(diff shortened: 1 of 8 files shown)


momo5502:
@claude review

PR #1114, opened 2026-06-28
gdi: blit 24bpp DIB sections and StretchDIBits sources

momo5502:
Splash screens (and other 24bpp bitmaps) rendered empty. The standard splash path is `LoadImage`/`CreateDIBSection` → app draws pixels into the section → `BitBlt` to the window. Two GDI paths silently dropped 24bpp:

### DIB-section guest↔host sync
`CreateDIBSection` allocates **guest** memory the app draws into, separate from the surface's host pixel buffer. `sync_surface_from_guest_dib` / `sync_surface_to_guest_dib` bridge the two, but assumed 32bpp: `row_bytes = width*4` and an early `return` when `guest_stride < row_bytes` — always true for 24bpp (stride ≈ width*3). So the host surface stayed zeroed and `BitBlt` copied an empty image.

Fix: record the section's bit depth on the surface (`guest_bpp`) and convert 24bpp (BGR) rows to/from the host 32bpp surface. 32bpp keeps its fast `memcpy` path.

### StretchDIBits
`NtGdiStretchDIBitsInternal` accepted only 4/32bpp; added a 24bpp case (used by software-present paths and some splashes).

## Verification
- Release build clean; smoke test 27/27.
- Logically: 24bpp DIB sections now sync instead of bailing; needs a visual splash check on a game.

Complements #1113, which covers the direct `NtGdiSetDIBitsToDeviceInternal` 24bpp blit path; together these cover the common 24bpp bitmap routes.

### DIFF
--- a/src/windows-emulator/process_context.hpp
+++ b/src/windows-emulator/process_context.hpp
@@ -91,6 +91,7 @@ namespace sogen
 
         emulator_pointer guest_bits{};
         uint32_t guest_stride{};
+        uint32_t guest_bpp{32};
         bool guest_top_down{};
         bool guest_owns_memory{};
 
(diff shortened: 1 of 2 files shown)


momo5502:
Folded into #1113 so all 24bpp bitmap fixes ship in one PR.

PR #1122, opened 2026-06-29
Implement NtUserSetSystemTimer and NtUserGetIconInfo

momo5502:
## Summary

Implements two previously-unimplemented win32u syscalls. Each one was aborting emulation when first hit.

### `NtUserSetSystemTimer`

Sets an internal *system* timer that posts `WM_SYSTIMER` messages and, unlike `NtUserSetTimer`, takes no timer callback proc.

- **`syscalls/user.cpp`**: Extract a shared `set_user_timer` helper; `handle_NtUserSetTimer` (proc, non-system) and the new `handle_NtUserSetSystemTimer` (no proc, system) both delegate to it.
- **`emulator_thread.hpp`**: Add an `is_system` flag to `user_timer` (serialized) and an `is_system` parameter to `create_user_timer`.
- **`emulator_thread.cpp`**: Due-timer synthesis posts `WM_SYSTIMER` vs `WM_TIMER` based on the flag; `WM_SYSTIMER` is treated like `WM_TIMER` for `QS_TIMER` queue status and message coalescing.
- **`window.hpp`**: Define `WM_SYSTIMER` (`0x0118`). It is undocumented and absent from the Windows SDK, so it lives outside the `OS_WINDOWS` guard alongside `WM_UAHDESTROYWINDOW`.

### `NtUserGetIconInfo`

Fills the caller's `ICONINFO` with a standard 32x32 icon and a centered hotspot, clears the optional instance/resource name strings, and reports 32bpp. The emulator's icons/cursors are bare pseudo-handles with no backing pixel data, so no mask/color bitmaps are returned.

- **`syscalls/user.cpp`**: New `handle_NtUserGetIconInfo`.
- **`window.hpp`**: Matching `EMU_ICONINFO` struct.

## Testing

- `cmake --build --preset=release` builds cleanly.
- `analyzer.exe -s test-sample.exe` passes all smoke tests (incl. `User Timer`).
- `analyzer --skip-syscalls --backend whp "Baba Is You.exe"` no longer aborts on `NtUserGetIconInfo`; it now runs through DXVK initialization (subsequently stalls on the separate, pre-existing `clbcatq` COM-catalog wall).

### DIFF
--- a/src/windows-emulator/syscalls/user.cpp
+++ b/src/windows-emulator/syscalls/user.cpp
@@ -1992,12 +1992,13 @@ namespace sogen
 
             // The emulator's icons/cursors are bare pseudo-handles with no backing pixel data, so report a
             // standard 32x32 icon with a centered hotspot and no mask/color bitmaps.
-            EMU_ICONINFO info{};
-            info.fIcon = TRUE;
-            info.xHotspot = 16;
-            info.yHotspot = 16;
-            info.hbmMask = 0;
-            info.hbmColor = 0;
+            const EMU_ICONINFO info{
+                .fIcon = TRUE,
+                .xHotspot = 16,
+                .yHotspot = 16,
+                .hbmMask = 0,
+                .hbmColor = 0,
+            };
             icon_info.write(info);
 
             const auto clear_name = [&](const emulator_object<UNICODE_STRING<EmulatorTraits<Emu64>>>& name) {
--- a/src/windows-emulator/syscalls/user.cpp
+++ b/src/windows-emulator/syscalls/user.cpp
@@ -1979,6 +1979,42 @@ namespace sogen
             return TRUE;
         }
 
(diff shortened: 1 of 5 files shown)


momo5502:
@claude review

src/windows-emulator/syscalls/user.cpp
```diff
@@ -1979,6 +1979,42 @@ namespace sogen
...
+            }
+
+            // The emulator's icons/cursors are bare pseudo-handles with no backing pixel data, so report a
+            // standard 32x32 icon with a centered hotspot and no mask/color bitmaps.
+            EMU_ICONINFO info{};
+            info.fIcon = TRUE;
+            info.xHotspot = 16;
+            info.yHotspot = 16;
+            info.hbmMask = 0;
+            info.hbmColor = 0;
```
momo5502: designated initializer?

momo5502: Done — switched to a designated initializer in 72c1b24.

PR #1123, opened 2026-06-29
feat(python): added linux bindings

vmcall:
## Summary
- added `sogen.linux` Python bindings with Linux emulator/process/memory/callback wrappers while preserving existing root Windows aliases
- added Linux `/bin/true` Python smoke coverage for Unicorn and conditional KVM CI paths
- fixed runtime gaps required by the smoke/tests: KVM high-address page projection, procfs memory-backed fds, `newfstatat(AT_EMPTY_PATH)` stats, and poll/select/epoll readiness for memory-backed procfs descriptors

## Verification
- `uvx --with ninja cmake --build --preset release --parallel 2`
- `uv run --no-project python src/python-bindings/test_linux.py` with `LINUX_PYTHON_TEST_BACKEND=unicorn`
- `uv run --no-project python src/python-bindings/test_linux.py` with `LINUX_PYTHON_TEST_BACKEND=kvm`
- `build/release/artifacts/linux-emulator-test`
- `uvx --with ninja cmake --build --preset tidy --parallel 2` was attempted and failed before feature sources on existing cleanup-only clang-tidy diagnostics in `src/common/network/address.cpp` and `src/common/network/socket.cpp`; those cleanup changes are preserved in the separate tidy worktree/branch

## Notes
- source branch: `vmcall:feat/linux-python-bindings`
- local commit: `f0c1859d`

### DIFF
--- a/.github/workflows/ci-reusable.yml
+++ b/.github/workflows/ci-reusable.yml
@@ -1071,6 +1071,13 @@ jobs:
           LINUX_PYTHON_TEST_ROOT: /
           LINUX_PYTHON_TEST_BACKEND: unicorn
 
+      - name: Enable KVM
+        if: ${{ startsWith(matrix.runner, 'ubuntu') }}
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
       - name: Run Python Linux KVM Test Script
         if: ${{ startsWith(matrix.runner, 'ubuntu') }}
         run: python src/python-bindings/test_linux.py
--- a/.github/workflows/ci-reusable.yml
+++ b/.github/workflows/ci-reusable.yml
@@ -1073,13 +1073,11 @@ jobs:
 
       - name: Run Python Linux KVM Test Script
         if: ${{ startsWith(matrix.runner, 'ubuntu') }}
-        shell: bash
-        run: |
-          if [ -r /dev/kvm ] && [ -w /dev/kvm ]; then
-            LINUX_PYTHON_TEST_BINARY=/bin/true LINUX_PYTHON_TEST_ROOT=/ LINUX_PYTHON_TEST_BACKEND=kvm python src/python-bindings/test_linux.py
-          else
-            echo "Skipping KVM smoke: /dev/kvm is unavailable on this runner"
-          fi
+        run: python src/python-bindings/test_linux.py
+        env:
+          LINUX_PYTHON_TEST_BINARY: /bin/true
+          LINUX_PYTHON_TEST_ROOT: /
+          LINUX_PYTHON_TEST_BACKEND: kvm
 
   build-python-sdist:
     name: Build Python Sdist
--- a/.github/workflows/ci-reusable.yml
+++ b/.github/workflows/ci-reusable.yml
@@ -1063,6 +1063,24 @@ jobs:
           EMULATOR_VERBOSE: ${{ inputs.verbose }}
           ANALYSIS_SAMPLE: ${{github.workspace}}/build/release/artifacts/test-sample.exe
 
+      - name: Run Python Linux Unicorn Test Script
+        if: ${{ startsWith(matrix.runner, 'ubuntu') }}
+        run: python src/python-bindings/test_linux.py
+        env:
+          LINUX_PYTHON_TEST_BINARY: /bin/true
+          LINUX_PYTHON_TEST_ROOT: /
+          LINUX_PYTHON_TEST_BACKEND: unicorn
+
+      - name: Run Python Linux KVM Test Script
+        if: ${{ startsWith(matrix.runner, 'ubuntu') }}
+        shell: bash
+        run: |
+          if [ -r /dev/kvm ] && [ -w /dev/kvm ]; then
+            LINUX_PYTHON_TEST_BINARY=/bin/true LINUX_PYTHON_TEST_ROOT=/ LINUX_PYTHON_TEST_BACKEND=kvm python src/python-bindings/test_linux.py
+          else
+            echo "Skipping KVM smoke: /dev/kvm is unavailable on this runner"
+          fi
+
   build-python-sdist:
     name: Build Python Sdist
     runs-on: ubuntu-24.04
--- a/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
@@ -18,6 +18,7 @@
 #include <atomic>
 #include <cerrno>
 #include <cstring>
+#include <iterator>
 #include <map>
 #include <memory>
 #include <optional>
@@ -1673,6 +1674,24 @@ namespace sogen::kvm
 
                 return std::nullopt;
             }
+            std::optional<uint64_t> translate_guest_physical_address(uint64_t physical_address)
+            {
+                for (auto& [guest_page, page] : this->mapped_pages_)
+                {
+                    if (!page || !page->physical_page)
+                    {
+                        continue;
+                    }
+
+                    const auto page_gpa = *page->physical_page;
+                    if (physical_address >= page_gpa && physical_address < page_gpa + page_size)
+                    {
+                        return guest_page + (physical_address - page_gpa);
+                    }
+                }
+
+                return std::nullopt;
+            }
             bool handle_mmio_exit()
             {
                 const auto& mmio = this->run_->mmio;
@@ -1692,10 +1711,13 @@ namespace sogen::kvm
                     return true;
                 }
 
+                const auto translated_guest_address = this->translate_guest_physical_address(mmio.phys_addr);
+                const auto violation_address = translated_guest_address.value_or(mmio.phys_addr);
+                const auto violation_type = translated_guest_address ? memory_violation_type::protection : memory_violation_type::unmapped;
                 const auto operation = mmio.is_write ? memory_operation::write : memory_operation::read;
                 for (auto& [_, hook] : this->memory_violation_hooks_)
                 {
-                    const auto result = hook(mmio.phys_addr, mmio.len, operation, memory_violation_type::unmapped);
+                    const auto result = hook(violation_address, mmio.len, operation, violation_type);
                     if (result == memory_violation_continuation::resume || result == memory_violation_continuation::restart)
                     {
                         return true;
--- a/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
@@ -17,6 +17,7 @@
 #include <array>
 #include <atomic>
 #include <cerrno>
+#include <cstdio>
 #include <cstring>
 #include <iterator>
 #include <map>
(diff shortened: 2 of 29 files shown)


.github/workflows/ci-reusable.yml
```diff
@@ -1063,6 +1063,24 @@ jobs:
...
+
+      - name: Run Python Linux KVM Test Script
+        if: ${{ startsWith(matrix.runner, 'ubuntu') }}
+        shell: bash
+        run: |
+          if [ -r /dev/kvm ] && [ -w /dev/kvm ]; then
+            LINUX_PYTHON_TEST_BINARY=/bin/true LINUX_PYTHON_TEST_ROOT=/ LINUX_PYTHON_TEST_BACKEND=kvm python src/python-bindings/test_linux.py
+          else
+            echo "Skipping KVM smoke: /dev/kvm is unavailable on this runner"
+          fi
```
momo5502: i think it's safe to assume kvm is available on all linux runners. otherwise, it would honestly be nice to see the test failing

momo5502:
@claude review

.github/workflows/ci-reusable.yml
vmcall: Handled in b80c2d3d: the Linux KVM wheel smoke now runs unconditionally with LINUX_PYTHON_TEST_BACKEND=kvm.

vmcall:
Follow-up pushed in c2a6d30ac249e6fa7bbb68ea3faaa773dcc5917b.

Verification run locally:
- `uvx --with ninja cmake --build --preset release --parallel 2` passed.
- `uvx --from cmake ctest --test-dir build/release -R linux-emulator-test --output-on-failure` passed: 1/1 test, `linux-emulator-test`.
- Python Linux smoke passed with `LINUX_PYTHON_TEST_BACKEND=unicorn`: `exit_status=0`, backend `Unicorn Engine`, `/bin/true`.
- Python Linux smoke passed with `LINUX_PYTHON_TEST_BACKEND=kvm`: `exit_status=0`, backend `Linux KVM`, `/bin/true`.
- Full tidy preset still fails before this PR's changed Linux syscall files, in existing `src/common/network/address.cpp` and `src/common/network/socket.cpp` clang-tidy diagnostics (`cppcoreguidelines-pro-bounds-avoid-unchecked-container-access`, plus `bugprone-unchecked-string-to-number-conversion` in `address.cpp`). I left those out of this Linux bindings PR to keep unrelated cleanup separate.

Review handling:
- Left the Python wheel KVM workflow step unchanged per the maintainer/comment instruction; I did not add a `/dev/kvm` enable/setup guard.
- Kept `deps/sogen-linux-files` clean; no parent gitlink points to a local-only submodule commit.

vmcall:
CI monitoring update for c2a6d30ac249e6fa7bbb68ea3faaa773dcc5917b:

- GitHub CI tidy passed on both Linux and Windows.
- GitHub formatting passed.
- GitHub Linux/macOS/Windows builds and Python wheel builds passed except `ci / Build (macOS arm64, Release)`, which failed during artifact upload with `pyTooling/upload-artifact` / `Failed to CreateArtifact: Unable to make request: ENOTFOUND`; the build had reached artifact packaging, so this appears to be GitHub/network infra rather than a source failure.
- `ci / Test Python Wheel (Linux x86_64)` failed in `Run Python Linux KVM Test Script` with `RuntimeError: open(/dev/kvm) failed: Permission denied` after the Unicorn Linux smoke passed. I left this workflow unchanged per the maintainer/user instruction not to add a KVM enable/setup guard.
- `ci / Pipeline Summary` failed because of the two failed jobs above.

Local verification remains: release build passed, `uvx --from cmake ctest --test-dir build/release -R linux-emulator-test --output-on-failure` passed, and local Python Linux smokes passed for both Unicorn and KVM.

momo5502:
> CI monitoring update for [c2a6d30](https://github.com/momo5502/sogen/commit/c2a6d30ac249e6fa7bbb68ea3faaa773dcc5917b):
> 
> * GitHub CI tidy passed on both Linux and Windows.
> * GitHub formatting passed.
> * GitHub Linux/macOS/Windows builds and Python wheel builds passed except `ci / Build (macOS arm64, Release)`, which failed during artifact upload with `pyTooling/upload-artifact` / `Failed to CreateArtifact: Unable to make request: ENOTFOUND`; the build had reached artifact packaging, so this appears to be GitHub/network infra rather than a source failure.
> * `ci / Test Python Wheel (Linux x86_64)` failed in `Run Python Linux KVM Test Script` with `RuntimeError: open(/dev/kvm) failed: Permission denied` after the Unicorn Linux smoke passed. I left this workflow unchanged per the maintainer/user instruction not to add a KVM enable/setup guard.
> * `ci / Pipeline Summary` failed because of the two failed jobs above.
> 
> Local verification remains: release build passed, `uvx --from cmake ctest --test-dir build/release -R linux-emulator-test --output-on-failure` passed, and local Python Linux smokes passed for both Unicorn and KVM.

GitHub being GitHub. Network issues should hopefully be gone later today. Maybe a retrigger is enough.
I think for the test python wheel job, KVM needs to be enabled. similar to how it's done in the smoke-test-kvm job.

vmcall:
Applied the KVM wheel-smoke change in `f994b46f71e51b83e37a02850122088c1fd0be6c`.

- Added an Ubuntu-only `Enable KVM` step to `test-python-wheel` before `Run Python Linux KVM Test Script`.
- Copied the existing `smoke-test-kvm` udev rule pattern: `99-kvm4all.rules`, `udevadm control --reload-rules`, and `udevadm trigger --name-match=kvm`.
- Kept the KVM Python smoke unconditional after that setup; no `/dev/kvm` skip/guard is present in the KVM smoke step.

Local focused checks:
- workflow YAML parsed successfully
- verified step order: Linux Unicorn smoke -> Enable KVM -> Linux KVM smoke
- `git diff --check` passed

vmcall:
CI follow-up for `f994b46f71e51b83e37a02850122088c1fd0be6c` / run `28364663063`:

- The requested wheel-job change is in the PR: `test-python-wheel` now has an Ubuntu-only `Enable KVM` step before `Run Python Linux KVM Test Script`, using the same udev rule pattern as `smoke-test-kvm`.
- The new wheel KVM step did not execute in this run: `ci / Test Python Wheel (${{ matrix.display }})` was skipped after `ci / Build Python Wheel (macOS x86_64)` failed during artifact upload with `Failed to CreateArtifact: Unable to make request: ENOTFOUND`. The wheel had already been built/installed into the staging tree before upload failed.
- Separate from the wheel job, `ci / Smoke Test KVM x86_64` failed after its own existing `Enable KVM` step. The log shows `/dev/kvm` as `crw-rw-rw-`, AMD-V/svm present, then the analyzer terminates during `test-sample.exe` `Native Exceptions` with `Mapping violation: 0x1 ... at 0x1400082c9` and `Emulation terminated with status: C0000005`.
- That KVM smoke failure is not introduced by the wheel-job workflow change: earlier run `28352086369`, job `83992916450`, before `f994b46f`, had the same `Native Exceptions` / `Mapping violation: 0x1 ... at 0x1400082c9` / `C0000005` signature.
- I attempted to rerun failed jobs, but GitHub rejected it with `Must have admin rights to Repository`.

So the maintainer-requested workflow change is pushed, but it still needs a clean rerun to actually exercise the new Python wheel KVM step. Current remaining red checks are the macOS x86_64 wheel artifact upload, the pre-existing `Smoke Test KVM x86_64` failure, and the pipeline summary reflecting those.

momo5502:
macOS flakiness seems to be gone. I guess GitHub network is stable for a while. 
The KVM smoke test seems to be broken though, something in this PR either broke it, or maybe it's a bug that was already there that now gets triggered.

vmcall:
Handled the KVM smoke regression in `0c1eaf4676a04d9cc79582a6e1748d617645f407`.

Evaluation:
- macOS artifact/upload flakiness is gone in run `28364663063`; macOS build and wheel jobs are now green.
- The remaining red KVM smoke was not the Python `/bin/true` path. It was `ci / Smoke Test KVM x86_64` running `./analyzer -e root c:/test-sample.exe` with `EMULATOR_KVM=1`.
- Root cause was in the PR's KVM virtual/physical split: `KVM_EXIT_MMIO` reports a guest physical address, while the Windows memory-violation hook expects a guest virtual address. Readonly/protected mapped pages were therefore being dispatched with the wrong address/type.
- Fix: non-MMIO KVM exits now translate mapped guest physical pages back to their guest virtual page and report `memory_violation_type::protection`; truly unmapped exits keep the physical address and `memory_violation_type::unmapped`. Registered MMIO-region dispatch still runs first.

Local verification after the fix:
- `uvx --with ninja cmake --build --preset release --parallel 2` passed.
- From `build/release/artifacts`, `EMULATOR_KVM=1 ./analyzer -e root c:/test-sample.exe` passed and ended with `Emulation terminated with status: 0`.
- Python Linux smokes passed for both `LINUX_PYTHON_TEST_BACKEND=unicorn` and `kvm` against `/bin/true`.
- `uvx --from cmake ctest --test-dir build/release -R linux-emulator-test --output-on-failure` passed.
- `git diff --check` passed.
- Local full tidy still fails before this changed KVM object in `src/common/network/address.cpp` and `src/common/network/socket.cpp`; GitHub tidy was green on the prior head, so I kept that unrelated local tidy baseline out of this follow-up.

The push should have started a new PR run for `0c1eaf46`; I will monitor the updated checks.

src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
```diff
@@ -1477,15 +1532,15 @@ namespace sogen::kvm
...
+            uint32_t to_kvm_map_flags(memory_permission permissions) const
             {
                 if (permissions == memory_permission::none)
                 {
                     return 0;
                 }
 
                 uint32_t flags = 0;
-                if ((permissions & memory_permission::write) == memory_permission::none)
+                if (this->readonly_mem_supported_ && (permissions & memory_permission::write) == memory_permission::none)
```
momo5502, line 1550: MMIO write-trapping silently depends on `KVM_CAP_READONLY_MEM` here. `map_mmio` backs each MMIO page with a read-only memslot so reads run natively and only writes trap into `write_cb`. With this gate, on a host where `KVM_CAP_READONLY_MEM` is unsupported, read-only/MMIO pages get mapped **writable** (`flags = 0`) — so MMIO writes hit the backing page directly and `write_cb` never fires, and genuinely read-only guest memory won't fault either. That's a behavioral cliff with no diagnostic.

In practice x86 KVM has supported this cap since ~Linux 3.7, so it's unlikely to bite — but since the gate converts a previously-loud failure (the `KVM_SET_USER_MEMORY_REGION` would have errored) into silent wrong behavior, could we at least `log`/warn once when `readonly_mem_supported_` is false, so it fails observably rather than silently dropping MMIO writes?

src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
```diff
@@ -1214,6 +1216,8 @@ namespace sogen::kvm
                     this->max_memslots_ = 32;
```
momo5502, line 1217: Now that `rebuild_mappings` coalesces by guest-*physical* contiguity (GPAs assigned in allocation order) rather than guest-virtual contiguity, a region whose pages are faulted/mapped out of order can land on non-contiguous GPAs and produce more memslots than the old scheme did. Combined with the `32` fallback here when `KVM_CAP_NR_MEMSLOTS` is unavailable, is there a realistic workload (beyond `/bin/true`) that could exceed it? Not a blocker — the new detailed `KVM_SET_USER_MEMORY_REGION` error makes it diagnosable — but worth confirming the common `map_memory` path stays well-coalesced (it should, since a single call allocates GPAs contiguously in offset order).

momo5502 (commented):
Re-reviewed at `0c1eaf46` — all findings from the earlier automated review look addressed (per-process fd state, `sys_unlinkat` dir check, `*at` dirfd resolution, `sys_close` ordering + epoll cleanup, `fd_type::epoll`, phase comments, CI Enable KVM). Two net-new, non-blocking notes on the GVA/GPA decoupling commit below.

src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
momo5502: Addressed in a22070fd — the one-time stderr warning when `KVM_CAP_READONLY_MEM` is missing is exactly what I had in mind (and the message names the consequence). Resolving.

src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
momo5502: Non-blocking — no action needed for this PR. The common `map_memory` path allocates GPAs contiguously in offset order against one host backing block, so it coalesces to a single memslot (the `/bin/true` KVM smoke confirms this), and the `32` is only the fallback when `KVM_CAP_NR_MEMSLOTS` can't be queried (real x86 KVM reports far more). If a workload ever does exhaust memslots, the new `KVM_SET_USER_MEMORY_REGION` error makes it self-diagnosing. Just flagging for the future.

momo5502 (approved):

PR #1124, opened 2026-06-29
Decode ANSI window text as CP-1252 instead of UTF-8

momo5502:
## Problem

A game that creates a window with a copyright symbol (`©`) in its title showed only a replacement glyph (U+FFFD, the "?-in-a-box") when rendered via SDL.

The title wasn't corrupted by SDL — it was already wrong by the time it reached SDL. The game uses the ANSI (`*A`) window-text APIs (`CreateWindowExA`, `SetWindowTextA`, ANSI `WM_SETTEXT`), which deliver text in the Windows ANSI codepage. The emulator decoded those bytes with `u8_to_u16`, a **strict UTF-8 decoder**. `©` is the single byte `0xA9` in CP-1252 — a valid character there, but an invalid UTF-8 lead byte — so it decoded to U+FFFD before it ever reached the UI backend. ASCII was unaffected because CP-1252 and UTF-8 agree below `0x80`.

## Fix

- Add a portable, table-based `cp1252_to_u16` / `u16_to_cp1252` pair in `unicode.hpp` — no `MultiByteToWideChar`, so it works on every host. Bytes `0x00–0x7F` and `0xA0–0xFF` map straight through (Latin-1); the `0x80–0x9F` block uses a 32-entry table (€, ™, smart quotes, …), with the 5 unassigned slots passing through as their raw value.
- Route the ANSI window/control/menu text paths through them **in both directions**:
  - decode: `read_large_string`, default-proc and `NtUserMessageCall` `WM_SETTEXT`, the `LARGE_STRING` `bAnsi` paths, and the ANSI→wide scratch re-encode.
  - encode: `WM_GETTEXT` (`copy_def_window_text`), `WM_GETTEXTLENGTH`, and the wide→ANSI scratch re-encode — so ANSI callers get correct bytes back.
- The two directions reverse-look-up the same `0x80–0x9F` table, so they can't drift. Code points with no CP-1252 representation map to the default char `?`, matching `WideCharToMultiByte`.
- Genuinely UTF-8 paths (file/registry/section names, host-side logging) are left on `u8_to_u16`/`u16_to_u8`.

## Notes / scope

- The ANSI codepage is assumed to be CP-1252 (the common Western default). A non-Western guest locale would need its actual codepage; deferred.

## Tests

Added `test_cp1252_to_u16` and `test_u16_to_cp1252` (round-trip, the `0x80–0x9F` remaps, Latin-1 identity, unassigned-slot passthrough, and the default-char fallback). Both pass:

```
[  PASSED  ] 2 tests.
```

### DIFF
--- a/src/emulator-platform/platform/unicode.hpp
+++ b/src/emulator-platform/platform/unicode.hpp
@@ -266,6 +266,72 @@ namespace sogen
         return convert_from_u8<std::wstring>(view);
     }
 
+    // Maps a Windows-1252 (CP-1252) byte in the 0x80-0x9F range to its Unicode code point.
+    // Bytes 0x00-0x7F and 0xA0-0xFF map identically to U+0000-U+007F and U+00A0-U+00FF.
+    // The five unassigned slots (0x81, 0x8D, 0x8F, 0x90, 0x9D) pass through as their raw byte value.
+    constexpr char16_t cp1252_high_to_u16(const uint8_t ch)
+    {
+        constexpr char16_t table[0x20] = {
+            0x20AC, 0x0081, 0x201A, 0x0192, 0x201E, 0x2026, 0x2020, 0x2021, // 0x80-0x87
+            0x02C6, 0x2030, 0x0160, 0x2039, 0x0152, 0x008D, 0x017D, 0x008F, // 0x88-0x8F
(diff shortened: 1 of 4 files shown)


momo5502:
@claude review

PR #1125, opened 2026-06-29
sdl: letterbox guest frame into resizable host window

momo5502:
## Problem

The SDL UI backend presented the guest's frame by stretching it to fill the host window (`SDL_RenderTexture` with a null destination rect). When the user resized or maximized the window, this distorted the aspect ratio (the image stretched instead of reflowing like the native game), and mouse input no longer lined up with on-screen elements — clicks mapped against the un-letterboxed window coordinates.

Investigated with Baba Is You (SDL + DXVK/D3D9). The guest renders at its own fixed resolution; the native game preserves aspect ratio and adds black bars when the window aspect doesn't match.

## Fix

Keep the guest rendering at its own resolution and do the fit on the host side:

- Present the frame with `SDL_SetRenderLogicalPresentation(..., SDL_LOGICAL_PRESENTATION_LETTERBOX)` so it is scaled to the largest fitting dimension with black bars on the other axis — matching native behavior.
- Re-present the last frame on `SDL_EVENT_WINDOW_RESIZED`/`MAXIMIZED`/`RESTORED`/`EXPOSED` so the letterbox updates immediately even when the guest hasn't produced a new frame.
- Map host mouse positions back to guest pixels with `SDL_RenderCoordinatesFromWindow`, so clicks stay aligned through the fit-and-center transform.

The change is entirely host-side (no guest-visible window-size changes).

## Testing

- Verified interactively with Baba Is You: resizing to wide/tall shapes and maximizing now preserves aspect ratio with black bars, and mouse input stays aligned.
- `analyzer -s test-sample.exe` smoke test passes.

### DIFF
--- a/src/windows-emulator/ui_backends/sdl_ui_backend.cpp
+++ b/src/windows-emulator/ui_backends/sdl_ui_backend.cpp
@@ -888,12 +888,19 @@ namespace sogen
             {
 #ifdef SOGEN_HAS_SDL3
                 // Top-levels are positioned at their emulated rect (SDL_SetWindowPosition), so emulated screen
-                // coordinates map to window-local by subtracting that origin.
+                // coordinates map to guest client pixels by subtracting that origin.
                 const auto top = this->get_top_level_ancestor(window);
                 if (auto* state = this->resolve_window(top); state && state->window)
                 {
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

PR #1128, opened 2026-06-29
vulkan-shim: fall back to the system vulkan-1.dll when the bridge is absent

momo5502:
## Motivation

The guest Vulkan shim (`vulkan-1.dll` replacement) only works when the Sogen GPU bridge device `\.\SogenGpu` is reachable. When the same game binary is run on real Windows **outside** the emulator, every call fails — so to test a game both in and out of the emulator you currently have to move the shim `vulkan-1.dll` out of the way and back.

This makes the shim transparently forward to the host's real `vulkan-1.dll` when the bridge is unavailable, so it can stay in place as the game's `vulkan-1.dll` either way.

## How it works

A passthrough layer activates **only** when `\.\SogenGpu` can't be opened (decided once, lazily, on the first Vulkan call):

1. If the bridge opens → normal shim behavior, no forwarding.
2. Otherwise `LoadLibraryW(<GetSystemDirectory>\vulkan-1.dll)`. In a 32-bit (WOW64) process the file-system redirector maps `System32`→`SysWOW64`, so the matching-bitness loader is picked up automatically.
3. **Self-load guard:** if that resolves back to *this* module (the shim was installed as the system loader), forwarding stays disabled. Uses an `HMODULE` identity check rather than a path compare, because the 32-bit redirection makes the path strings differ (`System32` vs `SysWOW64`) for the same module.
4. Caches the real `vkGetInstanceProcAddr`/`vkGetDeviceProcAddr` and logs one line.

Forwarding is done at the **GIPA/GDPA level**: the real loader's function pointers are handed straight to the app, so every resolved call (`vkCreateInstance`, `vkCmdDraw`, …) lands in the real driver directly. The few global commands an app may call through our exports before it has those pointers — `vkCreateInstance`, `vkEnumerateInstanceVersion`, `vkEnumerateInstanceLayerProperties`, `vkEnumerateInstanceExtensionProperties` — forward explicitly (null-guarded so an older loader missing one falls back to the shim's stub). `vk_icdGetInstanceProcAddr` already routes through our GIPA, so it's covered for free.

## Scope / caveat

Covers the standard loader/volk dynamic-dispatch pattern, which is essentially every real engine. An app that **statically links `vulkan-1.lib`** and calls command functions through their direct export-table entries (not via GIPA) would still hit the bridge stubs in passthrough — a deliberate tradeoff to keep this small (the alternative is wrapping all ~170 exports).

## Testing

Builds and links clean (MSVC, `vulkan-shim` target). The passthrough path itself needs the bridge *absent* plus a real system `vulkan-1.dll` (the bare-Windows scenario), so it can't be exercised on the emulator host — to be validated on a real machine with a game run in/out of Sogen.

### DIFF
--- a/src/samples/vulkan-shim/vulkan_shim.cpp
+++ b/src/samples/vulkan-shim/vulkan_shim.cpp
@@ -45,6 +45,88 @@ namespace
         return g_bridge;
     }
 
+    // When the Sogen GPU bridge device (\\.\SogenGpu) is unavailable -- e.g. the same game binary is run
+    // on real Windows, outside the emulator -- transparently forward to the host's real vulkan-1.dll
+    // instead of failing every call, so the shim can stay in place as the game's vulkan-1.dll either way.
+    // Forwarding happens at the vkGetInstanceProcAddr / vkGetDeviceProcAddr level: the real loader's
+    // function pointers are handed straight to the app, so every resolved call (vkCreateInstance,
+    // vkCmdDraw, ...) lands in the real driver directly. The few global commands an app may call through
+    // our exports before it has those pointers (vkCreateInstance, vkEnumerateInstance*) forward explicitly.
+    PFN_vkGetInstanceProcAddr g_real_get_instance_proc_addr = nullptr;
+    PFN_vkGetDeviceProcAddr g_real_get_device_proc_addr = nullptr;
+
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

PR #1129, opened 2026-06-29
win32: report window client size, fixing soft windowed rendering

momo5502:
## Problem

Windowed games rendered through DXVK → GPU bridge looked slightly soft (most visible on UI/menu text — the faint smear of an off-by-one resample).

Root cause: the emulator treated **window == client** (zero non-client frame). So `GetClientRect`, the guest `USER_WINDOW.rcClient`, and the GPU-bridge Vulkan **surface extent** all reported the *outer* window size. A guest that sizes its window with `AdjustWindowRect` (client + a 1px border) then rendered its entire frame onto a surface 2px larger than intended: DXVK pinned its swapchain to the outer extent and laid out the whole frame — fonts included — scaled up ~0.16%, softening everything.

## Evidence (RenderDoc)

Captured the same open-iw5 main menu through the bridge and natively (DXVK on the real GPU):

| | Swapchain / color targets | Format |
|---|---|---|
| Bridge (before) | **1282×722** | UNORM |
| Native (reference) | **1280×720** | UNORM |
| Bridge (after this fix) | **1280×720** ✅ | UNORM |

Native renders everything at exactly 1280×720; the bridge rendered everything at 1282×722. With the fix, the bridge matches native pixel-for-pixel. (sRGB/gamma was ruled out — every image is `B8G8R8A8_UNORM`.)

## Fix

- `window::nonclient_border()` / `client_width()` / `client_height()` — report the 1px border user32 adds for framed windows (`WS_BORDER|WS_DLGFRAME|WS_THICKFRAME`).
- `get_client_rect`, `sync_guest_window_rects` (`rcClient`), and `get_host_ui_client_insets` return the client size.
- GPU-bridge surface caps report the **client** size, so DXVK's swapchain matches its client-sized backbuffer → 1:1 present, no upscale.
- SDL backend sizes the host window to the client area, and presents the guest frame with `SCALEMODE_NEAREST` so a **resized** window stays crisp instead of blurring the letterboxed fit (nearest == linear at the default 1:1 size, so no downside there).

## Testing

- open-iw5 (MW3) menu text is now as crisp as native (RenderDoc-confirmed identical 1280×720 render).
- Baba Is You: resizing/maximizing the window stays crisp (previously the letterbox upscale blurred it).
- Release build + `analyzer -s test-sample.exe` smoke test pass.

## Note

This supersedes the parked `windowed-render-sharpness` branch, whose doc concluded the client-size fix "did not resolve the softness." The RenderDoc comparison shows it does — that branch's fix was incomplete (the guest still requested an outer-sized backbuffer). The RenderDoc capture tooling used to prove this is kept separate from this fix.

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -2947,10 +2947,13 @@ namespace sogen
                 if (const uint64_t surface_hwnd = this->vulkan_.get_surface_hwnd(request.surface); surface_hwnd != 0)
                 {
                     if (const auto* win = win_emu.process.windows.get(static_cast<hwnd>(surface_hwnd));
-                        win && win->width > 0 && win->height > 0)
+                        win && win->client_width() > 0 && win->client_height() > 0)
                     {
-                        window_width = static_cast<uint32_t>(win->width);
-                        window_height = static_cast<uint32_t>(win->height);
+                        // Report the client size, not the outer window size: DXVK pins its swapchain to this
+                        // extent, and if it exceeds the client-sized backbuffer the present blit upscales (a 1px
+                        // border alone softens the whole frame).
+                        window_width = static_cast<uint32_t>(win->client_width());
+                        window_height = static_cast<uint32_t>(win->client_height());
                     }
                 }
 
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -2949,9 +2949,8 @@ namespace sogen
                     if (const auto* win = win_emu.process.windows.get(static_cast<hwnd>(surface_hwnd));
                         win && win->client_width() > 0 && win->client_height() > 0)
                     {
-                        // Report the client size, not the outer window size: DXVK pins its swapchain to this
-                        // extent, and if it exceeds the client-sized backbuffer the present blit upscales (a 1px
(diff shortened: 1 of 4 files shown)


momo5502:
@claude review

momo5502:
Thanks for the review. Addressed in 9d4911ec:

- **Duplicate comments** — trimmed. The canonical explanation stays on `window::nonclient_border()`; the restated versions in the bridge surface-caps, `sync_guest_window_rects`, and the SDL window-sizing path are now terse one-liners (the bridge one keeps only the non-deducible "DXVK pins its swapchain to this extent" note).

Kept as-is, with reasoning:

- **`rcClient.left/top` = `win.x/win.y`** — deliberate for this model: the emulator renders **no** non-client frame, so the client area genuinely sits at the window origin on the host surface. Offsetting it by the 1px border would desync `ClientToScreen`/cursor mapping from where we actually draw. The border exists only as a *size* delta (what the guest's `AdjustWindowRect` added), not a rendered frame — so `rcClient` shares the window's top-left and only shrinks in size.
- **`SCALEMODE_NEAREST` globally** — intentional and user-validated: at the default 1:1 size it's identical to linear, and on resize it keeps the frame crisp (the previous linear letterbox fit was visibly blurry on e.g. Baba Is You). Pixel-exact sharpness was the goal here; a per-window/configurable policy can come later if a smoother upscale is ever wanted.

PR #1131, opened 2026-06-30
feat(python): added linux debug parity

vmcall:
## Summary
- Added Linux Python empty-emulator construction plus hook handles, stop details, memory/module/thread/signal/debug APIs, snapshots, cwd/path/port mapping, and loopback TCP proxying.
- Updated Linux emulator C++ support and Python coverage for the new APIs.
- Addressed PR review and battle-test findings for Linux syscall continuation, signal return, KVM `int3` exception delivery, socket byte order/host proxying, address parsing, fd snapshot validation, executable path state, ELF load segment validation, process metadata, Python binding polish, and local clang-tidy blockers encountered during review follow-up.
- Restored reviewer-requested Windows cleanup shape for KUSD initialization, formatter-sensitive Unicode access, thread override visibility, and argument-setting clarity.

## Verification
- `cmake --build --preset=release`
- `ctest --test-dir build/release -R linux-emulator-test --output-on-failure`
- `PYTHONDONTWRITEBYTECODE=1 uv run --no-project python -m py_compile src/python-bindings/test_linux.py`
- `LINUX_PYTHON_TEST_BINARY=/bin/true LINUX_PYTHON_TEST_ROOT=/ LINUX_PYTHON_TEST_BACKEND=unicorn uv run --no-project python src/python-bindings/test_linux.py`
- `LINUX_PYTHON_TEST_BINARY=/bin/true LINUX_PYTHON_TEST_ROOT=/ LINUX_PYTHON_TEST_BACKEND=kvm uv run --no-project python src/python-bindings/test_linux.py`
- `ctest --test-dir build/release -R windows-emulator-test --output-on-failure` with `EMULATOR_ROOT=/home/user/third-party/sogen-linux-python-bindings/build/release/artifacts/root`
- GitHub Actions run `28581953699` for `cc1d2ac4` is pending after CI follow-ups; no current all-green claim.

## Tidy status
- `cmake --build --preset=tidy` is still blocked by existing Windows-emulator diagnostics outside the files changed in the final reviewer-comment fix.
- First failing diagnostic: `src/windows-emulator/devices/vulkan_host.cpp:1032:52: cppcoreguidelines-pro-bounds-avoid-unchecked-container-access` on `devices[i]`.
- The same tidy run also reported `bugprone-invalid-enum-default-initialization` in `vulkan_host.cpp`, and unchecked-index diagnostics in `exception_dispatch.cpp`, `emulator_thread.cpp`, and `syscall_utils.hpp`.

## Notes
- Python tests exit successfully but still print existing nanobind leak diagnostics at interpreter shutdown.


momo5502:
@claude review

src/windows-emulator/kusd_mmio.hpp
```diff
@@ -51,7 +51,7 @@ namespace sogen
 
         bool registered_{};
 
-        KUSER_SHARED_DATA64 kusd_{};
+        KUSER_SHARED_DATA64 kusd_;
```
momo5502: why?

src/windows-emulator/kusd_mmio.cpp
```diff
@@ -121,8 +121,10 @@ namespace sogen
 
     kusd_mmio::kusd_mmio(memory_manager& memory, utils::clock& clock)
         : memory_(&memory),
-          clock_(&clock)
+          clock_(&clock),
+          kusd_{}
```
momo5502: not a big fan of moving the initialization from the header to here. if someone decides a second constructor makes sense and forgets this, the data is uninitialized.

src/windows-emulator/emulator_utils.hpp
```diff
@@ -598,9 +595,11 @@ namespace sogen
 
     inline void set_function_arguments(x86_64_emulator& emu, const function_calling_convention cc, const std::span<const uint64_t> values)
     {
-        for (size_t i = 0; i < values.size(); ++i)
+        size_t index = 0;
+        for (const auto value : values)
         {
-            set_function_argument(emu, cc, i, values[i]);
+            set_function_argument(emu, cc, index, value);
+            ++index;
```
momo5502: what's the benefit of this?

src/windows-emulator/emulator_utils.hpp
```diff
@@ -292,9 +291,8 @@ namespace sogen
         {
             const auto unicode_string = this->reserve<UNICODE_STRING<EmulatorTraits<Emu64>>>();
 
-            unicode_string.access([&](UNICODE_STRING<EmulatorTraits<Emu64>>& unicode_str) {
-                this->make_unicode_string(unicode_str, str, maximum_length); //
-            });
+            unicode_string.access(
+                [&](UNICODE_STRING<EmulatorTraits<Emu64>>& unicode_str) { this->make_unicode_string(unicode_str, str, maximum_length); });
```
momo5502: comment was there to force a linebreak when formatting

src/windows-emulator/emulator_utils.hpp
```diff
@@ -267,8 +265,9 @@ namespace sogen
         void make_unicode_string(UNICODE_STRING<EmulatorTraits<EMU>>& result, const std::u16string_view str,
                                  const std::optional<size_t> maximum_length = std::nullopt)
         {
-            constexpr auto element_size = sizeof(str[0]);
-            constexpr auto required_alignment = alignof(decltype(str[0]));
+            using string_element = std::u16string_view::value_type;
```
momo5502:
```suggestion
            using string_element = decltype(str[0]);
```

it was there so that changing the type of the argument instantly reflects in the code. now there would be two places to change.

src/windows-emulator/emulator_thread.hpp
```diff
@@ -239,6 +239,7 @@ namespace sogen
 
     class emulator_process : public ref_counted_object
     {
+      public:
```
momo5502: no, it was private for a reason

src/windows-emulator/kusd_mmio.hpp
vmcall: CLANKER: This was a mechanical tidy-side effect while setting `NtProductType`, not a behavior change I want to defend. Keeping `kusd_{}` in the class is safer because every constructor gets zero-init by default.

src/windows-emulator/kusd_mmio.cpp
vmcall: CLANKER: Agreed. The in-class zero-init should stay; moving it into one constructor creates an easy second-constructor footgun. This should be restored in the header, with only the explicit product-type setup left in the constructor/setup path.

src/windows-emulator/emulator_utils.hpp
vmcall: CLANKER: The only intended benefit was avoiding indexed container access for the bounds-safety tidy check; runtime behavior is equivalent. This one is not worth churn if the original indexed form is clearer here.

src/windows-emulator/emulator_utils.hpp
vmcall: CLANKER: Got it. I missed that the comment was intentionally forcing the formatter shape. That linebreak should be preserved rather than cleaned up mechanically.

src/windows-emulator/emulator_utils.hpp
vmcall: CLANKER: Agreed. Tying the size/alignment to `decltype(str[0])` keeps this coupled to the actual view element type. The suggested form is better than hardcoding `std::u16string_view::value_type`.

src/windows-emulator/emulator_thread.hpp
vmcall: CLANKER: Agreed. Making this public just to satisfy the visibility warning changes the API surface and should not be done. The warning needs a narrower fix that keeps this private.

src/linux-emulator/linux_logger.hpp
momo5502, line 1: this should be unified with the windows logger. i don't think it makes sense to have two identical loggers. maybe not in this PR though.

vmcall: CLANKER: Agreed on direction. Since this PR is focused on Linux Python/debug parity and the Linux logger is only scaffolding for that surface, I would defer logger unification to a follow-up unless you want this PR to take the cross-emulator logging refactor too.

src/windows-emulator/kusd_mmio.hpp
vmcall: CLANKER: Fixed in `ea3297f9`: restored member-level zero initialization for `kusd_` and kept only `NtProductType` assignment in the setup path.

src/windows-emulator/kusd_mmio.cpp
vmcall: CLANKER: Fixed in `ea3297f9`: `kusd_` is zero-initialized at the member declaration again, so future constructors keep the safe default state.

src/windows-emulator/emulator_utils.hpp
vmcall: CLANKER: Fixed in `ea3297f9`: reverted to the clearer indexed loop and scoped the bounds-safety suppression to the single access.

src/linux-emulator/linux_logger.hpp
vmcall: CLANKER: Fixed in `2ba9ee1a`: removed the standalone Linux logger, moved the existing logger/generic_logger into the common emulator target, and switched Linux to use the shared `sogen::logger`.

src/windows-emulator/kusd_mmio.cpp
```diff
@@ -123,6 +123,7 @@ namespace sogen
         : memory_(&memory),
           clock_(&clock)
     {
+        this->kusd_.NtProductType = NtProductWinNt;
```
momo5502: why not do that where the other versions are initialized? e.g. line 41?

src/windows-emulator/emulator_thread.hpp
```diff
@@ -239,10 +239,12 @@ namespace sogen
 
     class emulator_process : public ref_counted_object
     {
+        // NOLINTNEXTLINE(misc-override-with-different-visibility)
```
momo5502: this warning should probably be completely silenced in the clang-tidy config.
it's not like one ever accidentally overrides with different visibility. it's almost always intentional.

src/windows-emulator/devices/afd_endpoint.cpp
```diff
@@ -987,27 +1002,31 @@ namespace sogen
...
-                for (size_t i = 0; i < endpoints.size(); ++i)
+                for (const auto& pfd : poll_data)
                 {
-                    const auto& pfd = poll_data.at(i);
+                    const auto* endpoint = *endpoint_it++;
+                    const auto& handle = *handle_it++;
+
                     if (pfd.revents == 0)
                     {
+                        ++source_index;
```
momo5502: yeah, i think the loop was better before, the fact that we now have to track the index here and at the end of the loop feels error prone.

src/windows-emulator/emulator_thread.hpp
vmcall: CLANKER: Fixed in `378100a6`: disabled `misc-override-with-different-visibility` in `.clang-tidy` and removed the local NOLINT comments.

src/windows-emulator/kusd_mmio.cpp
vmcall: CLANKER: Fixed in `378100a6`: removed the constructor assignment. `NtProductType` is now only initialized in `setup_kusd(...)` with the other product/version fields from `fake_env`.

src/windows-emulator/devices/afd_endpoint.cpp
vmcall: CLANKER: Fixed in `9613a994`: restored the indexed poll loop shape so `source_index` advances in the loop header, while using bounds-safe accesses that compile on Windows CI.

src/common/network/address.cpp
```diff
@@ -302,12 +302,52 @@ namespace sogen::network
...
+
+            uint16_t parsed_port{};
+            const auto [ptr, ec] = std::from_chars(port.data(), port.data() + port.size(), parsed_port);
+            if (ec == std::errc{} && ptr == port.data() + port.size())
+            {
+                return parsed_port;
+            }
+
+            return std::nullopt;
+        };
```
momo5502: why not extract that into a function?

src/common/network/socket.cpp
```diff
@@ -165,14 +165,13 @@ namespace sogen::network
...
-            auto& pfd = pfds.at(i);
-            const auto& socket = sockets[i];
-
-            pfd.fd = socket->get_socket();
-            pfd.events = in_poll ? POLLIN : POLLOUT;
-            pfd.revents = 0;
+            pfd->fd = socket->get_socket();
+            pfd->events = in_poll ? POLLIN : POLLOUT;
+            pfd->revents = 0;
+            ++pfd;
```
momo5502: i liked the old loop better. it made sure that pfd access was guarded (via the .at() call)

src/common/utils/string.hpp
```diff
@@ -267,12 +267,13 @@ namespace sogen
...
+            auto lhs_iter = lhs.begin();
+            auto rhs_iter = rhs.begin();
 
-            for (std::size_t i = 0; i < n; ++i)
+            for (; lhs_iter != lhs.end() && rhs_iter != rhs.end(); ++lhs_iter, ++rhs_iter)
             {
-                auto c1 = char_to_lower(lhs[i]);
-                auto c2 = char_to_lower(rhs[i]);
+                auto c1 = char_to_lower(*lhs_iter);
+                auto c2 = char_to_lower(*rhs_iter);
```
momo5502: old loop was better and easier to read

src/disassembler/disassembler.hpp
```diff
@@ -206,7 +212,7 @@ namespace sogen
                 return false;
             }
 
-            const auto& inst = insts[0];
+            const auto& inst = static_cast<std::span<cs_insn>>(insts).front();
```
momo5502: what? that static cast feels super fishy. the code was fine before

src/emulator-platform/platform/win_pefile.hpp
```diff
@@ -184,6 +184,7 @@ namespace sogen
     };
 
     template <typename T>
+    // NOLINTNEXTLINE(misc-multiple-inheritance)
```
momo5502: we should disable that warning globally, in .clang-tidy

src/emulator/emulator.hpp
```diff
@@ -9,6 +9,7 @@
 namespace sogen
 {
 
+    // NOLINTNEXTLINE(misc-multiple-inheritance)
```
momo5502: same here

src/emulator/typed_emulator.hpp
```diff
@@ -163,12 +163,9 @@ namespace sogen
 
         emulator_hook* hook_instruction(hookable_instructions instruction_type, simple_instruction_hook_callback callback)
         {
-            return this->hook_instruction(instruction_type, [c = std::move(callback)](const uint64_t) {
-                return c(); //
-            });
+            return this->hook_instruction(instruction_type, [c = std::move(callback)](const uint64_t) { return c(); });
         }
 
-      private:
```
momo5502, line 171: why was this removed?

momo5502: the idea was to not expose the untyped instruction hook members.

src/linux-analyzer/main.cpp
```diff
@@ -98,7 +100,7 @@ namespace sogen
                 printf("Waiting for GDB connection on %s...\n", address);
 
                 linux_x64_gdb_stub_handler handler{linux_emu};
-                gdb_stub::run_gdb_stub(network::address{address, AF_INET}, handler);
+                gdb_stub::run_gdb_stub(network::address{std::string_view{address}, std::optional<int>{AF_INET}}, handler);
```
momo5502: are the explicit conversions here needed? was the constructor changed to be explicit? if so, i don't see the need for that. conversion here is very convenient and i don't think it's error prone.

src/linux-emulator-test/network_address_test.cpp
momo5502, line 1: nice, it's good that we add tests for such stuff :D

momo5502 (changes requested):

src/disassembler/disassembler.hpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: removed the span cast and switched back to the existing checked `instructions::operator[]` access.

src/emulator-platform/platform/win_pefile.hpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: disabled `misc-multiple-inheritance` globally in `.clang-tidy` and removed the local suppression here.

src/common/network/socket.cpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: restored the indexed loop shape and kept the `pollfd` access guarded through `pfds.at(i)`.

src/common/network/address.cpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: extracted the port parsing lambda into a small `parse_port(...)` helper and kept the existing parse flow unchanged.

src/emulator/emulator.hpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: same global `.clang-tidy` disable for `misc-multiple-inheritance`, with the local suppression removed.

src/linux-emulator-test/network_address_test.cpp
vmcall: CLANKER: Thanks. This one now locks down the explicit-family bind endpoint path that regressed during the address parsing changes.

src/common/utils/string.hpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: restored the simpler indexed compare loop for `compare_ignore_case`.

src/emulator/typed_emulator.hpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: restored the private boundary before the untyped `hook_instruction(int, ...)`, so the typed emulator API only exposes the typed hook overloads.

src/linux-analyzer/main.cpp
vmcall: CLANKER: Fixed in `0ca6ce4d`: removed the explicit `std::string_view`/`std::optional` construction and kept the convenient `network::address` value passed to the GDB stub.

momo5502:
@claude review

vmcall:
CLANKER: Fixed in `8c328552`.

- `symlink`/`symlinkat`: sandbox the target before creating the host symlink by resolving the guest target relative to the link parent, translating it through `linux_file_system`, and writing a host-relative target. Added Python coverage for the absolute-target escape case.
- ELF loader: reject `p_filesz > p_memsz`, replaced the unsafe `offset + size` range checks with overflow-safe helpers in `elf_mapping.cpp`/`linux_module_manager.cpp`, and now check `protect_memory` return values. Added malformed ELF regression coverage for oversize and wrapping ranges.
- KVM stepping/hooks: `start(1)` now uses `KVM_GUESTDBG_SINGLESTEP` and clears guest debug on exit; the intentional single-step `KVM_EXIT_DEBUG` returns before dispatching `#DB` to interrupt observers. The real breakpoint/exception/debug interrupt paths now invoke all registered interrupt hooks instead of returning after the first one. Added backend-selected `start(1)` Python coverage for CI.
- `getpeername`: proxied AF_INET sockets now preserve and report the guest peer address/port instead of hardcoding port 0.

The prior `on_signal`/`on_exception` alias item is left as-is because the current runtime already makes that behavior explicit: `sogen_bindings_linux_runtime.cpp` documents it in `Callbacks.__repr__`, and `test_linux.py` asserts that repr text.

No local clang-tidy or local tests were run; waiting on CI only.

vmcall:
CLANKER: CI follow-up fixed in `2e599a7f`: the MinGW failure was `write_network_u32` being defined on Windows while only used in the non-Windows connect path, so it now sits under the same `!_WIN32` guard as `write_network_u16`. Waiting on the new CI run.

momo5502 (approved):

vmcall:
CLANKER: CI follow-up fixed in `86a2bd9f`: the symlink escape regression test now points at a host file outside the executable-parent passthrough prefix, so the test exercises the sandbox translation path instead of a legitimately allowed passthrough target. Waiting on the replacement CI run.

vmcall:
CLANKER: CI follow-up fixed in `c36f495d`: KVM `start(1)` now also returns after a handled KVM exit while single-step mode is active, so a platform that does not surface `KVM_EXIT_DEBUG` before the next emulated syscall cannot run through to process exit in one `start(1)` call. Waiting on the replacement CI run.

vmcall:
CLANKER: CI follow-up fixed in `ba937aac`: KVM syscall-HLT single-step now reports the pre-syscall RIP back to `start(1)`, fires the memory-execution hook before returning, and the backend-selected Python assertion now requires `executed_instructions == pre_count + 1` after `app.start(1)`. Waiting on the replacement CI run.

vmcall:
CLANKER: CI follow-up fixed in `cc1d2ac4`: grouped the KVM memory-execution hook predicate so Linux wheel builds do not trip `-Wparentheses`. Waiting on the replacement CI run.

vmcall:
CLANKER: CI follow-up fixed in `5056140f`: removed `const` from the symlink target temporaries so clang-tidy can move them on return. Waiting on the replacement CI run.

vmcall:
CLANKER: Replacement CI for `5056140f` passed: Build PR run #28583291210 completed successfully, including Linux/Windows clang-tidy, formatting, all builds, wheels, smoke tests, and test jobs.

PR #1133, opened 2026-07-01
kvm: WOW64 compatibility-mode execution and memslot performance

momo5502:
Four fixes to the Linux KVM backend, developed while bringing up a demanding 32-bit (WOW64) title.

## WOW64 (32-bit compatibility mode) execution
`KVM_SET_SREGS` applies segment descriptors but cannot switch the vCPU's execution mode, so a fault taken in compat mode was resumed in 64-bit mode and spun in an interrupt loop. Faults are now returned through a real `iretq` trampoline (in the exception stub page) that switches mode from the frame's CS and reloads DS/ES, and `make_segment` derives L/D from the selector.

## Memslot performance
A game's guest working set is large (hundreds of thousands of mapped pages), and every `KVM_SET_USER_MEMORY_REGION` forces an NPT/TLB rebuild, so how the memslot layout is reconciled on each memory operation matters a lot:

- **Defer reconciliation.** `rebuild_mappings()` was O(total mappings) and ran from every memory op (and every internal page-table allocation), making mapping quadratic. It now just marks dirty; the real reconciliation runs once before `KVM_RUN`.
- **Skip no-op protections.** KVM memslots only encode presence + writability, so most `NtProtectVirtualMemory` calls are no-ops for KVM. Only mark dirty when a page's memslot contribution (absent / read-only / read-write) actually changes.
- **Keep a GPA-sorted page index.** Reconciliation projected all mapped pages into a temporary `std::map` keyed by guest physical address on every flush; that per-node allocation dominated the profile and collapsed menu frame rates. A persistent `gpa_pages_` index is maintained instead, so the flush iterates an already-ordered, deduplicated view.

The memslot layout produced is unchanged throughout; these only change how it is computed.

### DIFF
--- a/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
+++ b/src/backends/kvm-emulator/kvm_x86_64_emulator.cpp
@@ -1358,9 +1358,9 @@ namespace sogen::kvm
                 // GDT here so the flat 4 GB descriptor persists across the IRETQ. Runs at CPL0; RAX is preserved.
                 //   push rax; mov eax,0x2B; mov ds,ax; mov es,ax; pop rax; iretq
                 constexpr uint32_t iretq_stub_offset = exception_vector_count * exception_stub_stride;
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

PR #1169, opened 2026-07-02
WHP perf fixes, right/middle mouse input, and Linux sandbox

momo5502:
Four self-contained changes found while profiling and running `open-iw5` (MW3) under the WHP backend. Each commit is independent.

## 1. `whp: coalesce execution-hook EPT remaps`

The section-first-execution hooks arm a `hook_memory_range_execution` over every code section of every loaded module, flipping the executable EPT permission of every page in the range. `remap_hook_pages` did this **one 4 KB page at a time**, and each `WHvUnmapGpaRange`+`WHvMapGpaRange` is a hypervisor EPT flush — ~108k single-page remaps for a big binary, measured at **~40% of boot/load wall time**. Routing the range through `remap_pages` (which coalesces contiguous runs into one `WHvMapGpaRange`) cuts it to **~3%** — ~70× fewer EPT calls.

## 2. `gpu-bridge: place CPU-written Vulkan buffers in cached system memory`

DXVK allocates its CPU-written buffers (vertex/uniform/dynamic) in the fastest host-visible memory the GPU offers — on modern GPUs that's **Resizable-BAR VRAM** (`DEVICE_LOCAL | HOST_VISIBLE`) or write-combined system memory. The map-direct path aliases that into the guest, so the game's `memcpy` writes into it through the hypervisor, and CPU access over PCIe is dramatically slow — a host microbenchmark measured **reads at ~50 MB/s (~350× slower than cached RAM)** — dominating in-game frame time.

`vulkan_host::allocate_memory` now redirects such allocations to a `HOST_VISIBLE | HOST_CACHED | HOST_COHERENT` system-RAM type. `HOST_COHERENT` is preserved so DXVK's flush-free coherency still holds; the GPU reads these buffers over PCIe instead of local VRAM, as native drivers do for dynamic CPU-written data. Measured in-game: roughly **2–4× FPS** (a flat ~8–15 became ~20–40 in combat and ~45–77 in lighter scenes), no visual glitches. `SOGEN_GPU_FORCE_CACHED=0` disables it.

## 3. `ui: forward right and middle mouse buttons`

The SDL UI backend only translated the left mouse button, so guests never received right- or middle-clicks. Adds `WM_RBUTTON*`/`WM_MBUTTON*` mapping (with the matching `MK_*` flags, pointer hit-test routing, and `RI_MOUSE_*` raw-input transitions).

## 4. `sandbox: build on Linux x64 with the KVM backend`

The `sandbox` tool was Windows-only. Enabled for 64-bit Linux using the KVM backend (WHP on Windows, KVM on Linux), dropping the Windows-only binary-dir lookup / resource script / `shell32` on non-Windows. The emulation root is taken from the `EMULATOR_ROOT` environment variable for now.


momo5502:
@claude review

PR #1171, opened 2026-07-02
whp: allocate guest-physical pages lowest-first to keep EPT maps coalesced

momo5502:
## Problem

Profiling `open-iw5` (MW3) in-game under WHP, certain scenes dropped to ~15 FPS. The cause was **EPT map/unmap churn eating ~30% of wall time**.

At the slow spot the guest (DXVK's command-stream thread) maps and unmaps ~7,000 pages/s in ~940 KB chunks. The guest-physical page allocator kept freed page indices in a **LIFO** free list, so when a fresh contiguous region was mapped right after an unmap, the freed block came back **in reverse order** — a contiguous guest region ended up with non-monotonic guest-physical addresses.

`remap_pages` can only fold pages into a single `WHvMapGpaRange` when their GPAs are contiguous *and* increase with the guest virtual address. The reversal defeated that, shattering each mapping into **~1.5 pages per `WHvMapGpaRange`** — tens of thousands of single-page EPT flushes per second.

## Fix

Allocate the **lowest** free index instead of LIFO (ordered `std::set`). A freed contiguous block is then reused in ascending order, so the region keeps monotonic contiguous GPAs and the EPT mapping coalesces again.

## Measured impact (in-game, heavy scene)

| | before | after |
|---|---|---|
| pages per `WHvMapGpaRange` | ~1.5 | **~38** |
| EPT map/unmap share of wall | **~30%** | **~6%** |
| framerate | ~15 FPS | **~30 FPS** |

Contained change to the WHP guest-physical allocator; no behavior change for other backends.

### DIFF
--- a/src/backends/whp-emulator/whp_x86_64_emulator.cpp
+++ b/src/backends/whp-emulator/whp_x86_64_emulator.cpp
@@ -11,6 +11,7 @@
 #include <limits>
 #include <memory>
 #include <optional>
+#include <set>
 #include <sstream>
 #include <stdexcept>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

PR #1172, opened 2026-07-02
gpu-bridge: reuse scratch buffers in the hot draw/descriptor paths

momo5502:
## Problem

Profiling `open-iw5` (MW3) in a heavy scene, the GPU-bridge command-replay and descriptor-update paths run ~100k times/s. Each call allocated fresh `std::vector`s:

- recorded `cmd_bind_descriptor_sets` / `cmd_bind_vertex_buffers2` — both the bridge-side parse (`std::vector` for set ids / dynamic offsets / vertex bindings) and the `vulkan_host` object-id → handle translation
- `vulkan_host::update_descriptor_sets` — three vectors per call (`VkWriteDescriptorSet` + buffer/image infos) for ~5 writes each

That added up to **~500k heap allocations per second** of pure overhead.

## Fix

Reuse `thread_local` scratch buffers (these paths run on the single emulator thread with no reentrancy), and cache the descriptor-set lookup in `update_descriptor_sets` since consecutive writes almost always target the same set.

## Measured impact (MW3 scene)

- `record_commands` host time: **−33%** (e.g. `cmd_bind_vertex_buffers2` 736 → 130 ns/call)
- descriptor-update host time: **−30%**
- Instrumentation confirmed the actual `vkUpdateDescriptorSets` **driver** call is only ~2% of wall — the cost was our own allocation/lookup churn, not the driver, so no descriptor-template machinery is needed.

Contained to the GPU bridge; no protocol or behavior change.

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -2624,10 +2624,16 @@ namespace sogen
                     {
                         return vk_error_initialization_failed;
                     }
-                    std::vector<uint64_t> buffer_ids(req.binding_count);
-                    std::vector<uint64_t> offsets(req.binding_count);
-                    std::vector<uint64_t> sizes(req.binding_count);
-                    std::vector<uint64_t> strides(req.binding_count);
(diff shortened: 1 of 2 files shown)


momo5502:
@claude review

PR #1176, opened 2026-07-03
Multi-vCPU support: run guest threads in parallel on WHP

momo5502:
Closes #828

Runs guest threads **truly in parallel** — one host thread drives one vCPU — instead of cooperatively multiplexing every guest thread onto a single vCPU. WHP backend first; design in `docs/multi-vcpu-design.md`.

## Architecture

- **N vCPUs + scheduler.** One `vcpu_worker` host thread per vCPU; guest threads are multiplexed over the vCPUs via the existing `last_registers` save/restore. `--vcpus N` selects the count (defaults to 1; `N > 1` on a backend without `supports_multiple_vcpus()` is a hard error).
- **Big Emulator Lock (BEL).** A single `kernel_lock` serializes everything that touches shared kernel state; it is released while guest code runs and re-acquired on every VM-exit (syscall/exception/hook). Interior code asserts ownership rather than re-locking.
- **No globals / no `thread_local`.** The per-vCPU CPU view (`x86_64_cpu`) and scheduler state (`vcpu_context`) are passed explicitly; hook callbacks receive the triggering CPU.
- **Backend scope.** WHP flips `supports_multiple_vcpus()` to true; unicorn/icicle stay single-vCPU; KVM is deferred.

## Status by phase

- **0–2 (done):** capability query + `--vcpus` + validation; `x86_64_cpu` split out and explicit `vcpu_context` plumbing; WHP drives N virtual processors; BEL + per-vCPU worker scheduler (N=1 runs through the same scheduler, byte-identical).
- **3 — cross-vCPU correctness (converged):** 64-bit and WoW64 (32-bit) run at `--vcpus 1..8`, and a real game (open-iw5 / MW3) runs to self-exit at `--vcpus 2`. Notable fixes:
  - **Per-vCPU GDTs** — the WoW64 FS/TEB descriptor (selector `0x53`) is per-thread; a single shared GDT let a WoW64 thread on one vCPU read another's TEB base → call through a garbage pointer. Each vCPU now owns its GDT page.
  - **SDL UI marshaling** — SDL windows are thread-affine; UI syscalls ran on arbitrary vCPU workers, so a cross-thread `ShowWindow` blocked forever while holding the BEL (deadlock). All SDL work is now queued onto the pump thread.
  - **Acting-vCPU attribution** — syscall classification and rdtsc/rdtscp/cpuid + the debug read hooks read the RIP/thread through the facade (vCPU 0); they now resolve the dispatching vCPU (`active_cpu()` / `dispatch_on_cpu()`), which also closes a data race on the analyzer's cpuid hook.
- **4 — validation + profiling (done):**
  - `src/samples/stress-sample`: 8 threads hammering interlocked ops, a `CRITICAL_SECTION`-guarded counter, a semaphore producer/consumer, event ping-pong, and `VirtualAlloc` churn with write/read verification — each asserting a deterministic invariant. Passes at `--vcpus 1..8` (x64 and WoW64).
  - `SOGEN_LOCK_PROFILE` instruments the BEL (acquisitions / contended / wait / held). Measured contention: **1.4 %** (open-iw5, N=2) / 2.6 % (N=4) for a real game vs 43–54 % for a pure-synchronization worst case → the BEL is not a bottleneck for the target workloads, so **no lock refinement is warranted** (documented in the design doc).
- **5 — KVM parity / linux-emulator:** not in this PR.

## Testing

- `analyzer -s test-sample.exe` passes on unicorn (N=1) and WHP (N=1..4), x64 and WoW64.
- `stress-sample` passes at `--vcpus 1..8` for both bitnesses.
- open-iw5 / MW3 boots to menu and self-exits at `--vcpus 2`.
- `N=1` path is unchanged (runs through the new scheduler with an uncontended BEL).

## Known limitations

- Phase 3 is *converged, not closed* — more workloads will surface more cross-vCPU edge cases; the stress sample + BEL profiler are the nets for catching them.
- KVM remains single-vCPU (`supports_multiple_vcpus()` is false) until Phase 5.

### DIFF
--- a/src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
+++ b/src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
@@ -116,7 +116,7 @@ namespace sogen::icicle
         {
             uint64_t address{};
             uint64_t size{};
-            memory_access_hook_callback callback{};
+            std::function<void(uint64_t address, const void* data, size_t size)> callback{};
             bool is_read{};
         };
 
@@ -316,6 +316,14 @@ namespace sogen::icicle
             ice(res, "Failed to apply permissions");
         }
 
+        // The raw icicle hook wrappers are captureless function pointers, so the
+        // triggering CPU is bound into the stored function up front.
+        template <typename Ret, typename... Args>
+        std::function<Ret(Args...)> bind_cpu(std::function<Ret(cpu_interface&, Args...)> callback)
+        {
+            return [this, c = std::move(callback)](Args... args) { return c(*this, std::forward<Args>(args)...); };
+        }
+
         emulator_hook* hook_instruction(int instruction_type, instruction_hook_callback callback) override
         {
             if (static_cast<x86_hookable_instructions>(instruction_type) != x86_hookable_instructions::syscall)
@@ -324,7 +332,7 @@ namespace sogen::icicle
                 return nullptr;
             }
 
-            auto obj = make_function_object(std::move(callback), this->is_in_hook_);
+            auto obj = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = obj.get();
 
             const auto invoker = +[](void* cb) {
@@ -340,7 +348,7 @@ namespace sogen::icicle
 
         emulator_hook* hook_basic_block(basic_block_hook_callback callback) override
         {
-            auto object = make_function_object(std::move(callback), this->is_in_hook_);
+            auto object = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = object.get();
             auto* wrapper = +[](void* user, const uint64_t addr, const uint64_t instructions) {
                 basic_block block{};
@@ -359,7 +367,7 @@ namespace sogen::icicle
 
         emulator_hook* hook_interrupt(interrupt_hook_callback callback) override
         {
-            auto obj = make_function_object(std::move(callback), this->is_in_hook_);
+            auto obj = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = obj.get();
             auto* wrapper = +[](void* user, const int32_t code) {
                 const auto& func = *static_cast<decltype(ptr)>(user);
@@ -374,7 +382,7 @@ namespace sogen::icicle
 
         emulator_hook* hook_memory_violation(memory_violation_hook_callback callback) override
         {
-            auto obj = make_function_object(std::move(callback), this->is_in_hook_);
+            auto obj = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = obj.get();
             auto* wrapper = +[](void* user, const uint64_t address, const uint8_t operation, const int32_t unmapped) -> int32_t {
                 const auto violation_type = unmapped //
@@ -396,7 +404,7 @@ namespace sogen::icicle
 
         emulator_hook* hook_memory_execution(const uint64_t address, memory_execution_hook_callback callback) override
         {
-            auto object = make_function_object(std::move(callback), this->is_in_hook_);
+            auto object = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = object.get();
             auto* wrapper = +[](void* user, const uint64_t addr) {
                 const auto& func = *static_cast<decltype(ptr)>(user);
@@ -417,7 +425,7 @@ namespace sogen::icicle
                 return this->hook_memory_execution(address, std::move(callback));
             }
 
-            auto object = make_function_object(std::move(callback), this->is_in_hook_);
+            auto object = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = object.get();
             auto* wrapper = +[](void* user, const uint64_t addr) {
                 const auto& func = *static_cast<decltype(ptr)>(user);
@@ -432,7 +440,7 @@ namespace sogen::icicle
 
         emulator_hook* hook_memory_execution(memory_execution_hook_callback callback) override
         {
-            auto object = make_function_object(std::move(callback), this->is_in_hook_);
+            auto object = make_function_object(this->bind_cpu(std::move(callback)), this->is_in_hook_);
             auto* ptr = object.get();
             auto* wrapper = +[](void* user, const uint64_t addr) {
                 const auto& func = *static_cast<decltype(ptr)>(user);
@@ -450,7 +458,7 @@ namespace sogen::icicle
             return this->try_install_memory_access_hook(memory_access_hook{
                 .address = address,
                 .size = size,
-                .callback = std::move(callback),
+                .callback = this->bind_cpu(std::move(callback)),
                 .is_read = true,
             });
         }
@@ -460,7 +468,7 @@ namespace sogen::icicle
             return this->try_install_memory_access_hook(memory_access_hook{
                 .address = address,
                 .size = size,
-                .callback = std::move(callback),
+                .callback = this->bind_cpu(std::move(callback)),
                 .is_read = false,
             });
         }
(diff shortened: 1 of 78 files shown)


momo5502:
@claude review

src/windows-emulator/kusd_mmio.hpp
momo5502: these two methods don't lock. we should probably get rid of them and replace them with an access method that accepts a lambda, that receives the kusd obj and before calling the lambda, we lock the mutex, somethinig similar to this:

```C++
template<typename F>
void get(const F& functor)
{
std::lock_guard _{this->mutex_};
functor(this->kusd_);
}
```

i'm not sure if there can be cases where the caller of that method would already hold a lock. if so, we might need a deadlock prevention mechanism. some overload that works with std::lock or so. but not sure if necessary

src/tools/sandbox/main.cpp
```diff
@@ -62,7 +64,16 @@ namespace sogen::sandbox
                 .arguments = parse_arguments(args),
             };
 
+#ifdef _WIN32
+            // One vCPU per host core; WHP supports at most 64 per partition.
+            const auto vcpu_count = std::clamp(std::thread::hardware_concurrency(), 1u, 64u);
```
momo5502: there should be an env that allows overriding that

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
```diff
@@ -1168,6 +1247,11 @@ namespace sogen
             std::array<bool, SDL_SCANCODE_COUNT> key_down_{};
             std::unordered_map<hwnd, window_state> windows_{};
             std::unordered_map<SDL_WindowID, hwnd> guest_by_window_id_{};
+
+            std::mutex command_mutex_{};
+            std::vector<std::function<void()>> commands_{};
+            std::thread::id ui_thread_id_{};
```
momo5502, line 1253: is this thread needed when vCPU count is 1? does it make sense to keep it separate in that case?
in general, i think when vCPU count is 1, we should try to stay entirely single threaded.
is this easily possible? does this make sense?

src/windows-emulator/kusd_mmio.hpp
momo5502: Done in dcd1e8bf: replaced both `get()` overloads with a locked `access(functor)` template that takes `mutex_`, invokes the functor with `kusd_`, and forwards its result. Converted all call sites (the minidump multi-field write and the `ActiveProcessorCount`/`QpcFrequency`/`NtSystemRoot` reads). No caller already holds `mutex_`, so the `std::lock`-style deadlock-prevention overload is not needed for now.

src/tools/sandbox/main.cpp
momo5502: Done in 59d7162e: `EMULATOR_VCPU_COUNT` now overrides the auto-detected count (clamped to 1..64), matching the existing `EMULATOR_ROOT` environment convention.

src/windows-emulator/ui_backends/sdl_ui_backend.cpp
momo5502: The SDL backend does not spawn a thread here — `ui_thread_id_`/`command_mutex_`/`commands_` is only the marshaling for cross-thread SDL calls coming from vCPU workers, and it is inert at `--vcpus 1` (no workers exist). The one background thread at N=1 is the preemption/interrupt thread, which predates this PR and is required for the WHP backend: it runs guest code natively and cannot preempt cooperatively (its `hook_basic_block` is a no-op stub), so a host thread is the only way to time-slice guest threads multiplexed on a single vCPU. So N=1 already stays single-threaded apart from that pre-existing thread; left as-is.

momo5502:
@claude review

src/backends/icicle-emulator/icicle_x86_64_emulator.cpp
```diff
@@ -116,7 +116,7 @@ namespace sogen::icicle
         {
             uint64_t address{};
             uint64_t size{};
-            memory_access_hook_callback callback{};
+            std::function<void(uint64_t address, const void* data, size_t size)> callback{};
```
momo5502: can we name the callback type, like it was done before

src/windows-emulator/windows_emulator.hpp
```diff
@@ -92,6 +93,10 @@ namespace sogen
         bool use_relative_time{false};
         bool use_instruction_precision{true};
 
+        // Number of virtual CPUs to run guest threads on. Values above 1 require a
+        // backend where supports_multiple_vcpus() is true (see docs/multi-vcpu-design.md).
+        uint32_t vcpu_count{1};
```
momo5502: i don't think this should be a setting, this should be derived from the emulator backend. it knows how many vcpus there are, no need to specify twice

src/windows-emulator/window_destroy_orchestrator.hpp
```diff
@@ -37,8 +38,9 @@ namespace sogen
         void finalize_frame(window_destroy_frame& frame, const window& win) const;
 
         window_destroy_state& state_;
-        x86_64_emulator& emu_;
+        x86_64_cpu& emu_;
         process_context& proc_;
+        const emulator_thread& thread_;
```
momo5502, line 43: is it good there is a thread reference here? can this not become a dangling pointer? or does the thread take care of destroying this object, as in, is it guaranteed that the thread outlives this object?

momo5502: if not, maybe it's better to store the thread id instead

src/python-bindings/sogen_wrappers.cpp
momo5502, line 1: are there any docs to update regarding the changes in the python bindings?

src/windows-emulator/kernel_lock.hpp
```diff
@@ -0,0 +1,124 @@
...
+        std::mutex mutex_{};
+        std::atomic<std::thread::id> owner_{};
+
+        // Guarded by mutex_ (written/read only by the current holder).
+        clock::time_point held_since_{};
+
+        std::atomic<uint64_t> acquisitions_{};
+        std::atomic<uint64_t> contended_{};
+        std::atomic<uint64_t> wait_nanos_{};
+        std::atomic<uint64_t> held_nanos_{};
```
momo5502, line 121: why not std::atomic<std::chrono::high_resolution_clock::duration> ? then there is no need to keep the nanos in the name, cause the type is explicit

src/windows-emulator/exception_dispatch.cpp
```diff
@@ -244,14 +244,25 @@ namespace sogen
         return false;
     }
 
-    void dispatch_exception(windows_emulator& win_emu, const DWORD status, const std::vector<EmulatorTraits<Emu64>::ULONG_PTR>& parameters)
+    void dispatch_exception(windows_emulator& win_emu, vcpu_context& vcpu, const DWORD status,
+                            const std::vector<EmulatorTraits<Emu64>::ULONG_PTR>& parameters)
     {
+        auto& thread = vcpu.thread();
+
+        win_emu.record_exception_trace({
```
momo5502, line 252: what's this?

src/windows-emulator/windows_emulator.cpp
```diff
@@ -1265,15 +1502,85 @@ namespace sogen
         if (event.message == WM_CLOSE || event.message == WM_COMMAND || event.message == WM_KEYDOWN || event.message == WM_LBUTTONDOWN ||
             event.message == WM_LBUTTONUP)
         {
-            this->switch_thread_ = true;
-            this->emu().stop();
+            // Kick the vCPU currently running the target thread so it promptly re-checks its message
+            // queue; if the thread is not running on any vCPU right now, fall back to vCPU 0 to force a
+            // reschedule that can pick up the now-ready thread.
+            auto* running_on = find_vcpu_running_thread(*this, *thread);
```
momo5502, line 1503: can this race? while we have kicked the CPU, another CPU might have picked up the thread? or are we within a lock that guards from that?

PR #1180, opened 2026-07-04
ci: make the icicle cargo build resilient to network flakiness

momo5502:
CI frequently fails fetching the icicle git deps / crates.io index with transient network errors (curl 56, OpenSSL "unexpected eof while reading").

Two complementary layers:
- **`.cargo/config.toml`** — `net.retry=10`, `git-fetch-with-cli` (system git is more robust than libgit2 for the git deps), `http.multiplexing=false` to avoid mid-stream HTTP/2 EOFs.
- **`cargo_build_with_retry.cmake`** — wraps the ExternalProject build step in a process-level retry loop (5 attempts, 15s backoff) so a whole-invocation failure re-runs instead of failing the build.

Validated locally: config accepted, git deps fetched via CLI, clean build; retry loop confirmed to retry on failure and pass through on success.


src/backends/icicle-emulator/icicle-bridge/cargo_build_with_retry.cmake
```diff
@@ -0,0 +1,41 @@
...
+if(CARGO_TRIPLE)
+  list(APPEND cargo_args --target=${CARGO_TRIPLE})
+endif()
+
+set(attempt 1)
+while(TRUE)
+  message(STATUS "icicle: cargo build (attempt ${attempt}/${CARGO_RETRIES})")
+
+  execute_process(
+    COMMAND ${CMAKE_COMMAND} -E env "CARGO_TARGET_DIR=${CARGO_TARGET_DIR}" -- cargo ${cargo_args}
```
momo5502: this will also retry if there are build failures right?
is it possible to only retry on dependency downloads?

PR #1181, opened 2026-07-04
windows-emulator: security hardening for GPU bridge and file system

mjkzy:
fixes several sandbox boundary issues found during a security audit of the windows-emulator, focused on the GPU bridge IOCTL handlers and syscall layer. this was done with my local LLM experimenting for auditing, letting Claude Opus confirm and write the fixes for sending here.

- **destroy_device/destroy_instance UAF**: direct_mappings_ entries survived
    device teardown, leaving guest VA aliases to freed Vulkan memory. this is the best solution for security purposes, but it can clean all mappings and is **not** instance specific.
- **query pool stride overflow**: guest-controlled stride in
    get_query_pool_results could overflow the host heap buffer
- **specialization map entry OOB**: invalid offset+size forwarded to host
    vkCreateGraphicsPipelines, reading past the specialization data buffer
- **file rename info leak**: exists-check used untranslated guest path,
    letting guest probe host filesystem
- **cmd_begin_rendering DoS**: uint32_t wrap in color_attachment_count
    bypassed the bounds check

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -396,6 +396,16 @@ namespace sogen
 
                 const auto request = emulator_object<gpu_bridge::destroy_instance_request>{win_emu.emu(), context.input_buffer}.read();
 
+                // destroy_instance internally tears down all devices for this instance via erase_device,
+                // which frees Vulkan memory without consulting direct_mappings_. Release all guest VA
+                // aliases first to prevent stale mappings to freed host pages.
+                for (auto it = this->direct_mappings_.begin(); it != this->direct_mappings_.end();)
+                {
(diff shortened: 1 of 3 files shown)


momo5502:
@claude review

momo5502:
big thanks for those fixes ❤️

formatting requires clang-format 21

momo5502 (approved):

momo5502:
it saddens me that these bugs exist, but i guess that's the price we pay for the speed AI offers.

mjkzy:
> it saddens me that these bugs exist, but i guess that's the price we pay for the speed AI offers.

indeed. it's impressive Claude can work fast, but working fast, it can fail to write secure code. for the greater of Sogen, addressing security every once in a blue moon should make the sandbox safer for everyone 😄

PR #1183, opened 2026-07-04
ci: stabilize GitHub Pages deployment against transient failures

momo5502:
The `Deploy Page` job intermittently fails with **"Deployment failed, try again later"** from the GitHub Pages backend ([example](https://github.com/momo5502/sogen/actions/runs/28715238479/job/85157936657)). The artifact uploads fine; the deployment status poll comes back failed — a transient Pages service issue, not a build problem.

Two mitigations on the `deploy-page` job:
- **Concurrency group** (`group: pages-deploy`, `cancel-in-progress: false`) so two runs never deploy to the `github-pages` environment simultaneously — a common cause of this error. In-progress deploys are allowed to finish rather than being cancelled.
- **Retry** the `actions/deploy-pages@v5` step up to 3 times (each attempt creates a fresh deployment, 45s apart). The job only fails if the final attempt fails; a success short-circuits the rest.

Note: the deploy job is main-only, so this path is only exercised after merge.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -126,6 +126,12 @@ jobs:
     needs: [ci]
     if: ${{ github.repository_owner == 'momo5502' && github.ref == 'refs/heads/main' && needs.ci.result == 'success' }}
     runs-on: ubuntu-24.04
+    # Serialize publishes so two runs never upload to PyPI at once (a concurrent
+    # upload of the same version races and one attempt fails). Let an in-progress
+    # publish finish rather than cancelling it mid-upload.
+    concurrency:
+      group: pypi-publish
+      cancel-in-progress: false
     permissions:
       contents: read
       actions: read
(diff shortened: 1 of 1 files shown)


.github/workflows/build-push.yml
```diff
@@ -66,6 +66,12 @@ jobs:
     needs: [ci]
     if: ${{ github.repository_owner == 'momo5502' && github.ref == 'refs/heads/main' && needs.ci.result == 'success' }}
     runs-on: ubuntu-24.04
+    # Serialize Pages deployments so two runs never deploy to the github-pages
+    # environment at once (a common cause of "Deployment failed, try again later").
+    # Don't cancel an in-progress deploy — let it finish rather than leaving Pages half-updated.
+    concurrency:
+      group: pages-deploy
```
momo5502, line 73: shoult maybe do that for the python deploy as well?

PR #1184, opened 2026-07-04
Refactor emulator stack allocation leak tracking

Igoorx:
This PR replaces the assertion-based stack-allocation checks with runtime errors and adds allocation-site diagnostics for leaks during callback completion.

Example:
image

### DIFF
--- a/src/emulator/typed_cpu.hpp
+++ b/src/emulator/typed_cpu.hpp
@@ -4,8 +4,13 @@
 #include "memory_interface.hpp"
 #include "serialization.hpp"
 
+#include <algorithm>
+#include <array>
 #include <limits>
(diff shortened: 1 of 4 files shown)


momo5502:
@claude review

Igoorx:
> That's likely fine given the emulator is already stopped/erroring out in that path, but worth confirming it's intentional rather than an oversight, since it's a behavior change from the old assert-based check (which fired unconditionally, in debug builds, on any leaked destruction).

Error stops are unrecoverable, so this is fine. That was also partially a bug, because if an exception occurred before the stack allocations were freed, the assertion would fail, swallowing the exception. This new design solves that.

PR #1186, opened 2026-07-07
windows-emulator: fix section-map host OOB read + validate KsecDD ioctl buffers

momo5502:
Two sandbox-boundary hardening fixes, following #1181.

## 1. Host OOB read in `NtMapViewOfSection` (file-backed sections)

On the **file-backed** mapping path, the guest-controlled `SectionOffset` is only clamped to `>= 0` and never checked against the file size. When the offset exceeds the file length, the unsigned subtraction

```cpp
const auto size = static_cast<size_t>(file_data.size() - offset);
```

underflows to a near-`2^64` value, and

```cpp
c.emu.write_memory(address, file_data.data() + offset, size);
```

copies from `file_data.data() + offset` — a host pointer past the end of the `std::vector` — into the guest view. The copy **source** is host-side and not MMU-checked, so this reads out-of-bounds host heap memory into guest-readable memory (host info disclosure, or a crash).

The sibling **pagefile-backed** path already rejects this (`aligned_offset >= backing_size`); this adds the equivalent bound on the file-backed path before the copy size is computed.

**Trigger:** `NtCreateFile` on a small file → `NtCreateSection` file-backed data section (no `SEC_IMAGE`) → `NtMapViewOfSection` with `SectionOffset` larger than the file.

## 2. Unvalidated buffers in `\Device\KsecDD` ioctl 0x390400

The handler read a `USHORT` at `input+6` and 16 bytes at `input+0x30`, and wrote a fixed 224/216-byte response to `output_buffer`, without checking any guest-controlled buffer pointer or length. A guest could point `output_buffer` at a small mapped buffer and have the handler overwrite bytes past it. Now validates `input_buffer_length` covers the read fields and rejects an undersized output buffer, mirroring the guards the GPU bridge applies to every fixed-size response.

### DIFF
--- a/src/windows-emulator/devices/security_support_provider.cpp
+++ b/src/windows-emulator/devices/security_support_provider.cpp
@@ -3,11 +3,27 @@
 
 #include "../windows_emulator.hpp"
 
+#include <utils/string.hpp>
+
 namespace sogen
 {
 
     namespace
     {
+        // Partial layout of the KsecDD ioctl 0x390400 request, covering the two fields the handler
+        // consumes: the operation selector and the requested algorithm name.
+        struct ksec_algorithm_request
+        {
+            std::array<uint8_t, 6> reserved0;
+            uint16_t operation;
+            std::array<uint8_t, 0x28> reserved1;
+            std::array<char16_t, 8> algorithm_name;
+        };
+
+        static_assert(offsetof(ksec_algorithm_request, operation) == 6);
+        static_assert(offsetof(ksec_algorithm_request, algorithm_name) == 0x30);
+        static_assert(sizeof(ksec_algorithm_request) == 0x40);
+
         struct security_support_provider : stateless_device
         {
             // RNG Microsoft Primitive Provider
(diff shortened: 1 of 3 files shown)


momo5502:
@claude review

src/windows-emulator/devices/security_support_provider.cpp
```diff
@@ -48,37 +48,50 @@ namespace sogen
                     return STATUS_NOT_SUPPORTED;
                 }
 
+                // The input/output buffers and their lengths are guest-controlled. Validate the input
+                // covers the fields read below (USHORT at +6, 16 bytes at +0x30) before touching it.
+                if (!c.input_buffer || c.input_buffer_length < 0x40)
```
momo5502: what's the 0x40 here? can we extract that into a constant, or even better, is there a struct we can base that on and use sizeof it here?

src/windows-emulator/devices/security_support_provider.cpp
```diff
@@ -48,37 +48,50 @@ namespace sogen
...
                 if (operation == 2)
                 {
                     std::array<char16_t, 8> alg_name_buffer{};
                     win_emu.emu().read_memory(c.input_buffer + 0x30, alg_name_buffer.data(), sizeof(alg_name_buffer));
 
-                    const std::u16string algorithm_name(alg_name_buffer.data());
+                    // alg_name_buffer is guest-controlled and may not be NUL-terminated; bound the view to
+                    // the fixed array so we never scan past it looking for a terminator.
+                    const std::u16string_view raw_name(alg_name_buffer.data(), alg_name_buffer.size());
+                    const auto algorithm_name = raw_name.substr(0, raw_name.find(u'\0'));
```
momo5502: can we maybe extract that into a util, to get a string_view of type T based on a span<const T>

src/windows-emulator/devices/security_support_provider.cpp
momo5502: Done in a0a9e03b — replaced the magic `0x40` with `sizeof(ksec_algorithm_request)`, a struct modelling the request layout (operation at +6, algorithm name at +0x30) with `static_assert`s pinning the field offsets and total size. The handler now reads the whole struct in one go.

src/windows-emulator/devices/security_support_provider.cpp
momo5502: Done in a0a9e03b — extracted `utils::string::to_string_view<T>(std::span<const T>)` in `src/common/utils/string.hpp`. It builds a `basic_string_view<T>` bounded to the span and truncates at the first NUL, so it never scans past the buffer. The KsecDD handler uses it for the algorithm name.

PR #1187, opened 2026-07-07
windows-emulator: bound guest-controlled allocations in device handlers

momo5502:
Robustness hardening for the device IOCTL layer (companion to the memory-safety fixes in #1186).

Several handlers sized a host allocation directly from a guest-controlled count/length with no upper bound, so a malformed request could force a multi-gigabyte allocation and an uncaught `bad_alloc`. **None of these are memory-safety issues** — a guest can already terminate its own emulated process — but the emulator should reject bad input rather than crash, and the codebase already caps such allocations elsewhere (e.g. the 256 MB descriptor-update limit).

- **vulkan_host `create_device`**: reject `queue_count` above a sane per-family max before it sizes the priorities vector. A single queue-create entry with `queue_count = 0xFFFFFFFF` otherwise requests a ~16 GB `std::vector<float>`.
- **network_store_interface `write_zeroes`**: zero the guest region in bounded chunks instead of materializing the full 64-bit guest `size` as one host allocation.
- **mount_point_manager `get_drive_letter`**: cap the guest `input_buffer_length` (a device name is tiny) before reading it into a host buffer.
- **afd `send_datagram`**: reject a negative/oversized `RemoteAddressLength` (a signed `LONG`) before it is cast to `size_t` and used as a read length.

### DIFF
--- a/src/windows-emulator/devices/afd_endpoint.cpp
+++ b/src/windows-emulator/devices/afd_endpoint.cpp
@@ -1168,6 +1168,14 @@ namespace sogen
                 const auto send_info = emu.read_memory<AFD_SEND_DATAGRAM_INFO<Traits>>(c.input_buffer);
                 const auto buffer = emu.read_memory<EMU_WSABUF<Traits>>(send_info.BufferArray);
 
+                // RemoteAddressLength is a signed LONG; a negative or oversized value would turn into a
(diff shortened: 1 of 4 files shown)


momo5502:
@claude review

momo5502:
Not really security critical, but at the edge of it, so let's just add the tag

Issue #1055, opened 2026-06-15
Sandbox Escape via Arbitrary R/W in Vulkan Host


ioncodes:
Hey, as per Discord here's what I found. Wrote a quick n dirty PoC for it too: https://github.com/ioncodes/sogen-sandbox-escape. I have not looked at it in depth, please let me know if I forgot something important lol

Here's a short outline taken directly from the README:

[`vulkan_host::upload_memory`](https://github.com/momo5502/sogen/blob/8e6e74adccb7bcd125bd8f351cb841ee12c3bcb7/src/windows-emulator/devices/vulkan_host.cpp#L2514) / [`download_memory`](https://github.com/momo5502/sogen/blob/8e6e74adccb7bcd125bd8f351cb841ee12c3bcb7/src/windows-emulator/devices/vulkan_host.cpp#L2491) forward the `offset` ([guest controlled](https://github.com/momo5502/sogen/blob/8e6e74adccb7bcd125bd8f351cb841ee12c3bcb7/src/windows-emulator/devices/gpu_bridge.cpp#L1204)) straight to `vkMapMemory(offset, size)` and never validate it against the [allocation size](https://github.com/momo5502/sogen/blob/8e6e74adccb7bcd125bd8f351cb841ee12c3bcb7/src/windows-emulator/devices/vulkan_host.cpp#L2365)

```cpp
// not verbatim ofc i copied the relevant lines out

// allocate_memory(): the guest chooses the size
memories.emplace(id, memory_data{ .handle = memory, .allocation_size = size /* e.g. 64 */ });

// but upload_memory()/download_memory() never validate allocation_size:
const size_t copy_bytes = std::min(static_cast<size_t>(size), data_size);
dev->second.map_memory(dev->second.handle, mem->second.handle, offset, size, 0, &mapped); // mapped = base + offset
std::memcpy(mapped, data, copy_bytes); // copies past the end of the allocation
```

Allocating that memory makes the analyzer map it into its own address space at some host
address, I call it `base`. Every transfer then reads/writes at `base + offset`, and we fully
control `offset`. So to hit any address `target` we want, we just send `offset = target - base` (wraps ofc).

Effectively gives you read/write primitive. The PoC illustrates an example on how to leverage that for code execution on host, but that was more for shits n giggles. I even disabled ASLR for it

<img width="1248" height="497" alt="Image" src="https://github.com/user-attachments/assets/b98d9784-438f-44ca-9270-7dda85414481" />

PR #1188, opened 2026-07-07
windows-emulator: fix two guest-triggerable host OOB writes in the GPU host

momo5502:
Two host out-of-bounds writes reachable from untrusted guest GPU commands, found while re-auditing `gpu_bridge` / `vulkan_host` for more of the arbitrary-R/W class from #1055. Both let the guest drive a write past a host buffer.

## 1. Swapchain readback buffer overflow — HIGH

`create_swapchain` sizes the host readback buffer at `width * height * 4` (`vulkan_host.cpp:3489`), but the presentable images are created with the **guest-chosen** `VkFormat` (`:3446`, forwarded unvalidated from `gpu_bridge.cpp:1627`). At present time, `vkCmdCopyImageToBuffer` copies the full extent with the image's real texel size, writing `width * height * texelSize(format)` bytes into the 4-byte-per-pixel buffer.

A guest creating a swapchain with an 8-byte (`R16G16B16A16_SFLOAT`) or 16-byte (`R32G32B32A32_SFLOAT`) format triggers a **2×–4× overflow** of the readback buffer's backing memory on the next present — no rendering required, just create + acquire + present. On the software (SwiftShader) driver the readback allocation is plain host heap.

**Fix:** restrict swapchain formats to the 4-byte formats the readback/present path actually supports (it is hardcoded to BGRA8 end to end); reject others with `VK_ERROR_INITIALIZATION_FAILED`.

## 2. `get_query_pool_results` element-size gap — Medium

The only bound was `stride * query_count <= data_size` (bridge side, `gpu_bridge.cpp:1851`). That ignores the **actual result element size** and is skipped entirely when `stride == 0`. The per-query element can be larger than a guest-chosen small `stride` (via `VK_QUERY_RESULT_64_BIT` / `VK_QUERY_RESULT_WITH_AVAILABILITY_BIT`, or a pipeline-statistics pool), so `vkGetQueryPoolResults` writes past the host results vector. Driver-dependent, and the written bytes are driver counter data, hence Medium.

**Fix:** store the pool's query type and pipeline-statistics mask, compute the element size from the type + flags, and bound `(query_count - 1) * max(stride, element) + element` against `out_size` in `vulkan_host::get_query_pool_results` (covers `stride == 0`).

## Verification
- MSVC syntax-check (`/Zs`) of the translation unit passes.
- The #1055 fix (`is_memory_range_valid`) and the rest of the map/transfer/direct-mapping/descriptor paths were re-audited and are correctly bounded.

## Not included (noted for follow-up, not memory-safety)
- Cross-device object confusion: memory/command handlers don't verify `object.device_id == device`. Only reachable with >1 `VkDevice`; impact is driver UB / crash (DoS), not host corruption.
- `handle_map_memory_direct` returns `base + request.offset` with `offset` unvalidated — contained to the guest's own VA (beyond the alias faults on unmapped guest pages).

### DIFF
--- a/src/windows-emulator/devices/vulkan_host.cpp
+++ b/src/windows-emulator/devices/vulkan_host.cpp
@@ -4,6 +4,7 @@
 
 #include <algorithm>
 #include <array>
+#include <bit>
 #include <cmath>
 #include <cstddef>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

PR #1189, opened 2026-07-07
windows-emulator: harden GPU host against cross-device ids and out-of-range direct-map offsets

momo5502:
Defense-in-depth hardening for the GPU host, following up on the two host-OOB fixes in #1188. Neither item here is host memory corruption on its own (both are DoS/contained), but both feed attacker-controlled guest input to the driver or to pointer arithmetic without a check.

## 1. Cross-device object ownership

The memory handlers (`free_memory`, `map_memory`, `unmap_memory`, `upload_memory`, `download_memory`, `flush_mapped_memory_range`, `invalidate_mapped_memory_range`) and `bind_buffer_memory` / `bind_image_memory` looked up the guest-supplied object id **globally** and never verified it belongs to the passed `device`. `memory_data`/`buffer_data`/`image_data` all record their owning `device_id`, and the destroy paths already filter on it, but the per-call handlers did not.

**Trigger:** a guest that creates more than one `VkDevice` can name memory allocated on device A while passing device B, making the host call e.g. `vkMapMemory(deviceB, memoryFromA)` — Vulkan undefined behaviour (driver crash / UB). Bounds bookkeeping stayed correct (`allocation_size` is device-independent), so this is host-driver UB, not host-heap corruption.

**Fix:** reject when `object.device_id != device` in each handler (and `free_memory` ignores a cross-device free).

## 2. Out-of-range direct-map offset

`handle_map_memory_direct` returned `guest_address = base + request.offset` (both the already-aliased fast path and the fresh-map path) with `request.offset` never bounded against the mapping size, and the addition could wrap.

**Impact:** contained — the returned VA beyond the alias is unmapped guest memory (faults) or another guest allocation; the guest can't reach host bytes outside the aliased region. Robustness only.

**Fix:** return `guest_address = 0` for offsets outside the mapping, so the shim falls back to the bounded staging path (the existing can't-alias signal).

## Verification
- MSVC syntax-check (`/Zs`) of both translation units passes.
- `clang-format --dry-run --Werror` clean.
- Single-device usage (the normal case, e.g. DXVK) always passes the `device_id` check, so no functional change there.

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -1390,7 +1390,10 @@ namespace sogen
                 if (const auto existing = this->direct_mappings_.find(request.memory); existing != this->direct_mappings_.end())
                 {
                     response.vk_result = 0; // VK_SUCCESS
-                    response.guest_address = existing->second.guest_address + request.offset;
+                    // Only hand back an aliased address for offsets inside the mapping; an out-of-range
+                    // offset returns guest_address = 0 so the shim falls back to the bounded staging path.
+                    response.guest_address =
+                        (request.offset < existing->second.size) ? (existing->second.guest_address + request.offset) : 0;
                     return write_output(win_emu, context, response);
                 }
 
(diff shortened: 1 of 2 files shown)


momo5502:
@claude review

momo5502:
Good catch on the destroy handlers — addressed in da5d6fb8. Extended the `device_id == device` ownership check to all the singular `destroy_*(device, object)` handlers plus the matching `reset_*`/`set_event`/`free_command_buffer` handlers that call the driver with a cross-device handle (24 handlers total). Void destroy handlers ignore a cross-device request (matching `free_memory`); the value-returning reset/set handlers return `VK_ERROR_INITIALIZATION_FAILED`. MSVC `/Zs` and clang-format both clean.

PR #1190, opened 2026-07-07
spike: host-side fuzzing harness for IO device handlers

momo5502:
**Draft / spike.** A host-side fuzzing harness that drives the emulator's IO device handlers with untrusted guest bytes, to continuously catch the guest-data → host-memory bug class (OOB read/write, etc.) we've been fixing by hand. This fuzzes **the emulator itself** — distinct from the guest-application fuzzer in `src/fuzzer`.

## Design (engine-independent)
- **`sogen::fuzz::run(std::span<const uint8_t>)`** — the single entry point. Feeds bytes to a target `io_device` through a persistent, constructed-once headless emulator, then calls `io_control()`. Thin drivers on top:
  - `libfuzzer_main.cpp` → `LLVMFuzzerTestOneInput`
  - `standalone_main.cpp` → file replay (any toolchain; smoke / regression / crash repro)
- **`mock_emulator.hpp`** — a minimal `x86_64_emulator` backend: a guest-memory arena with no code execution and stubbed CPU/hooks. No unicorn, deterministic, ASAN-friendly.
- **`emulator_settings::load_registry`** (default `true`) — when `false`, the `windows_emulator` ctor skips the eager registry-hive parse, so the harness constructs a bare emulator with **no emulation root / registry on disk** and never `start()`s the guest.

## Build
`SOGEN_ENABLE_FUZZING` (default OFF) builds the fuzz targets. On clang it also instruments all code with `-fsanitize=fuzzer-no-link` so the fuzzer sees the emulator's own handlers — pair with `SOGEN_ENABLE_SANITIZER=On` for ASAN. The standalone replay driver builds on MSVC too.

## Status — proven end-to-end ✅
- **libFuzzer + ASAN runs**, driving `security_support_provider` at **~9,400 exec/s** in persistent mode; 133,989 instrumented PCs; coverage-guided (corpus grew, new edges found from the KsecDD seeds); **ASAN clean** on the hardened handler (exit 0).
- Built on Windows with clang-cl 21 + `/MT` (`SOGEN_BUILD_STATIC=ON SOGEN_STATIC_CRT=ON`, to match libFuzzer's static-CRT runtime). Linux is the intended CI/OSS-Fuzz host and is friction-free there.
- The standalone replay also runs on MSVC.

Windows-specific bring-up notes: link needs LLVM's `clang_rt.fuzzer` + `clang_rt.asan_dynamic` + `asan_static_runtime_thunk` (this MSVC BuildTools lacks the ASAN component, so an empty `stl_asan.lib` stub satisfies the STL auto-link directive); run with `ASAN_OPTIONS=detect_odr_violation=0` (benign duplicate-string-global false positive).

## Follow-ups (not in this spike)
- Structured/stateful input (TLV of ioctl sequences) + a dictionary, so coverage reaches deep handlers.
- More targets: GPU bridge (needs SwiftShader or a driver mock), AFD, the buffer-indexing syscalls.
- A scheduled Linux-clang `fuzz.yml` workflow with corpus persistence; eventual OSS-Fuzz.
- Validation gate: revert a known fix (e.g. the query-pool / section OOB) and confirm the fuzzer rediscovers it.


src/windows-emulator-fuzz/device_fuzz.cpp
```diff
@@ -0,0 +1,96 @@
...
+        ctx.input_buffer_length = static_cast<ULONG>(in_len);
+        ctx.output_buffer = out_buf;
+        ctx.output_buffer_length = static_cast<ULONG>(half);
+
+        try
+        {
+            // Guest memory access is MMU-checked and throws by design; only sanitizer aborts are real bugs.
+            (void)s.device->io_control(s.emu, ctx);
+        }
+        catch (...)
```
momo5502: we should not catch ..., only std::exception derived exceptions. with /EHa, SEH exceptions would be caught by this. we don't want that. SEH exceptions should be fatal

momo5502:
@claude review

src/emulator/logger.cpp
```diff
@@ -157,7 +157,9 @@ namespace sogen
         // terminal (e.g. --silent, or a Python wrapper capturing via callback).
         this->sink_(c, message);
 
-        if (!force && this->disable_output_)
+        // silent_ suppresses everything (even force-printed errors); disable_output_ suppresses all but
+        // force-printed messages. Sinks above still see every line.
```
momo5502: comment is useless? we can read the code?

src/windows-emulator-fuzz/device_fuzz.cpp
```diff
@@ -0,0 +1,158 @@
...
+                }
+            };
+
+            add([] { return create_security_support_provider(); });
+            add([] { return create_console_device(); });
+            add([] { return create_mount_point_manager(); });
+            add([] { return create_network_store_interface(); });
+            add([] { return create_afd_endpoint(false); });
+            add([] { return create_afd_async_connect_hlp(false); });
+            add([] { return create_gpu_bridge(); });
```
momo5502, line 88: i think a device registry might make sense, one central place to register devices, so they are visible to the fuzzer and the regular emulation environment. adding new devices will lead to them being forgotten here, for sure. but that's something for another PR.

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,136 @@
...
+# emulator ITSELF with untrusted guest input; it is not the guest-application fuzzer in src/fuzzer.
+# Runs nightly for an hour and can be dispatched manually. Not attached to PRs.
+
+on:
+  # TEMPORARY: run on push to the spike branch so the workflow can be exercised before it lands on
+  # the default branch (workflow_dispatch's "Run workflow" button only appears once it's on main).
+  # Remove this push trigger before merge.
+  push:
+    branches:
+      - spike/handler-fuzzing
```
momo5502: should be removed

src/windows-emulator-fuzz/device_fuzz.cpp
momo5502: Fixed in 4264a57c — now `catch (const std::exception&)` so SEH stays fatal under /EHa.

src/emulator/logger.cpp
momo5502: Removed in 4264a57c.

.github/workflows/fuzz.yml
momo5502: Removed the temporary push trigger in 4264a57c; only `schedule` + `workflow_dispatch` remain.

src/windows-emulator-fuzz/device_fuzz.cpp
momo5502: Agreed — a central device registry shared by the fuzzer and the real emulation path would prevent devices being forgotten here. Leaving it for a follow-up PR as you suggested.

PR #1191, opened 2026-07-08
Handle unsupported DEP/NX protection

Igoorx:
This PR fixes sogen so it can run executables that don't support DEP/NX.

The issue was that these executables aren't supposed to honor executable memory permissions, but in sogen, if the sections weren't marked as executable, they would crash when executing the entry point. I tested this by compiling a program with `/NXCOMPAT:NO /SECTION:.text,RW` and confirmed that it works on native Windows but was failing on sogen.


momo5502:
@claude review

momo5502:
minor nits can be done in a follow up pr, if @Igoorx wants to

PR #1194, opened 2026-07-08
io_device: device registry + fuzzer corpus retention

momo5502:
Two fuzzing follow-ups from #1190.

## 1. Device registry (prototype)
Addresses the review note that the fuzzer keeping its own device list means new devices get forgotten. Replaces the hand-written `create_device()` name→factory if-chain with a single table:

```cpp
struct device_registration {
    std::span<const std::u16string_view> names;   // NT device names this handler serves
    std::unique_ptr<io_device> (*create)(bool is_32_bit);
};
std::span<const device_registration> get_device_registrations();
```

- `create_device(name, is_32_bit)` resolves a name through the table (same behavior, incl. the multi-name dummy/transport stubs).
- The handler **fuzzer** now iterates `get_device_registrations()` instead of a hardcoded list — a newly added device is fuzzed automatically.

Design choices to confirm (prototype):
- Uniform factory signature `(bool is_32_bit)`; nullary factories ignore the flag.
- Names live centrally in the table rather than each device declaring its own — open to moving them to the devices if preferred.
- The fuzzer enumerates **all** registered devices including stubs (dummy/transport/named_pipe). Harmless and matches "nothing forgotten", but could be filtered to real handlers.

## 2. Corpus retention (`fuzz.yml`)
The combined `actions/cache@v4` only saves on job **success**, but the fuzz job fails when it finds a crash — so the grown corpus from the most valuable runs was lost. Split into `cache/restore` + `cache/save` with `if: always()`, so the corpus persists across nightly runs regardless of a crash.

## Verification
- `windows-emulator` builds clean (full build, not just syntax) with the constexpr registry table.
- clang-format + YAML validated.

### DIFF
--- a/src/windows-emulator-fuzz/device_fuzz.cpp
+++ b/src/windows-emulator-fuzz/device_fuzz.cpp
@@ -55,18 +55,27 @@ namespace sogen::fuzz
             return windows_emulator{mock::make_mock_emulator(), settings, {}, std::move(interfaces)};
         }
 
-        // One persistent instance of every registered io_device, drawn from the shared device registry
-        // so newly added devices are fuzzed automatically. Constructing a device that reaches out to host
-        // resources (e.g. gpu_bridge's vulkan_host) may fail on a headless box, so guard each.
+        // One persistent instance of every registered io_device type, drawn from the shared device
+        // registry so newly added devices are fuzzed automatically. The registry maps names to factories
+        // and several names share a factory (e.g. the transport stub), so dedupe by factory pointer.
+        // Constructing a device that reaches out to host resources (e.g. gpu_bridge's vulkan_host) may
+        // fail on a headless box, so guard each.
         std::vector<std::unique_ptr<io_device>> make_devices()
         {
             std::vector<std::unique_ptr<io_device>> devices;
             const device_creation_context context{.is_32_bit = false};
-            for (const auto& registration : get_device_registrations())
+            std::vector<device_factory> seen;
+            for (const auto& [name, factory] : get_device_registry())
             {
+                if (std::ranges::find(seen, factory) != seen.end())
+                {
+                    continue;
+                }
+                seen.push_back(factory);
+
                 try
                 {
-                    if (auto d = registration.create(context))
+                    if (auto d = factory(context))
                     {
                         devices.push_back(std::move(d));
                     }
--- a/src/windows-emulator-fuzz/device_fuzz.cpp
+++ b/src/windows-emulator-fuzz/device_fuzz.cpp
@@ -11,6 +11,7 @@
 #include <exception>
 #include <filesystem>
 #include <memory>
+#include <unordered_set>
 
 #include <windows_emulator.hpp>
 #include <network/static_socket_factory.hpp>
@@ -64,14 +65,13 @@ namespace sogen::fuzz
         {
             std::vector<std::unique_ptr<io_device>> devices;
             const device_creation_context context{.is_32_bit = false};
-            std::vector<device_factory> seen;
+            std::unordered_set<device_factory> seen;
             for (const auto& [name, factory] : get_device_registry())
             {
-                if (std::ranges::find(seen, factory) != seen.end())
+                if (!seen.insert(factory).second)
                 {
                     continue;
                 }
-                seen.push_back(factory);
 
                 try
                 {
--- a/src/windows-emulator-fuzz/device_fuzz.cpp
+++ b/src/windows-emulator-fuzz/device_fuzz.cpp
@@ -61,11 +61,12 @@ namespace sogen::fuzz
         std::vector<std::unique_ptr<io_device>> make_devices()
         {
             std::vector<std::unique_ptr<io_device>> devices;
+            const device_creation_context context{.is_32_bit = false};
             for (const auto& registration : get_device_registrations())
             {
                 try
                 {
-                    if (auto d = registration.create(false))
+                    if (auto d = registration.create(context))
                     {
                         devices.push_back(std::move(d));
                     }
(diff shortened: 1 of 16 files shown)


src/windows-emulator/io_device.cpp
```diff
@@ -43,68 +43,99 @@ namespace sogen
...
         }
-
-        if (device == u"Nsi")
+        std::unique_ptr<io_device> make_nsi(bool)
         {
             return create_network_store_interface();
         }
-
-        if (device == u"Afd\\Endpoint")
+        std::unique_ptr<io_device> make_afd_endpoint(bool is_32_bit)
```
momo5502: all these make wrappers are useless. let's do something else, every create_* function receives a device_creation_context struct that for now only contains the is_32_bit bool, but later can be filled with additional data. each create function then has an identical signature and we don't need these overloads.

src/windows-emulator/io_device.hpp
```diff
@@ -137,6 +139,18 @@ namespace sogen
     bool needs_32_bit_devices(const windows_emulator& win_emu);
     std::unique_ptr<io_device> create_device(std::u16string_view device, bool is_32_bit);
 
+    // Single source of truth for the emulator's IO devices: which NT device names each handler serves
+    // and how to construct it. create_device() resolves a name through this table, and consumers that
+    // need every device (e.g. the handler fuzzer) can enumerate it, so a newly added device is visible
+    // everywhere without a second list to update.
+    struct device_registration
+    {
+        std::span<const std::u16string_view> names;
```
momo5502: not sure if I'm happy with span here, i see that it deduplicates the actual backing devices

momo5502: The span lets one factory serve several NT names without repeating the entry — `dummy` covers 6 names (CNG/RasAcd/PcwDrv/…) and the transport stub covers 4 (Tcp/Tcp6/Udp/RawIp), while each device is still constructed once. The tradeoff is that the names live in separate `constexpr` backing arrays next to the table rather than inline in the entry.

Alternatives if you'd prefer:
- Inline the names in the entry via a small fixed-capacity `std::array` + count (no backing arrays, but a fixed max and some padding).
- Have each device expose its own `names` so they sit with the handler rather than centrally — closer to self-registration.

Happy to go either way — do you have a preference, or is the span acceptable for now?

src/windows-emulator/io_device.cpp
momo5502: Done in aebef86f — dropped the make_* wrappers. Every create_* device factory now takes a `const device_creation_context&` (holds is_32_bit for now), so they slot into the registry directly. Full windows-emulator build passes.

src/windows-emulator/io_device.hpp
momo5502: Dropped the span entirely in 7a50b5bc — the registry is now a flat `unordered_map<u16string_view, device_factory>`. Registering a device is a single map entry, multiple names sharing a handler are just multiple entries pointing at the same factory, and the fuzzer dedupes by factory pointer. No more constexpr arrays / backing-storage concerns.

src/windows-emulator/io_device.cpp
```diff
@@ -43,71 +43,66 @@ namespace sogen
...
 
-        if (device == u"NamedPipe")
-        {
-            return std::make_unique<named_pipe>();
-        }
+    const std::unordered_map<std::u16string_view, device_factory>& get_device_registry()
+    {
+        using namespace std::string_view_literals;
+
+        static const std::unordered_map<std::u16string_view, device_factory> registry = {
```
momo5502: the map should not be unordered, it should be ordered, so std::map instead.so that the order is deterministic in the fuzzing enumeration

src/windows-emulator/io_device.hpp
```diff
@@ -184,7 +202,7 @@ namespace sogen
 
         void setup()
         {
-            this->device_ = create_device(this->device_name_, this->is_32_bit_);
+            this->device_ = create_device(this->device_name_, device_creation_context{.is_32_bit = this->is_32_bit_});
```
momo5502: don't inline the device_creation_context construction here, move it above

src/windows-emulator/io_device.cpp
momo5502: Switched to std::map in b3d52000 — enumeration is now ordered/deterministic.

src/windows-emulator/io_device.hpp
momo5502: Done in b3d52000 — the device_creation_context is now a named local above the create_device() call.

src/windows-emulator-fuzz/device_fuzz.cpp
```diff
@@ -61,31 +55,35 @@ namespace sogen::fuzz
...
+        // registry so newly added devices are fuzzed automatically. The registry maps names to factories
+        // and several names share a factory (e.g. the transport stub), so dedupe by factory pointer.
+        // Constructing a device that reaches out to host resources (e.g. gpu_bridge's vulkan_host) may
+        // fail on a headless box, so guard each.
         std::vector<std::unique_ptr<io_device>> make_devices()
         {
             std::vector<std::unique_ptr<io_device>> devices;
-            const auto add = [&](auto factory) {
+            const device_creation_context context{.is_32_bit = false};
+            std::vector<device_factory> seen;
```
momo5502: why not a set?

momo5502: Done in the latest push — switched the dedupe to std::unordered_set<device_factory> (function-pointer hashing works via std::hash<T*>; verified on clang too).

PR #1196, opened 2026-07-08
windows-emulator: bound output_buffer_length-driven host allocations

momo5502:
Found by the IO-device handler fuzzer once it started fuzzing the guest-declared `output_buffer_length` (the harness change on main). Several handlers size a **host** allocation directly from that guest-controlled length, so a bogus output length forces a multi-GB allocation → OOM.

## Fixes
- **transport stub** (`io_device.cpp`) — did `std::vector<std::byte>(output_buffer_length)` to zero-fill the output. Now zeros the guest region in bounded 4 KB chunks. Reachable via `\Device\Tcp`/`Tcp6`/`Udp`/`RawIp` (the fuzzer hit this one first: any ioctl code triggers it).
- **gpu_bridge** `get_physical_device_properties` (471), `get_physical_device_memory_properties` (1184), `get_surface_capabilities` (3043) — each staged a **small fixed-size struct** (< 1 KB) into an `output_buffer_length`-sized vector. Cap the staging allocation at `max_readback_bytes` (64 KiB). No behavior change for legit callers (their buffers are struct-sized); it only bounds the allocation.

Same DoS / unbounded-allocation class as #1187 and #1193.

## Verification
- Reproduced the OOM locally (`malloc(2795939533)` from the transport stub), applied the fixes, and re-ran the fuzzer 90 s unclamped — no OOM, stable (`DONE cov: 1641`, exit 0).
- Not touched here (let the fuzzer surface them next if reachable): the `output_buffer_length - sizeof(response_t)` underflow candidates and the `input_buffer_length`-sized allocations (the harness caps input at 4 KB, so those aren't reachable yet).

### DIFF
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -478,7 +478,7 @@ namespace sogen
                 }
 
                 win_emu.emu().write_memory(context.output_buffer, properties.data(), properties.size());
-                set_information(context, context.output_buffer_length);
+                set_information(context, static_cast<ULONG>(properties.size()));
                 return STATUS_SUCCESS;
             }
 
@@ -1191,7 +1191,7 @@ namespace sogen
                 }
 
                 win_emu.emu().write_memory(context.output_buffer, properties.data(), properties.size());
-                set_information(context, context.output_buffer_length);
+                set_information(context, static_cast<ULONG>(properties.size()));
                 return STATUS_SUCCESS;
             }
 
@@ -3068,7 +3068,7 @@ namespace sogen
                 }
 
                 win_emu.emu().write_memory(context.output_buffer, caps.data(), caps.size());
-                set_information(context, context.output_buffer_length);
+                set_information(context, static_cast<ULONG>(caps.size()));
                 return STATUS_SUCCESS;
             }
         };
--- a/src/windows-emulator/devices/gpu_bridge.cpp
+++ b/src/windows-emulator/devices/gpu_bridge.cpp
@@ -427,8 +427,9 @@ namespace sogen
 
                 const auto request = emulator_object<request_t>{win_emu.emu(), context.input_buffer}.read();
 
-                const auto array_capacity =
-                    static_cast<uint32_t>((context.output_buffer_length - sizeof(response_t)) / sizeof(gpu_bridge::object_id));
+                const auto array_capacity = static_cast<uint32_t>(
+                    std::min<uint64_t>((context.output_buffer_length - sizeof(response_t)) / sizeof(gpu_bridge::object_id),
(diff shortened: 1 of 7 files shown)


src/windows-emulator/io_device.cpp
```diff
@@ -29,8 +29,15 @@ namespace sogen
...
-                    win_emu.emu().write_memory(context.output_buffer, output.data(), output.size());
+                    // output_buffer_length is guest-controlled; zero the region in bounded chunks instead
+                    // of materializing the whole (possibly multi-GB) span as one host allocation.
+                    constexpr size_t chunk_size = 0x1000;
+                    const std::vector<std::byte> zeros(std::min<size_t>(context.output_buffer_length, chunk_size), std::byte{0});
+                    for (ULONG offset = 0; offset < context.output_buffer_length; offset += static_cast<ULONG>(zeros.size()))
+                    {
+                        const auto count = std::min<size_t>(zeros.size(), context.output_buffer_length - offset);
+                        win_emu.emu().write_memory(context.output_buffer + offset, zeros.data(), count);
+                    }
```
momo5502: i think what makes sense is a set_memory overload (similar to memset) on the emulator/memory interface that just does what this loop here does. i think we've had this pattern a few times.

src/windows-emulator/devices/gpu_bridge.cpp
```diff
@@ -3040,7 +3048,9 @@ namespace sogen
                     }
                 }
 
-                std::vector<std::byte> caps(context.output_buffer_length);
+                // output_buffer_length is guest-controlled; the response is a small fixed-size struct, so
+                // bound the staging allocation instead of trusting the declared length.
```
momo5502: these comments are always the same, make sure to not add useless comments. only the ones that are strictly necessary

src/windows-emulator/io_device.cpp
momo5502: Done in a7d860a8 — added `memory_interface::set_memory(address, value, size)` (memset-like, chunked fill, `uint8_t` value / `uint64_t` size), with a `typed_cpu` forwarder + the `x86_emulator` using-declaration to disambiguate. Migrated the transport stub, NSI `write_zeroes`, and both `system.cpp` zero-fills to it (the two system.cpp ones were also unbounded output-length allocations, now bounded too).

src/windows-emulator/devices/gpu_bridge.cpp
momo5502: Removed the three identical comments in a7d860a8; the max_readback_bytes constant already documents the cap.

momo5502:
@claude review

momo5502:
@claude review

momo5502:
@claude review

PR #1202, opened 2026-07-08
gpu_bridge: shrink download_memory staging cap in fuzz builds

momo5502:
## Why

The host-side fuzzer drives `download_memory` with bogus requests — there's no real GPU memory behind them — but the handler still allocates and **zero-fills** its staging buffer up to `max_memory_transfer_bytes` (256 MiB). Measured at **~121 ms per execution**. Coverage-guided fuzzing keeps those inputs (they hit new edges) and replays them across every worker, which measurably dominated throughput: dropping just this cap 256 MiB → 1 MiB recovered **~4×** exec/s in a CI A/B (per-worker ~700 → ~2,750).

## What

Emit `SOGEN_ENABLE_FUZZING=1/0` as a compile definition from CMake (in `compiler-env.cmake`, following the existing `SOGEN_BUILD_STATIC` / `SOGEN_ENABLE_RUST_CODE` pattern), and use it in `gpu_bridge.cpp` to cap the `download_memory` staging buffer at **1 MiB in fuzz builds only**:

```cpp
#if SOGEN_ENABLE_FUZZING
    static constexpr size_t max_memory_transfer_bytes = size_t{1} * 1024 * 1024;
#else
    static constexpr size_t max_memory_transfer_bytes = size_t{256} * 1024 * 1024;
#endif
```

The **shipped (non-fuzz) build keeps the 256 MiB cap unchanged** — this only affects `SOGEN_ENABLE_FUZZING=ON` builds. The code path is still fully exercised by the fuzzer; only the buffer size shrinks, so no coverage is lost.

## Verified

- Production build (`SOGEN_ENABLE_FUZZING=0`): compiles, uses 256 MiB.
- Fuzz build (`SOGEN_ENABLE_FUZZING=1`): replaying the `download_memory` OOM reproducer drops from **~121 ms/exec → ~0 ms/exec**.

Note: adding a new global compile definition triggers a one-time full rebuild when it lands (same as any change to `SOGEN_BUILD_STATIC` et al.).

### DIFF
--- a/cmake/compiler-env.cmake
+++ b/cmake/compiler-env.cmake
@@ -40,6 +40,16 @@ endif()
 
 ##########################################
 
+# Let handler code shrink fuzz-only tunables (e.g. gpu_bridge's bulk-transfer staging cap) so the
+# fuzzer doesn't burn most of its time on large allocations that are already proven bounded.
+if(SOGEN_ENABLE_FUZZING)
+    add_compile_definitions(SOGEN_ENABLE_FUZZING=1)
+else()
+    add_compile_definitions(SOGEN_ENABLE_FUZZING=0)
+endif()
+
+##########################################
+
 set(SOGEN_ENABLE_RUST OFF)
 if(SOGEN_ENABLE_RUST_CODE AND NOT MINGW AND NOT CMAKE_SYSTEM_NAME MATCHES "Emscripten")
   find_program(CARGO cargo)
--- a/cmake/compiler-env.cmake
+++ b/cmake/compiler-env.cmake
@@ -40,12 +40,8 @@ endif()
 
 ##########################################
 
-# Let handler code shrink fuzz-only tunables (e.g. gpu_bridge's bulk-transfer staging cap) so the
-# fuzzer doesn't burn most of its time on large allocations that are already proven bounded.
 if(SOGEN_ENABLE_FUZZING)
-    add_compile_definitions(SOGEN_ENABLE_FUZZING=1)
-else()
-    add_compile_definitions(SOGEN_ENABLE_FUZZING=0)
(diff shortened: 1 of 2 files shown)


cmake/compiler-env.cmake
```diff
@@ -40,6 +40,16 @@ endif()
 
 ##########################################
 
+# Let handler code shrink fuzz-only tunables (e.g. gpu_bridge's bulk-transfer staging cap) so the
+# fuzzer doesn't burn most of its time on large allocations that are already proven bounded.
+if(SOGEN_ENABLE_FUZZING)
+    add_compile_definitions(SOGEN_ENABLE_FUZZING=1)
+else()
+    add_compile_definitions(SOGEN_ENABLE_FUZZING=0)
```
momo5502: i think we should not define it at all otherwise, and then just use #ifdef instead of #if

cmake/compiler-env.cmake
```diff
@@ -40,6 +40,16 @@ endif()
 
 ##########################################
 
+# Let handler code shrink fuzz-only tunables (e.g. gpu_bridge's bulk-transfer staging cap) so the
+# fuzzer doesn't burn most of its time on large allocations that are already proven bounded.
```
momo5502: remove that comment

cmake/compiler-env.cmake
momo5502: Done — only defined when fuzzing now, and switched to `#ifdef`. Bonus: dropping the non-fuzz `=0` branch means normal builds no longer get a new global define, so no forced full rebuild either.

cmake/compiler-env.cmake
momo5502: Removed.

PR #1206, opened 2026-07-10
Map image sections from existing loaded modules

Igoorx:
This PR fixes secondary image mappings for modules that are already loaded.

When `NtMapViewOfSection` maps another view of an existing loader-managed image, the new view must preserve the relocation base of the original image. Previously, the emulator treated it as an independent module load and relocated it against the new view address, producing incorrect image contents.

I've used the following code to test this change: [test_manualmap.cpp](https://pastebin.com/0cBwL2M8)

### DIFF
--- a/src/windows-emulator/module/module_manager.cpp
+++ b/src/windows-emulator/module/module_manager.cpp
@@ -188,9 +188,10 @@ namespace sogen
     }
 
     // PE32 Mapping Strategy Implementation
-    mapped_module pe32_mapping_strategy::map_from_file(memory_manager& memory, std::filesystem::path file, windows_path module_path)
+    mapped_module pe32_mapping_strategy::map_from_file(memory_manager& memory, std::filesystem::path file, windows_path module_path,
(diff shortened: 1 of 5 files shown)


Igoorx:
By the way, I think it’s unclear whether this is the right abstraction for this fix, but it was enough to address the observable behavior in the tests. So even if it isn’t the correct way to fix it, I believe it’s a good stopgap. If you think there’s a better way to address this issue, feel free to say so.

momo5502:
@claude review

Igoorx:
No changes should be necessary.

> section.cpp:404-415 reaches directly into `mod_manager.modules()` and the PEB internals to find “the loaded module for this path.” `module_manager` already has similar find-by-path logic in `map_local_module` (`module_manager.cpp:513-521`, used when `!allow_duplicate`) and a `get_module_load_count_by_path` helper.

The code in `section.cpp` is very specific to that purpose, so I don’t think it makes sense to move it to `module_manager` or to make a change just for it.

PR #1211, opened 2026-07-10
Report FastFail availability and fix it on WHP

Igoorx:
This PR fixes the fastfail interrupt so it works on WHP and is correctly used for `std::abort`.


momo5502:
seems like the sample on the web is broekn. not sure if because of this

momo5502:
> seems like the sample on the web is broekn. not sure if because of this

ok, the broken web sample has nothing to do with this. it's an LLVM/emscripten compiler bug triggered by the CLI11 dependency.

Issue #1170, opened 2026-07-02
Add Steam bridge


momo5502:
Steam games might need to be able to communicate with Steam outside the sandbox/emulator.
I think much of proton can be copied here. Maybe the sandbox can even be built ontop of proton.

PR #1213, opened 2026-07-12
Steam bridge: paravirtualized Steamworks for running licensed games in the sandbox

momo5502:
## What

A paravirtualization bridge that lets a guest Windows game running inside Sogen talk to the **real,
genuine Steam client on the host**. The game's own unmodified `steam_api(64).dll` loads our
`steamclient` shim, which forwards Steamworks interface calls across the emulator boundary to the
host's real `steamclient64.dll` and the running Steam client. It is the Steam analogue of the GPU
bridge.

The design mirrors Proton's `lsteamclient`, but the host is Windows, so the ABI matches on both sides
and the compiler — not hand-written marshalling — handles calling conventions. Generated
proxies/thunks derive from the real `ISteam*` interfaces; 32- and 64-bit guests both work.

## Why

The goal is to run **genuinely-licensed games safely inside the sandbox** — many games refuse to
start without Steam, so bridging to the real client is what makes them runnable under Sogen. This is
**interoperability, not emulation**: it forwards to the genuine Steam client on the user's genuine
account. It does not reimplement, emulate, or circumvent anything in Steam.

## How it works

- **Guest shim** (`src/samples/steam-shim/`) — a `steamclient(64).dll` replacement (the game's own
  versioned `steam_api` is never touched); exports `CreateInterface` + the `Steam_*` callback pump.
- **`\\.\SogenSteam` io_device** (`src/windows-emulator/devices/steam_bridge.cpp`) — hardened; treats
  all guest input as hostile (bounds-checks every length/offset, caps allocations).
- **`steam-host-backend/`** — a pure-C ABI over the host `steamclient64` plus generated thunks, so the
  emulator device never sees `<windows.h>`/SDK types.
- **Generator** (`src/tools/steam-bridge-generator/`) — parses the SDK headers to emit version-exact
  guest proxies + host thunks; both `#include` the real headers so the compiler emits correct ABI.
- A synthetic "Steam is running" process (registry seed + `NtOpenProcess`/liveness) so a game's
  `steam_api` reaches the genuine, already-installed client through the bridge.

See `docs/steam-bridge.md` and `docs/steam-bridge-versioning.md`.

## Licensing posture

- **No Valve material is redistributed.** No SDK headers, no `steam_api.json` (only a hand-written
  MIT-safe mock), no generated/derived code (gitignored), no Valve binaries.
- The Steamworks SDK headers are obtained at the **user's build time** via `FetchContent` from Valve's
  own `ValveSoftware/Proton` repository. The project ships only a URL and a code generator; any derived
  marshalling code exists solely in the user's build tree.
- Gated behind `SOGEN_ENABLE_STEAM`.

## Status

**Draft.** Proven in-emulator against real Steam (32- and 64-bit guests): interface calls return the
real account (persona, SteamID, friends), plain callbacks, and async call-results. Reverse-callback
(server-browser) delivery and full retail-game boot are still in progress.


src/steam-host-backend/steam_host_runtime.hpp
```diff
@@ -0,0 +1,185 @@
...
+                ++p;
+            }
+            if (p < end)
+            {
+                ++p; // consume the NUL
+            }
+            else
+            {
+                static const char empty = '\0';
+                return &empty; // unterminated: hand back a safe empty string
```
momo5502: why overengineer this?

```suggestion
                return "";
```

momo5502:
@claude review

src/samples/steam-shim-test/CMakeLists.txt
momo5502, line 1: i think we can completely remove the steam-shim-test.
it won't run in CI anyway

src/samples/steam-shim/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+  return()
+endif()
+
+# Version-agnostic base TUs: the flat Steamworks entry points + steamclient boundary (steam_shim.cpp, SDK-
+# free so its exported SteamAPI_* don't clash with the SDK's inline ones) and the reverse-callback dispatch
+# (steam_shim_reverse.cpp, built against the newest snapshot for the stable response interface types).
+set(_base_srcs
+  "${CMAKE_CURRENT_SOURCE_DIR}/steam_shim.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/steam_shim_reverse.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/steam_shim_runtime.hpp")
```
momo5502: why not glob like every other target does it in CMake?

.github/workflows/ci-reusable.yml
```diff
@@ -140,34 +140,47 @@ jobs:
...
+
       - name: Enable Developer Command Prompt (x86)
         uses: ilammy/msvc-dev-cmd@v1.13.0
         with:
           arch: x86
 
-      - name: Build 32-bit Vulkan Shim
-        run: cmake --preset=release && cmake --build --preset=release -t vulkan-shim
+      - name: Build 32-bit Vulkan and Steam Shims
+        run: cmake --preset=release && cmake --build --preset=release -t vulkan-shim steam-shim
```
momo5502: i think it might make sense to have a meta target in cmake where we just call add_dependency(<metatarget> vulkan-shim) etc. so that we only have to specify the meta target here. obv the meta-target needs a fitting name.

momo5502:
This fixes #1170

src/samples/steam-shim/steam_shim_runtime.hpp
momo5502, line 1: the comments in this file are way too verbose. also go through other comments in this PR. we don't need entire books as comments. only information that goes beyond what the code already describes

src/steam-bridge-protocol/steam_bridge_protocol.hpp
momo5502, line 1: same here. tons of useless comments...

src/tools/steam-bridge-generator/steam_api.mock.json
momo5502, line 1: no need for this. remove this.

src/windows-emulator/devices/afd_endpoint.cpp
momo5502, line 1: the changes here should be part of a separate PR. it has nothing to do with steam

src/CMakeLists.txt
```diff
@@ -1,5 +1,80 @@
...
+        message(WARNING "Steam bridge generator failed (is the 'libclang' pip package installed?):\n${_gen_err}\n"
+                        "Steam bridge disabled.")
+      else()
+        message(STATUS "Steam bridge enabled (base ${SOGEN_STEAM_BASE_DIR}).")
+        set(SOGEN_STEAM_ENABLED TRUE)
+      endif()
+    endif()
+  endif()
+endif()
+
```
momo5502: can this be extracted into an isolated CMake file/script?

.github/workflows/ci-reusable.yml
```diff
@@ -140,34 +140,47 @@ jobs:
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      # libclang lets the Steam-bridge generator run at configure time, so the guest steamclient shims are
+      # actually built here rather than silently skipped. The root CMakeLists resolves the PATH interpreter
+      # (this one), so it and pip agree.
+      - name: Install libclang (Steam bridge codegen)
+        run: python -m pip install --disable-pip-version-check libclang
```
momo5502, line 152: i think we need this for all jobs that build the project

momo5502:
@claude review

momo5502:
@claude review

PR #1215, opened 2026-07-12
afd: answer NAT-PMP locally for online play

momo5502:
Games (via miniupnpc) send NAT-PMP (RFC 6886) requests to a gateway's UDP port 5351 to open a port before online play. The emulator has no real NAT-PMP gateway, so this synthesizes a well-formed response (public-address and UDP/TCP port-map opcodes) and hands it back on the guest's next recvfrom, so online play can proceed. Also adds env-gated recv/recv_datagram trace logging.

Split out of the Steam bridge PR (#1213) — this is network/NAT plumbing, unrelated to Steam.

### DIFF
--- a/src/windows-emulator/devices/afd_endpoint.cpp
+++ b/src/windows-emulator/devices/afd_endpoint.cpp
@@ -8,6 +8,11 @@
 #include <network/address.hpp>
 #include <network/socket.hpp>
 
+#include <cstdio>
+#include <cstdlib>
+#include <deque>
+#include <optional>
+
 #include <utils/finally.hpp>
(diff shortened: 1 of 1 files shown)


src/windows-emulator/devices/afd_endpoint.cpp
```diff
@@ -915,6 +1003,12 @@ namespace sogen
 
                 emu.read_memory(wsabuf.buf, host_buffer.data(), host_buffer.size());
 
+                if (std::getenv("SOGEN_AFD_TRACE"))
```
momo5502, line 1006: please remove all the tracing from this PR.

PR #1222, opened 2026-07-13
steam-bridge: compile the generated code with warnings-as-errors

momo5502:
## What

The Steam per-tag object libraries compile *only* our generated marshalling
thunks/proxies (plus the runtime headers), but they opted out of the project's
warnings-as-errors via `disable_warnings`. So warnings in the generated code went
unnoticed, and the GCC build spewed noise from the bundled SDK headers.

This holds the generated code to the same standard as the rest of `src/`:

- The Valve SDK headers are marked **SYSTEM** on the per-tag targets (generated
  headers stay on the normal include path). That suppresses the SDK's own warnings
  and the cross-snapshot `-Wodr` noise under LTO, while our generated code still warns.
- The per-tag libs drop the `disable_warnings` opt-out and inherit `-Werror`. On MSVC
  the external-header warning level is lowered (`/external:W0`) and
  `_CRT_SECURE_NO_WARNINGS` is defined. Hand-written base TUs are unchanged.

Enabling this surfaced (and this PR fixes) two generator issues in the guest proxies:
stub bodies for non-marshallable methods, and `null_ptr` params (double-pointer filter
lists the host substitutes with `nullptr`), both named their parameters without using
them → `-Wunused-parameter`. They're now cast to void.

Also fixes a real `-Wtautological-compare` in `steam_host_runtime.hpp` (the reader's
`p + n >= p` overflow guard is always true since pointer overflow is UB; replaced with
the equivalent `n <= end - p`), and makes `sogen_target_disable_warnings` actually
silence warnings on GCC/Clang (`-w`), which it previously only did on MSVC.

## Why nothing errored before

`sogen_targets_set_warnings_as_errors` skips any target with the
`SOGEN_WARNINGS_DISABLE` property, and every Steam target set it. So the generated
code (and the SDK headers) compiled without `-Werror`; the `-Wodr` lines are LTO
link-time diagnostics from linking many SDK-version object libs into one binary.

## Validation

Generated v164 host + guest TUs compiled with `-Wall -Wextra -pedantic -Werror`
and the SDK as `-isystem` — clean at 64-bit (host + guest) and 32-bit (guest).
Full matrix (GCC, MSVC, all ~90 snapshots) validated by CI.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -196,10 +196,12 @@ function(sogen_target_disable_warnings target)
   sogen_target_remove_compile_options(${target} /W3 -W3 /W4 -W4)
 
   if(MSVC)
-  set(compile_options
-    /W0
-    /D_CRT_SECURE_NO_WARNINGS=1
-  )
+    set(compile_options
+      /W0
+      /D_CRT_SECURE_NO_WARNINGS=1
+    )
+  else()
+    set(compile_options -w) # GCC/Clang: actually silence warnings for third-party / generated code
   endif()
 
   sogen_target_c_and_cxx_compile_options(${target} PRIVATE ${compile_options})
@@ -213,6 +215,17 @@ function(sogen_targets_disable_warnings)
   foreach(target ${ARGV})
     sogen_target_disable_warnings(${target})
   endforeach()
+endfunction()
+
+##########################################
+
+# MSVC extras for a generated-Steam-code target whose SDK headers the caller marks SYSTEM. C4743/C4744 are
+# MSVC's cross-TU -Wodr analogue (ISteam* types differ across snapshots) and fire at LTCG link.
+function(sogen_target_generated_code_warnings target)
+  if(MSVC)
+    target_compile_options(${target} PRIVATE /external:W0 /wd4743 /wd4744)
+    target_compile_definitions(${target} PRIVATE _CRT_SECURE_NO_WARNINGS=1)
+  endif()
 endfunction()
 
 ##########################################
--- a/src/samples/steam-shim/CMakeLists.txt
+++ b/src/samples/steam-shim/CMakeLists.txt
@@ -41,10 +41,11 @@ foreach(_entry ${SOGEN_STEAM_TAG_LIST})
   configure_file("${CMAKE_CURRENT_SOURCE_DIR}/steam_shim_tag.cpp.in"
                  "${CMAKE_CURRENT_BINARY_DIR}/steam_shim_${_tag}.gen.cpp" @ONLY)
   add_library(steam-shim-${_tag} OBJECT EXCLUDE_FROM_ALL "${CMAKE_CURRENT_BINARY_DIR}/steam_shim_${_tag}.gen.cpp")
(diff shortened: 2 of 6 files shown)


src/samples/steam-shim/CMakeLists.txt
```diff
@@ -41,10 +41,12 @@ foreach(_entry ${SOGEN_STEAM_TAG_LIST})
   configure_file("${CMAKE_CURRENT_SOURCE_DIR}/steam_shim_tag.cpp.in"
                  "${CMAKE_CURRENT_BINARY_DIR}/steam_shim_${_tag}.gen.cpp" @ONLY)
   add_library(steam-shim-${_tag} OBJECT EXCLUDE_FROM_ALL "${CMAKE_CURRENT_BINARY_DIR}/steam_shim_${_tag}.gen.cpp")
-  sogen_target_disable_warnings(steam-shim-${_tag})
+  # Warnings-as-errors stays ON here (our generated proxies); only the Valve SDK headers are marked SYSTEM.
```
momo5502: that comment is useless

src/steam-host-backend/CMakeLists.txt
```diff
@@ -26,9 +26,13 @@ foreach(_entry ${SOGEN_STEAM_TAG_LIST})
   configure_file("${CMAKE_CURRENT_LIST_DIR}/steam_host_tag.cpp.in"
                  "${CMAKE_CURRENT_BINARY_DIR}/steam_host_${_tag}.gen.cpp" @ONLY)
   add_library(steam-host-${_tag} OBJECT "${CMAKE_CURRENT_BINARY_DIR}/steam_host_${_tag}.gen.cpp")
-  sogen_target_disable_warnings(steam-host-${_tag})
+  # Warnings-as-errors stays ON here (inherited from OWN_TARGETS): this compiles our generated thunks, which
+  # we want held to standard. Only the Valve SDK headers are marked SYSTEM, so their noise -- and the
+  # cross-snapshot -Wodr under LTO -- is suppressed while the generated code is not.
```
momo5502: useless comment

momo5502:
@claude review

cmake/utils.cmake
```diff
@@ -217,6 +219,21 @@ endfunction()
...
+# headers (which the caller marks SYSTEM). On MSVC, SYSTEM includes still warn unless the external warning
+# level is lowered, and the SDK/runtime touch CRT functions MSVC flags as deprecated.
+function(sogen_target_generated_code_warnings target)
+  if(MSVC)
+    # /external:W0: don't warn on the SYSTEM (SDK) headers. /wd4743 /wd4744: the MSVC analogue of GCC's
+    # -Wodr -- identically-named ISteam* types legitimately differ across SDK snapshots, and MSVC flags the
+    # vtable/data size mismatch at LTCG link time; suppress it like we do -Wodr (via SYSTEM headers) on GCC.
+    target_compile_options(${target} PRIVATE /external:W0 /wd4743 /wd4744)
+    target_compile_definitions(${target} PRIVATE _CRT_SECURE_NO_WARNINGS=1)
+  endif()
```
momo5502: very noisy comments

PR #1223, opened 2026-07-14
security: hardening pass for the Steam bridge, Vulkan host, and AFD

momo5502:
Hardening pass following a security audit of the guest→host boundaries (Steam
bridge, Vulkan host / GPU bridge, and the AFD socket device). The guest is
untrusted: every guest-controlled length/index/handle that reaches a host
allocation, memory operation, or real API is now bounded and must not disclose a
host pointer. No arbitrary-R/W was found; these fixes close DoS, info-leak, and
hardening-gap findings.

## Steam bridge
- **Opaque-handle indirection table** (info leak / ASLR defeat → fixed). `HServerListRequest`
  was returned to the guest as the **raw steamclient pointer** (and echoed in reverse
  callbacks). `register_opaque_handle` now hands out a dense index backed by an
  `index ↔ real-handle` table; `resolve_opaque_handle` maps a guest index back to the
  real handle, returning 0 for anything never issued. The guest never sees a host
  pointer, and a forged index can't become an arbitrary pointer.
- **Pointer-struct guard extended to outputs/returns.** It previously only covered input
  structs. A pointer member in an out-param or by-value struct return would disclose a
  host pointer. This flags `UpdatePublishedFileTags` / `Enumerate{Published,UserShared}WorkshopFiles`
  (they take `SteamParamStringArray_t`, were mis-marshalled as zeroed out-structs that
  silently dropped the guest's tags) — now blocked.
- **Invoke allocation caps** — arg blob and reply staging capped at 64 MiB so a
  guest-declared ~4 GiB IOCTL buffer can't force a multi-gigabyte host allocation.
- `struct_return` zero-init (padding-leak mitigation), `announced_connected` bound, and
  a `reverse_bytes` underflow guard.

## Vulkan host / GPU bridge
- **`create_swapchain`** rejects an absurd `min_image_count` (>16) before the per-image
  `create_image` + `allocate_memory` loop — a deterministic host-resource DoS
  (`create_device` already caps `queue_count` this way).
- **`allocate_memory`** rejects an out-of-range `memory_type_index` (drivers index internal
  arrays with it unchecked → OOB in the host driver) and **zeroes freshly allocated
  host-visible memory**, which `map_memory`/`download_memory` alias/copy to the guest
  (uninitialized-host-memory leak).
- **`get_query_pool_results`** clamps the host-returned `written` to the staging buffer
  before the guest write (defense in depth).

## AFD
- Stream `recv`/`send` cap the staging allocation at 64 MiB; a guest can declare a
  ~4 GiB `WSABUF` without backing it (partial-transfer semantics keep this correct).

## Validation
- Generator regenerated against SDK v164; host thunks + guest proxies compile clean
  under `-Wall -Wextra -pedantic -Werror -Wshorten-64-to-32`. The opaque-handle
  indirection and blocked methods emit as expected.
- clang-format clean; full build/matrix validated by CI.

## Notes / residual
- `struct_return` padding: only `InputAnalogActionData_t` (`GetAnalogActionData`) has
  trailing padding; the zero-init is best-effort — a trivially-copyable return can still
  copy source padding. Not generically eliminable without per-struct field marshalling.
- The opaque-handle table and matchmaking proxy maps are not pruned on release (a slow,
  bounded growth); left for a follow-up.
- AFD unrestricted egress (loopback / link-local) is the emulator's intentional
  real-networking model and is **not** changed here.

### DIFF
--- a/src/windows-emulator/devices/vulkan_host.cpp
+++ b/src/windows-emulator/devices/vulkan_host.cpp
@@ -647,6 +647,18 @@ namespace sogen
             return index;
         }
 
+        // Fills `props` from the device's physical device. Returns false if the query is unavailable.
+        bool query_memory_properties(const device_data& dev, VkPhysicalDeviceMemoryProperties& props)
+        {
+            const auto instance = this->instances.find(dev.instance_id);
+            if (instance == this->instances.end() || !instance->second.get_physical_device_memory_properties)
+            {
+                return false;
+            }
+            instance->second.get_physical_device_memory_properties(dev.physical_device, &props);
+            return true;
+        }
+
         template <typename Map, typename Pred>
         void erase_owned(Map& map, Pred pred)
         {
@@ -2589,6 +2601,15 @@ namespace sogen
             return VK_ERROR_INITIALIZATION_FAILED;
         }
 
+        VkPhysicalDeviceMemoryProperties mem_props{};
+        const bool have_props = this->impl_->query_memory_properties(dev->second, mem_props);
+        // The spec requires memoryTypeIndex < memoryTypeCount; drivers index their internal type/heap arrays
+        // with it unchecked, so an out-of-range guest value is an OOB read inside the host driver.
+        if (have_props && memory_type_index >= mem_props.memoryTypeCount)
+        {
+            return VK_ERROR_INITIALIZATION_FAILED;
+        }
+
         // Round the allocation up to a whole number of guest pages. The map-direct path aliases the host
         // mapping straight into the guest, which can only be done at page granularity; rounding the actual
         // allocation up means the page-aligned alias never covers anything beyond this allocation, so the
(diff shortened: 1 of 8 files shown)


src/windows-emulator/devices/vulkan_host.cpp
```diff
@@ -3483,6 +3519,13 @@ namespace sogen
             return VK_ERROR_INITIALIZATION_FAILED;
         }
 
+        // A real swapchain has only a handful of images; reject an absurd guest count before it drives the
+        // per-image allocation loop below (each iteration does create_image + allocate_memory).
+        constexpr uint32_t max_swapchain_images = 16;
```
momo5502: is 16 a reasonable number here? i think we could safely go a bit bigger, e.g. 128

momo5502:
@claude review

src/windows-emulator/devices/vulkan_host.cpp
momo5502: Good call — bumped to 128. It's only a sanity ceiling to stop the billions-of-allocations DoS, so a higher value costs nothing.

momo5502:
Addressed the non-coherent memory point in the zero-fill: after the memset, `allocate_memory` now flushes the range with `vkFlushMappedMemoryRanges` when the memory type lacks `HOST_COHERENT` (gated so coherent memory — the common case after `substitute_cached_memory_type` steers host-visible allocations to a coherent type — skips the extra call). Skipping the zero-fill on non-coherent memory would have left the leak, so flushing is the right fix. Good catch.

momo5502:
@claude review

PR #1227, opened 2026-07-14
Run Unity/D3D11 games: D3DKMT Render, Vulkan extension filtering, Steam interface versions

momo5502:
Getting Paleo Pines (Unity 2022, D3D11, Steamworks.NET) to run surfaced four independent bugs.

## Unimplemented syscalls (aborted the emulation)

- **`NtGdiDdDDIRender`** — the D3D11 runtime submits a command buffer as soon as it renders, so any title that reaches the real D3DKMT path died here. The context's command buffer, allocation list and patch-location list now live in `dxgk_state`, so `CreateContext` and `Render` hand them back and grow them on request. Struct layout verified against the SDK's `D3DKMT_RENDER`.
- **`NtUserGetWindowCompositionAttribute`** — `DwmGetWindowAttribute` funnels into it. There is no DWM in the emulated session, so windows are never composed, never cloaked, and the extended frame equals the window rect.

## GPU bridge advertised extensions it cannot marshal

The bridge passed the host driver's full device extension list to the guest, including the Win32 interop extensions. DXVK saw `VK_KHR_external_memory_win32`, enabled it, created a shared texture, and jumped to a null `vkGetMemoryWin32HandleKHR` (RIP=0). HANDLE-based interop cannot work across the bridge anyway — guest handles are emulator objects — so those extensions are now filtered out of the enumeration. The shim additionally implements `vkGetMemoryWin32HandleKHR` as a clean failure, because DXVK asks for the export even after its own capability check has already failed.

## Steam interface versions were silently empty

The bridge generator matched an interface class to its version string by name similarity, which yields an empty version for every interface Valve did not name after its class:

| Interface | Version string |
|---|---|
| `ISteamGameSearch` | `SteamMatchGameSearch001` |
| `ISteamInventory` | `STEAMINVENTORY_INTERFACE_V003` |
| `ISteamVideo` | `STEAMVIDEO_INTERFACE_V002` |

The version string is the key the guest proxy factory looks up, so those three could not be proxied. `steam_api64` walks the `ISteamClient` vtable during `SteamAPI_Init`, got a null back from `GetISteamGameSearch`, and aborted — no error, no further calls. Every game using the interface context failed to init Steam.

`STEAM_DEFINE_*_INTERFACE_ACCESSOR` states the class ↔ version-macro pairing authoritatively; the generator now reads it and keeps the name heuristic only as a fallback for snapshots predating the accessors.

## Testing

- `analyzer -s test-sample.exe` passes.
- Paleo Pines boots, DXVK comes up on the host GPU, Unity initializes URP, DXVK creates a 1920x1080 swapchain, `SteamAPI_Init` succeeds, and the game runs.

### DIFF
--- a/src/emulator-platform/platform/gdi.hpp
+++ b/src/emulator-platform/platform/gdi.hpp
@@ -382,6 +382,31 @@ namespace sogen
         UINT64 CommandBuffer;
     };
 
+    struct EMU_D3DKMT_RENDER
+    {
+        UINT32 hContext;
+        UINT32 CommandOffset;
+        UINT32 CommandLength;
+        UINT32 AllocationCount;
+        UINT32 PatchLocationCount;
+        UINT64 pNewCommandBuffer;
+        UINT32 NewCommandBufferSize;
+        UINT64 pNewAllocationList;
+        UINT32 NewAllocationListSize;
+        UINT64 pNewPatchLocationList;
+        UINT32 NewPatchLocationListSize;
+        UINT32 Flags;
+        UINT64 PresentHistoryToken;
+        UINT32 BroadcastContextCount;
+        UINT32 BroadcastContext[64];
+        UINT32 QueuedBufferCount;
+        UINT64 NewCommandBuffer;
+        UINT64 pPrivateDriverData;
+        UINT32 PrivateDriverDataSize;
+    };
+
+    static_assert(sizeof(EMU_D3DKMT_RENDER) == 368);
+
     struct EMU_D3DDDI_ALLOCATIONINFO
     {
         UINT32 hAllocation;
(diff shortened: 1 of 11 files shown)


momo5502:
@claude review

momo5502:
@claude review

momo5502:
Addressed the review points in b63550c0:

- **#1 (NtGdiDdDDIRender unbounded/overflow alloc)** — the real one. `NewCommandBufferSize`/`NewAllocationListSize`/`NewPatchLocationListSize` are now clamped before sizing the host allocation (command buffer ≤ 64 MiB, list counts ≤ 64k entries), which also keeps `count * entry_size` from overflowing uint32. Verified against the `reserve_buffer` path (it builds a host `std::vector<uint8_t>` of the size, so this was a direct multi-GB host-OOM per call).
- **#2 (WoW64 composition attribute)** — `NtUserGetWindowCompositionAttribute` now reads the 12-byte 32-bit layout for `is_wow64_process` guests instead of the native 24-byte one.
- **#3 (create_device extensions)** — `create_device` now strips `unsupported_device_extensions` from the guest-supplied enabled list, matching what enumeration hides, rather than trusting the guest.

Built `windows-emulator` locally (clean) and each file passes a full semantic check.

PR #1229, opened 2026-07-15
steam: make bridge codegen incremental (skip regen + recompile when unchanged)

momo5502:
## Problem

The Steam bridge marshalling code was **regenerated on every configure** and **recompiled on every build**, even when nothing had changed. Two independent causes:

1. `generate.py` rewrote every output header unconditionally (`open(..., "w")`), bumping their mtimes — so every TU that `#include`s them recompiled even when the content was byte-for-byte identical.
2. The codegen runs at CMake configure time via `execute_process`, and libclang parses **all 90 SDK snapshots** — by far the slowest part of configure — every single reconfigure.

## Fix

**`generate.py` — write-if-different.** All output writes now go through a `write_if_changed()` helper that leaves byte-identical files untouched. It reads with `newline=""` so CRLF translation on Windows doesn't cause spurious rewrites. Even if the generator runs, unchanged headers keep their mtime and downstream compilation is skipped.

**`steam-bridge.cmake` — input stamp.** The output is fully determined by the generator script, the target bit width, and the SDK snapshot set (which only changes when Proton headers are re-fetched for a new ref). We now stamp those into `steam-generated/.inputs.stamp` (SHA256 of `generate.py` + bits + Proton repo/ref + resolved tag list) and skip the Python run when the stamp is unchanged since the last successful generation. The stamp re-triggers exactly when it should: editing `generate.py` or updating Proton.

## Verification

- Running the generator twice against the fetched snapshots: all output mtimes unchanged on the second run; a genuine content change still rewrites.
- Standalone CMake test of the stamp logic: fresh → generate, unchanged → skip, generator edited → generate, Proton ref changed → generate.
- Full project reconfigure twice:
  - Configure #1: `Steam bridge: generating 64-bit marshalling for 90 SDK snapshot(s)...` → `enabled`
  - Configure #2: `Steam bridge: generated marshalling up to date for 90 snapshot(s); skipping codegen.`

Net effect: after the first generation, plain rebuilds do no Steam regeneration and no Steam recompilation; reconfigures skip the Python run; and both correctly re-trigger when the generator or the SDK snapshots actually change.

### DIFF
--- a/cmake/steam-bridge.cmake
+++ b/cmake/steam-bridge.cmake
@@ -56,18 +56,42 @@ if(SOGEN_ENABLE_STEAM AND SOGEN_STEAM_HOST_PLATFORM)
     else()
       math(EXPR _bits "${CMAKE_SIZEOF_VOID_P} * 8")
       set(SOGEN_STEAM_GEN_DIR "${CMAKE_BINARY_DIR}/steam-generated")
+      set(_gen_py "${CMAKE_CURRENT_SOURCE_DIR}/tools/steam-bridge-generator/generate.py")
       list(LENGTH SOGEN_STEAM_TAG_LIST _tag_count)
-      message(STATUS "Steam bridge: generating ${_bits}-bit marshalling for ${_tag_count} SDK snapshot(s)...")
-      execute_process(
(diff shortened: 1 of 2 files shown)


momo5502:
CI built, GitHub just had a hickup... out of disk space, lul, on clang-tidy. Clang-tidy doesn't build steam though, so it's unrelated.

Issue #1001, opened 2026-05-30
Loading from minidump doesn't work


po0p:
https://github.com/momo5502/sogen/blob/223ce0e4f4c394833726e95c2d49825070a6557f/src/windows-emulator/module/module_manager.cpp#L536-L559

detect_from_memory memcpy's from host memory (when we definetely want to read from the dump content / this->memory_)

small hack just to make it work 

```cpp
const auto header_size = static_cast<size_t>(std::min<uint64_t>(image_size, 0x1000));
        std::vector<std::byte> header(header_size);
        if (!this->memory_->try_read_memory(base_address, header.data(), header.size()))
        {
            logger.error("Failed to read PE header for module at 0x%" PRIx64 "\n", base_address);
            return nullptr;
        }

        auto detection_result = pe_architecture_detector::detect_from_memory(
            reinterpret_cast<uint64_t>(header.data()), header.size());
```

cc #333

momo5502:
Feel free to create a pull request with your fix. At the moment, I don't have the time to investigate and also don't actively use the minidump support.

PR #1233, opened 2026-07-15
emulator: read PE headers from minidump memory

ukint-vs:
## Summary

Read PE header bytes from dump-backed module memory before detecting the PE architecture. This fixes module loading from minidumps, including sparse or unreadable header cases, using fixed-size header reads bounded by the image size.

## Related issue

https://github.com/momo5502/sogen/issues/1001

### DIFF
--- a/src/emulator-platform/platform/win_pefile.hpp
+++ b/src/emulator-platform/platform/win_pefile.hpp
@@ -10,7 +10,6 @@
 #include <array>
 #include <limits>
 #include "common/utils/buffer_accessor.hpp"
-#include <emulator/memory_interface.hpp>
 
 #include "primitives.hpp"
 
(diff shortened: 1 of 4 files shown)


momo5502:
Thanks for the contribution 🥳

To fix the formatting issues, you need clang-format from LLVM 21.
I think the other CI issues should be clear, if not, feel free to ask for clarification :)

momo5502:
@claude review

src/emulator-platform/platform/win_pefile.hpp
```diff
@@ -8,7 +8,9 @@
 #include <system_error>
 #include <variant>
 #include <array>
+#include <limits>
 #include "common/utils/buffer_accessor.hpp"
+#include <emulator/memory_interface.hpp>
```
momo5502: this is not good. the emulator-platform does not have the emulator as dependency. this is an architectural violation.

i think instead the function should accept a functor that it can call to resolve memory

momo5502 (approved):

Issue #1110, opened 2026-06-28
Support side-by-side (SxS) activation contexts / manifests


momo5502:
## Problem

Apps that load the **Visual C++ 2005/2008 CRT** (`msvcr80.dll` / `msvcr90.dll`) **app-local** fail during process initialization with `STATUS_APP_INIT_FAILURE` (`0xC0000145`), raised via `NtRaiseHardError`.

Repro: **Unreal Tournament III** (`UT3.exe`, 32-bit / WoW64). It loads its full launcher stack (wxWidgets + VC8 CRT) and then the loader aborts.

## Root cause

`msvcr80.dll` is a side-by-side (SxS) assembly. Its `DllMain` (`DLL_PROCESS_ATTACH`) runs the well-known **R6034 anti-DLL-hijacking check**: if the CRT is *not* loaded from the system directory, it must be resolvable through an **active activation context**, verified via `FindActCtxSectionStringW(0, 0, 2, L"msvcr80.dll", ...)`.

Sogen has no SxS / activation-context support, so the in-guest `ntdll!RtlFindActivationContextSectionString` (`kernel32!FindActCtxSectionStringW`) returns failure. The CRT concludes it was loaded incorrectly → its `DllMain` returns `FALSE` → the loader aborts the process.

Confirmed by static analysis (`msvcr80!DllMain` → check function): the check's result is what gates `DllMain`'s return value:

```c
result = crt_location_check(hModule);   // calls FindActCtxSectionStringW
if (result != 0) { ... return TRUE; }   // PASS
crt_report_error(0x22);                  // R6034 path
... return FALSE;                        // APP_INIT_FAILURE
```

On real Windows, the app's embedded `Microsoft.VC80.CRT` manifest creates an activation context, so the lookup succeeds and the check passes.

## Impact

Blocks the whole class of **VS2005/2008-CRT apps loaded app-local** (common for games and older applications). Modern UCRT-based apps are unaffected (no such check), which is why `ucrtbase`/`msvcrt` init fine while `msvcr80` does not.

## Scope of a real fix

Implement activation-context / SxS manifest support: the loader parsing the application's and DLLs' manifests and building the activation-context section data that the in-guest `ntdll` queries (`RtlFindActivationContextSectionString*`, `RtlQueryActivationContext*`, etc.).

## Workaround (verified)

[...]

PR #1234, opened 2026-07-17
FEX backend port to macOS/Apple Silicon, WoW64 bring-up

JackTYM:
## Summary

Brings the FEXCore (x86-64/x86 → ARM64 dynamic binary translation) backend up on
macOS/Apple Silicon, adds 32-bit WoW64 process support, and validates the result
against real commercial software (a full WoW64 game reaching sustained real-time
rendering) plus ~40 existing sample executables.

Builds directly on this branch's existing "backend builds and links on ARM64 Linux
CI" state — this PR is what makes it actually *run*, on macOS, including 32-bit
guests.

## What's in this PR

**Backend port (macOS/Apple Silicon)**

- Real `HostFeatures` detection via `sysctlbyname` (was previously stubbed).
- Fault delivery via plain POSIX `sigaction` — no Mach exception port needed.
  Handles JIT guard-page races, MMIO, misaligned load-acquire/store-release,
  synthetic `#GP`-to-vector remapping, and cooperative thread-stop via FEXCore's
  `InterruptFaultPage` mechanism.
- A per-guest-4KB-page permission shadow table reconciling Apple Silicon's 16KB
  host page granularity against the guest's 4KB architectural page size.
- Real JIT code-buffer overflow guard pages. Apple Silicon's `MAP_JIT` memory
  can't have its protection changed by an ordinary `mprotect()` after mapping
  (confirmed empirically: always fails `EACCES`), so FEXCore's own end-of-buffer
  guard page silently never took effect. The sogen-side FEXCore-internal
  allocator arena now places the real guard by construction instead: it maps
  only the buffer's leading portion as the executable `MAP_JIT` region and
  leaves the trailing host page as part of the arena's permanent `PROT_NONE`
  reservation, which genuinely faults on any access (see the accompanying
  `deps/FEX` submodule bump for the matching guard-size fix in FEXCore's own
  temp-buffer allocator).
- GPU-bridge host-memory coherency (`mach_vm_remap` aliasing, `sys_dcache_flush`)
  for the paravirtualized-GPU memory-sharing path.

**WoW64 (32-bit) support**

FEXCore is fixed-bitness per `Context`; a WoW64 process needs both a 64-bit
Context (native ntdll/wow64*.dll code) and a 32-bit one (the guest image and its
32-bit ntdll), with state marshaled and control switched at the real x86 "heaven's
gate" bitness-switch trampoline. This PR implements:

- A second, lazily-created 32-bit FEXCore Context/Thread per process, with its
  own GDT and call-ret shadow stack.
- Gate-crossing interception (both directions) via a generic non-executable-range
  mechanism, marshaling GPRs/XMM/x87/EFLAGS/segments between the two Contexts
  while preserving each engine's own JIT-internal bookkeeping.
- A guest-address rebase for 32-bit memory (Apple Silicon's mandatory 4GB
  `__PAGEZERO` makes the low 4GB permanently unmapped, conflicting with this
  project's guest-VA==host-VA model) — implemented as a FEXCore-side
  runtime-conditional IR `Select` (see the accompanying `deps/FEX` submodule
  bump), not a sogen-side workaround.
- WoW64-specific memory-manager fixes: fixed-address allocations and the
  auto-placement path both now rescan live host VM state before deciding an
  address is free, closing intermittent collisions with host-owned memory.
- A stale-JIT-code-cache fix for a WoW64 process unmapping a DLL while the
  *other* Context is active.

**Shared (backend-agnostic) fixes found via broad validation**

Running ~40 pre-existing sample executables under both this backend and Unicorn
surfaced several bugs affecting *both* backends identically:

- user32's client-side `HMValidateHandle` (aheList slot lookup) expected
  `HWND & 0xFFFF` to equal the handle's storage slot; a recent handle-4-alignment
  change had shifted the id without updating aheList's own write/free slot.
- `NtUserQueryDisplayConfig` was implemented against the wrong (6-arg, public-API)
  signature instead of the true 5-arg kernel syscall ABI, corrupting a caller's
  stack frame.
- A GDI sub-4GB allocation guard, correct for 32-bit/WoW64 GDI pointers, was also
  applied to ordinary 64-bit processes — always tripping under this backend.
- `NtUserCallMsgFilter`, `NtGdiDdDDISubmitCommand`, `NtGdiDdDDIPresent`,
  `NtCreateRegistryTransaction`, `NtOpenRegistryTransaction` implemented
  (previously unimplemented on every backend).

## Testing performed

- Full regression suite (`test-sample.exe`, `hello.exe`, `busybox.exe`) on both
  this backend and Unicorn, matching byte-for-byte behavior throughout.
- ~40 pre-existing sample executables (process/thread info, sync primitives,
  sections, security, registry, pipes, sockets, timers, I/O, GUI dialogs,
  audio/multimedia, 32-bit WoW64 binaries, DXGK harnesses) run under both
  backends; every remaining discrepancy is a shared, backend-agnostic gap (an
  unimplemented syscall or a missing staged DLL export), not a difference
  between this backend and the existing one.
- A real, unmodified commercial WoW64 game (a 32-bit Call-of-Duty-engine title)
  reaches its main menu and sustains real-time rendering, confirmed across
  multiple independent runs on this branch's final state.
- GPU paravirtualization independently confirmed via a minimal, dedicated Vulkan
  guest test (instance/device creation, buffer fill+readback, image
  clear+readback) exercising the guest-shim → IOCTL bridge → host-Vulkan path
  more directly than the game's D3D9-over-DXVK route — 24/25 clean runs,
  byte-identical output to Unicorn.
- Full CI matrix (dedicated FEX build, FEX smoke test, FEX GPU paravirtualization
  smoke test, and a WoW64-under-FEX smoke test exercising genuine 32-on-64
  translation) green on the final, squashed, rebased-onto-`main` state of this
  branch.

## Performance

A pure compute-bound benchmark (2 billion iterations of an ALU-only loop, no
syscalls beyond start/end timestamps) shows this backend at roughly **129x less
wall-clock time** than the Unicorn backend on the same hardware (~3.4s vs ~439s)
— the expected result of JIT-compiled native ARM64 execution vs.
interpretation/binary-translation with per-block hook overhead. An earlier, mixed
I/O smoke-test benchmark (syscall-heavy, not compute-bound) showed a smaller but
still large ~27-30x gap.

## Known quirks and limitations

- One sample (a calc.exe-style app) fails identically on both backends due to a
  staged `comctl32.dll` export-table mismatch — a staged-asset version issue, not
  a code bug.
- The one visible rendering artifact on the confirmed-working game (menu text
  rendering as blank boxes) was traced byte-for-byte from disk through
  decompression through the GPU-bridge IOCTL to the exact bytes handed to the
  real Vulkan driver call — all verified correct — and is narrowed to a MoltenVK
  limitation sampling one specific large, non-power-of-two BC3/DXT5 texture
  shape, outside this project's scope.
- Real `wow64cpu.dll` dispatch is not yet exercised — this backend intercepts and
  marshals state directly at the bitness-switch trampoline; the real
  dispatch-table-driven convention is decoded but not wired up as an
  alternate/fallback path.
- `deps/FEX` currently points at a personal fork (`JackTYM/FEX`, `macos-arm64`
  branch) carrying the WoW64-rebase and `MAP_JIT`-guard-page patches described
  above — these need their own upstreaming path before this backend could
  depend on stock FEX.
- One CI-only, backend-agnostic WoW64 gap remains open and documented as a
  follow-up: a guest process's buffered CRT stdio never flushes to a real
  `WriteFile` syscall for a 32-bit WoW64 process specifically (not observed for
  native 64-bit processes), on both this backend and Unicorn — worked around in
  the WoW64 CI smoke test (a direct, unbuffered `WriteFile`-based completion
  marker instead of relying on stdout) but not yet root-caused.
- A pre-existing fault-classification gap: `handle_general_memory_violation`'s
  store/load decode only recognizes the STLR/STLRB/STLRH store-release family,
  so a genuine guest write to a read-only page can be misclassified as a read
  and routed into the wrong fault-recovery path instead of raising
  `STATUS_ACCESS_VIOLATION` to the guest. Not fixed in this PR: an earlier,
  broader attempt at widening this decode table caused an unexplained,
  never-root-caused hang, so this is left as a documented, tracked follow-up
  rather than risking a rushed fix in JIT-adjacent crash-handling code.

## Future work

- Wire up FEXCore's work-in-progress persistent code-cache feature (an on-disk
  JIT cache keyed by binary identity, analogous to how Rosetta 2 avoids
  re-translating a binary on every launch) to cut cold-start JIT latency on
  repeat runs of the same guest program.
- Investigate whether Apple Silicon's hardware TSO memory-ordering mode (used by
  Rosetta 2 to avoid emulating x86 memory ordering with ARM64 acquire/release
  instructions in software) is reachable from a third-party JIT on this
  hardware; if so, it would let this backend drop its
  misaligned-load-acquire/store-release emulation path entirely.
- Background/speculative compilation of likely-next blocks to hide cold-path JIT
  latency behind already-executing code.
- Populate the real WoW64 gate-crossing convention for real `wow64cpu.dll`
  dispatch.
- Extend real multi-vCPU support (proven for the WHP backend, see
  `docs/multi-vcpu-design.md`) to this backend. FEXCore's own `Context`/
  `InternalThreadState` API is built for exactly this — `CreateThread`/
  `ExecuteThread` already support many independent thread states executing
  concurrently on separate host threads, which is how upstream FEX runs real
  multi-threaded Linux binaries — so this backend's current single-active-
  thread, cooperatively-swapped-`CPUState` model is a simplification made to
  get WoW64 correctness working first, not a hard FEXCore limitation.
  Realizing it means giving each logical guest thread its own persistent
  `InternalThreadState` instead of a swapped snapshot, and wiring per-vCPU
  host threads into the existing WHP-proven `vcpu_worker`/kernel-lock
  scheduler. Worth doing for guest workloads with genuinely parallel threads
  (audio mixing, asset streaming, worker pools) where this backend's own JIT
  speed can't substitute for real concurrency — but a substantial rework of
  the current backend, not attempted here.


JackTYM:
Technically boots MW2 on M5 MacOS, but there are some graphics issues (Potentially MoltenVK's end?)
MODERN WARFARE

Boot takes about 2 minutes and animations are clearly slowed, but much faster than the multi-hour boot on Unicorn and has lots of room for optimization

JackTYM:
As a note, this is built upon a modified version of [FEX](https://github.com/JackTYM/FEX) which will not be pushed to the upstream due to their AI assisted code policy

deps/FEX
momo5502, line 1: The thing is, fex has at least 1 submodule, to my knowledge, that's super big and useless for us. That would mean we clone a big bunch of files, for nothing. Is there a solution to that? I see you already use a fork, maybe we can ditch the useless (to us) submodules from there?

momo5502:
Also, the "comctl32.dll" quirk is known and caused by missing SxS support: #1110

momo5502:
Thank you for this awesome contribution. I will review it as soon as I have time.

Also: This fixes #1045

deps/FEX
JackTYM: Good catch, will see what I can remove safely!

JackTYM: Removed 10 unused submodules from FEX, saved 376MB

momo5502:
@claude review

AGENTS.md
momo5502, line 1: i don't think this needs to go here. that is not context that is beneficial for every agent session in the future

src/common/utils/path_key.hpp
```diff
@@ -48,10 +48,20 @@ namespace sogen
...
+                // built `key` from a UTF-16 registry key name in the first place, so going back up to
+                // UTF-16 here only to immediately transcode it right back down inside the
+                // std::filesystem::path constructor below was pure waste. UTF-8<->UTF-16 is identity
+                // for well-formed content, so this is byte-for-byte equivalent to the previous
+                // implementation for every input, unconditionally (verified via a differential harness
+                // against the old implementation, 19 representative registry-path inputs incl. mixed
+                // case, both slash directions, doubled/trailing separators, '.'/'..' components, and
+                // non-ASCII - 0 mismatches).
+                auto native = key.native();
+                std::ranges::replace(native, '\\', '/');
```
momo5502, line 62: not sure if I like this. what if the native type is some weird, non ascii compatible string type? not sure if the standard would allow that, but replacing '\\' with '/' if the underlying type is wchar_t based for example, feels error prone.

momo5502 (changes requested):
so there is a lot going on in this PR. can you please split it up into multiple PRs, each contributing their own, self contained features.

there seem to be a lot of separate features in here that don't seem to directly have anything to do with each other. there is FEX support, there is GDI/user/etc syscall changes, there is macOS vulkan changes, there is WOW64 compatibility changes and probably much more.

it's extremely hard for me to review and understand this bulk of changes and it would be much easier if it were multiple, isolated PRs.

thanks for your understanding.

src/windows-emulator/user_handle_table.hpp
```diff
@@ -113,6 +139,13 @@ namespace sogen
         std::pair<handle, emulator_object<T>> allocate_object(handle_types::type type)
         {
             const auto index = find_free_index();
+            const auto object_handle = make_handle(index, type, false);
+
+            // user32's client-side handle validation (HMValidateHandle and friends) indexes the shared
+            // aheList by the HANDLE's low 16 bits, not by our internal handle id. Since handles are now
+            // 4-aligned (low 2 bits reserved for the NtClose ABI), the id lives at bit 2, so the aheList
+            // slot the guest reads is id << 2. Write the entry there so the client-side lookup resolves.
+            const auto ahe_slot = handle_index_to_ahe_slot(index);
```
Igoorx, line 148: HMValidateHandle doesn't seem to be broken today, so what is this fixing? Maybe it's an outdated issue? If it's just a cosmetic change, I don't think it's worth it for the number of changes this introduces in the patch.

src/windows-emulator/syscalls/user.cpp
```diff
@@ -4623,80 +4642,53 @@ namespace sogen
...
+            // The kernel-side NtUserQueryDisplayConfig ABI is NOT the public QueryDisplayConfig signature. user32's
+            // QueryDisplayConfig (verified by RE of syswow64\user32.dll + wow64win.dll) collapses its six public
+            // arguments into a five-argument syscall:
+            //   NtUserQueryDisplayConfig(flags, pNumPathArrayElements, pPathModalityBuffer, pCurrentTopologyId, 0)
+            // pPathModalityBuffer is a private, 200-bytes-per-path modality buffer that user32 allocates and, on
+            // return, unpacks into the caller's DISPLAYCONFIG_PATH_INFO / DISPLAYCONFIG_MODE_INFO arrays; the public
+            // pNumModeInfoArrayElements / pModeInfoArray pointers are handled entirely user-side and never reach the
+            // kernel. The previous handler assumed the public 6-pointer layout, so it wrote 128 bytes of mode data
+            // over the caller's 4-byte currentTopologyId stack slot (arg 4 here), smashing the return address and
+            // faulting on function return (observed as dcfg32.exe's 0x4016a7 crash).
```
Igoorx, line 4677: This is wrong. The previous syscall handler wasn't 6-arg if you exclude the syscall_context, which I would guess the clanker didn't. And the syscall handler does work as far as I can tell:

```
[*] Starting Display Configuration Test...

[1] Calling GetDisplayConfigBufferSizes...
    Success! Needed Buffers -> Paths: 1, Modes: 2

[2] Calling QueryDisplayConfig...
    Success! Retrieved 2 mode structures.

[3] Iterating Modes & Fetching Device Info...

    --- Mode Index 0 ---
    InfoType: Target
    ID: 0
    Adapter LUID: 0:1000
    Active Size: 1920x1080
    Refresh Rate: 60 Hz
    [>] Requesting Target Name...
        Friendly Name: Generic PnP Monitor
        Device Path:   \\?\DISPLAY#GSM0001#4&1234567&0&UID0#{e6f07b5f-ee97-4a90-b076-33f57bf4eaa7}

    --- Mode Index 1 ---
    InfoType: Source
    ID: 0
    Adapter LUID: 0:1000
    Resolution: 1920x1080
    [>] Requesting Source Name...
        GDI Device Name: \\.\DISPLAY1

[4] Fetching Adapter Name for LUID 0:1000...
    Adapter Path: \\?\PCI#VEN_10DE&DEV_1C03&SUBSYS_00000000&REV_A1#4&1234567&0&0008#{5b45201d-f2f2-4f3b-85bb-30ff1f953599}

[*] Test Complete.
```

So there are two possibilities: the mentioned crash was caused by something else, and the clanker didn't undo this change despite it being ineffective, or there's some use of "QueryDisplayConfig" that causes a crash. That would be fair, but I think fixing it would be more appropriate than just nuking the output.

.github/workflows/ci-reusable.yml
```diff
@@ -814,6 +866,182 @@ jobs:
...
+      # stdout capture can't reliably observe it even when the test genuinely passes.
+      - name: Run WoW64 Smoke Test
+        uses: nick-fields/retry@v4.0.0
+        with:
+          max_attempts: 5
+          timeout_minutes: 10
+          command: |
+            cd build/release/artifacts && chmod +x ./analyzer
+            rm -f root/filesys/c/wow64_test_sample_status.txt
+            ./analyzer -e root c:/wow64-test-sample.exe | tee wow64-smoke-test.log
```
Igoorx, line 1038: This is the new `wow64-test-sample.exe` instead of the normal `test-sample`. It doesn’t appear that the normal `test-sample` is being run at all under WoW64 in FEX, which is important given the number of changes required to get it working there. The PR description mentions that smoke tests were run, but it’s unclear whether they were run only with the x64 `test-sample` or with the x86 version as well.

src/windows-emulator/port.cpp
```diff
@@ -67,6 +60,11 @@ namespace sogen
             return create_lsa_policy_lookup_port();
         }
 
+        if (port == u"\\RPC Control\\Audiosrv" || port == u"\\RPC Control\\AudioClientRpc" || port == u"\\RPC Control\\AudioSrvServiceRpc")
+        {
+            return create_audio_service_port(port);
+        }
+
```
Igoorx, line 67: This wasn't mentioned in the PR body and, more than anything else, definitely seems like one of those things that should be in a separate PR.

JackTYM:
Thanks for the feedback! In hindsight, this was much more of a full push of running MW2 on MacOS than it is adding FEX support explicitly.

I will work more on making a few quality PRs split up by features and do another pass to remove anything that isn't actually needed.

momo5502:
> Thanks for the feedback! In hindsight, this was much more of a full push of running MW2 on MacOS than it is adding FEX support explicitly.
> 
> I will work more on making a few quality PRs split up by features and do another pass to remove anything that isn't actually needed.

Awesome, thank you. Don't take any of the feedback personal. This is still an awesome PR and I would love to see most of it land. It's just much easier for us maintainers if we can review small chunks :)

JackTYM:
> > Thanks for the feedback! In hindsight, this was much more of a full push of running MW2 on MacOS than it is adding FEX support explicitly.
> 
> > 
> 
> > I will work more on making a few quality PRs split up by features and do another pass to remove anything that isn't actually needed.
> 
> 
> 
> Awesome, thank you. Don't take any of the feedback personal. This is still an awesome PR and I would love to see most of it land. It's just much easier for us maintainers if we can review small chunks :)

Of course! I appreciate it

momo5502:
also, as a hint. try testing with MW2 multiplayer. I bet you implemented partial audio support, as the singleplayer crashes with an error. Multiplayer doesn't. it doesn't care about audio. then you can omit that, cause i'm working on full audio support in parallel: https://github.com/momo5502/sogen/tree/wasapi-audio

JackTYM:
> also, as a hint. try testing with MW2 multiplayer. I bet you implemented partial audio support, as the singleplayer crashes with an error. Multiplayer doesn't. it doesn't care about audio. then you can omit that, cause i'm working on full audio support in parallel: https://github.com/momo5502/sogen/tree/wasapi-audio

Thanks for that!

PR #1235, opened 2026-07-20
Windows audio support (WASAPI/DirectSound) + WoW64 fixes

momo5502:
Adds Windows audio support by emulating the Windows Audio Service (Audiosrv) ALPC/RPC interface, so `IAudioClient`/WASAPI and DirectSound clients can open a render stream and play PCM through a host audio sink (SDL). The bulk of the branch brings up the 64-bit path; the most recent work makes the **32-bit (WoW64)** DirectSound path — the one Call of Duty: MW3's Miles Sound System drives — work end to end, with audible in-game sound.

## WoW64 (32-bit) DirectSound fixes (this series)

A 32-bit `DirectSoundBuffer::Play` returned `E_INVALIDARG`/`DEVICE_INVALIDATED` while the 64-bit path worked. Root cause: the emulated Audiosrv replies and ALPC plumbing assumed the 64-bit transfer syntax. Fixed in order of discovery:

- **GetDevicePeriod NDR marshaling** — the two `[in,out,unique] hyper*` out-params must be interleaved (classic NDR) for a 32-bit client, not referents-first (NDR64).
- **WoW64 ALPC reply handle-attribute layout** — reply HANDLE message-attributes were placed at 64-bit offsets; the preceding SECURITY/VIEW/CONTEXT attrs and the HANDLE fields are smaller under WoW64, so the client read the handle count/handle from the wrong offset and aborted the system-handle import (surfaced as `AUDCLNT_E_DEVICE_INVALIDATED`, non-deterministic under WHP). Now sized per guest bitness. Also implements `AudioSessionDestroy` (opnum 55).
- **WASAPI CrossProcessEndpoint control port** — event-driven (EVENTCALLBACK) clients open a dedicated per-stream ALPC port and call `SetEventHandle`; it was served by `dummy_port` (NOT_SUPPORTED). Modeled as `endpoint_control_port`.
- **`NtAlpcDisconnectPort`** — implemented the previously-unimplemented syscall used during stream teardown.
- **EVENTCALLBACK wake-up** — the client fills one buffer then waits on its render event for the engine to report a consumed period. We acked `SetEventHandle` but never signaled the event, so the stream ran but stayed *silent*. The handle delivered by the client→server ALPC HANDLE attribute is now carried through to the port handler and signaled from the render thread.
- **Bounded output latency** — the play cursor was estimated from a wall clock, which drifts ahead of the device clock, growing the host queue into seconds of lag. It is now derived from what the sink actually played (submitted minus still-queued) plus a fixed cushion: self-correcting, while the cushion keeps the device fed (pacing straight off the queue starves it, since the guest's own buffer is only ~20 ms).

## Verification

- 32-bit `dsprobe` → `Play 0x00000000 OK`, deterministic, full opnum sequence (GetMixFormat → IsFormatSupported → GetDevicePeriod → Initialize → CreateStream → GetAudioSession → GetSessionState → StartStream) identical to 64-bit.
- 64-bit `dsprobe64` still OK — no regression.
- A 32-bit tone probe plays a 440 Hz sine end to end.
- **MW3 (`open-iw5.exe`) has audible in-game audio with no perceptible delay**, running clean (no unimplemented syscalls, no crashes).
- Smoke test (`analyzer -s test-sample.exe`) green.

## Notes

`clang-tidy` (tidy preset) has not been run over this series yet. The latency cushion is a single constant (`k_host_queue_cushion_ms`, 60 ms) if it needs tuning per host.


momo5502:
@claude review

docs/audio-capture/render_section_control.bin
momo5502, line 1: i don't want binary files in the repo. please remove them from the history

docs/audio-capture/audioclient_rpc_idl.txt
momo5502, line 1: are all these text docfiles still needed? i feel like they don't serve much purpose

src/samples/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ add_subdirectory(test-sample)
 add_subdirectory(vulkan-shim)
 add_subdirectory(vulkan-shim-test)
 add_subdirectory(steam-shim)
+add_subdirectory(audio-sample)
+add_subdirectory(dsound-sample)
+add_subdirectory(speaker-config-sample)
```
momo5502: i think these samples can be removed. they are nice for development, but unless we actually run & test them in CI, which we don't, cause we have no sound in CI, they are practically useless.

src/windows-emulator/ports/service_control.cpp
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    // PnP / other ntsvcs interfaces (e.g. CM_Register_Notification): an empty success reply is
+                    // enough for the current callers to continue.
+                    return STATUS_SUCCESS;
+                }
+
+                if (getenv("EMULATOR_LOG_RPC"))
+                {
+                    win_emu.log.error("[svcctl] opnum=%u send=%u\n", procedure_id, c.send_buffer_length);
+                }
```
momo5502: we should remove all those logs

momo5502:
@claude review

src/windows-emulator/registry/registry_manager.cpp
```diff
@@ -244,6 +244,93 @@ namespace sogen
...
                                machine / "SOFTWARE" / "RegisteredApplications");
+
+        this->alias_remote_audio_endpoints(machine);
+    }
+
+    // On a headless/server host the local Render/Capture endpoint folders are empty, while an RDP session
+    // populates RemoteRender/RemoteCapture with the redirected device. mmdevapi resolves the default endpoint
+    // by opening MMDevices\Audio\Render\{id} (or Capture\{id}), so alias each remote endpoint into the local
+    // folder it expects, making the RDP audio device usable as the default playback/recording endpoint.
+    void registry_manager::alias_remote_audio_endpoints(const std::filesystem::path& machine)
```
momo5502: not a big fan of this. can't we just prepare the registry, just like steam does it?
in src/tools/grab-registry.bat

PR #1236, opened 2026-07-21
Assorted emulator fixes (split from audio work)

momo5502:
A set of self-contained emulator fixes, split out of the WASAPI/DirectSound audio work (#1235) so they can land independently. Each was found while bringing audio up, but none of them depend on the audio code — they are general correctness fixes.

- **`NtQueryValueKey`: set `DataOffset` for `KeyValueFullInformation`** — the field was left unset, so callers that locate the value data through it (`GetComputerNameW`/`GetComputerNameExW` among others) read the struct header instead and saw an empty string. This unblocks rpcrt4's `RpcStringBindingComposeW`, and with it RPC binding for clients such as mmdevapi.

  The same commit also drops a stray `-1` from the reported `required_size`. That is a small out-of-bounds write fix, not just a cosmetic one: a buffer exactly one byte short of the true size passed the `required_size > length` guard, after which the handler still wrote `base_size + name_size + value_size` bytes — one byte past the caller's buffer.
- **4-align section handles** — rpcrt4's system-handle import requires section handle values to be 4-byte aligned; unaligned values were rejected. Uses the existing `handle_store` index shift, so encoding/decoding stay self-consistent.
- **Stub the MMCSS scheduler device** — `avrt!AvSetMmThreadCharacteristics` opens `\Device\MMCSS\MmThread` to raise a worker's scheduling priority. The emulator has no priority classes, so accepting the open and succeeding its ioctls lets the caller proceed instead of failing. Follows the existing `create_dummy_device` pattern.
- **Default the foreground window to the desktop** — the shared `SERVERINFO` foreground window started out null, so a client-side `GetForegroundWindow()` (which reads `gpsi` directly and never syscalls) returned null before any window activated. Callers that pass the result on — e.g. `IDirectSound::SetCooperativeLevel`, which Miles feeds from `GetForegroundWindow` — then failed with `E_INVALIDARG`. Seeded at construction and re-applied as a fallback on UI events, so a deactivation cannot re-null it.
- **`NtQuerySecurityAttributesToken`: report `BUFFER_TOO_SMALL` for size queries** — a null output buffer is how a caller asks for the required size, so it should be answered like any other undersized buffer rather than rejected as an invalid parameter.
- **Implement `NtAlpcDisconnectPort`** — the syscall was unimplemented, so a guest tearing down an ALPC connection terminated emulation.

## Verification

Builds clean; `analyzer -s test-sample.exe` smoke test passes; `clang-format` check clean.

### DIFF
--- a/src/windows-emulator/process_context.cpp
+++ b/src/windows-emulator/process_context.cpp
@@ -560,6 +560,11 @@ namespace sogen
             window.processId = process_context::process_id;
         });
 
+        // Seed the shared foreground window with the desktop so the guest's client-side GetForegroundWindow
+        // never returns null before an app window activates (UI activation events later refine it).
+        this->user_handles.get_server_info().access(
+            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = this->default_desktop_window_handle.bits; });
+
         const auto create_shell_window = [&](const std::u16string_view class_name, const std::u16string_view title, const int32_t x,
                                              const int32_t y, const int32_t width, const int32_t height) {
             auto [handle, shell_win] = this->windows.create(win_emu.memory);
--- a/src/windows-emulator/process_context.cpp
+++ b/src/windows-emulator/process_context.cpp
@@ -562,8 +562,9 @@ namespace sogen
 
         // Seed the shared foreground window with the desktop so the guest's client-side GetForegroundWindow
         // never returns null before an app window activates (UI activation events later refine it).
-        this->user_handles.get_server_info().access(
-            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = this->default_desktop_window_handle.bits; });
+        this->user_handles.get_server_info().access([&](USER_SERVERINFO& server_info) {
+            server_info.foregroundWindow = this->default_desktop_window_handle.bits; //
+        });
 
         const auto create_shell_window = [&](const std::u16string_view class_name, const std::u16string_view title, const int32_t x,
                                              const int32_t y, const int32_t width, const int32_t height) {
--- a/src/windows-emulator/syscalls/port.cpp
+++ b/src/windows-emulator/syscalls/port.cpp
@@ -134,8 +134,6 @@ namespace sogen
 
         NTSTATUS handle_NtAlpcDisconnectPort(const syscall_context& c, const handle port_handle, const ULONG /*flags*/)
         {
-            // Tear down the ALPC connection. The handle itself is released separately via NtClose; just drop our
-            // port state if we track it.
             c.proc.ports.erase(port_handle);
             return STATUS_SUCCESS;
         }
--- a/src/windows-emulator/syscalls/port.cpp
+++ b/src/windows-emulator/syscalls/port.cpp
@@ -132,6 +132,14 @@ namespace sogen
                                             connection_message, buffer_length, out_message_attributes, in_message_attributes, timeout);
         }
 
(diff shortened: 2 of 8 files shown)


momo5502:
@claude review

src/windows-emulator/syscalls/port.cpp
```diff
@@ -132,6 +132,14 @@ namespace sogen
                                             connection_message, buffer_length, out_message_attributes, in_message_attributes, timeout);
         }
 
+        NTSTATUS handle_NtAlpcDisconnectPort(const syscall_context& c, const handle port_handle, const ULONG /*flags*/)
+        {
+            // Tear down the ALPC connection. The handle itself is released separately via NtClose; just drop our
+            // port state if we track it.
```
momo5502: comment is useless

src/windows-emulator/process_context.cpp
```diff
@@ -560,6 +560,11 @@ namespace sogen
             window.processId = process_context::process_id;
         });
 
+        // Seed the shared foreground window with the desktop so the guest's client-side GetForegroundWindow
+        // never returns null before an app window activates (UI activation events later refine it).
+        this->user_handles.get_server_info().access(
+            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = this->default_desktop_window_handle.bits; });
```
momo5502: Add an empty comment like this, this will force clang-format to break lines and make it more readable

```suggestion
        this->user_handles.get_server_info().access(
            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = this->default_desktop_window_handle.bits; //
             });
```

src/windows-emulator/windows_emulator.cpp
```diff
@@ -1553,8 +1553,14 @@ namespace sogen
...
         // GetForegroundWindow (which reads gpsi directly, never syscalling) returns the active window.
+        // Fall back to the desktop window when no app window is active: real Windows always has a
+        // foreground window, and code that needs a valid HWND (e.g. DirectSound's SetCooperativeLevel,
+        // which Miles feeds from GetForegroundWindow) breaks on a null one.
+        const auto foreground = this->process.foreground_window != 0
+                                    ? this->process.foreground_window
+                                    : this->process.default_desktop_window_handle.bits;
         this->process.user_handles.get_server_info().access(
-            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = this->process.foreground_window; });
+            [&](USER_SERVERINFO& server_info) { server_info.foregroundWindow = foreground; });
```
momo5502: same here, add a trailing comment to break the lambda

momo5502:
Addressed the review:

- **`port.cpp`** — dropped the `NtAlpcDisconnectPort` comment entirely, matching sibling `erase()` call sites.
- **`process_context.cpp` / `windows_emulator.cpp`** — added the trailing `//` to break both `foregroundWindow` lambdas across lines.
- **clang-format** — the failure was the ternary my foreground-window change introduced in `windows_emulator.cpp`. Reproduced the CI check locally (`find … -exec clang-format -style=file --dry-run --Werror`, LLVM 21 to match) and it now reports zero violations across `src`.
- **PR description** — called out the `required_size` `-1` removal explicitly. Verified the reviewer's analysis: with the old value a buffer exactly one byte short passed the `required_size > length` guard and the handler then wrote one byte past it, so it is a genuine (if small) OOB-write fix rather than cosmetic.

PR #1239, opened 2026-07-23
FEX backend (x86-64→ARM64 JIT) for macOS/Apple Silicon (native 64-bit only)

JackTYM:
FEX (x86-64→ARM64 JIT) backend for macOS/Apple Silicon. **Native 64-bit processes only** — WoW64 (32-bit) support is a separate follow-up PR (`feat/wow64-fex`) built on top of this one; this PR alone does not add WoW64 coverage.

## What this adds

- A new FEX-based CPU backend (`sogen::fex`), selected at runtime via `EMULATOR_FEX=1`, mirroring the structure of the existing backends. `deps/FEX` is pinned to a personal fork (`https://github.com/JackTYM/FEX`) rather than upstream directly — the fork carries a 10-submodule removal (376MB trim) and a couple of small opdispatch/behavior patches; see the fork's own commit history for the exact patch list. **Tradeoff:** pinning to a fork means these patches don't automatically track upstream FEX-Emu security fixes — rebasing onto a newer upstream release periodically is a maintenance cost the maintainer should weigh.
- Host-reserved-address-space tracking that stays correct under Apple's `mmap(MAP_FIXED)` semantics: MMIO region insertion properly carves out host-reserved holes instead of violating the overlap invariant, and a new `memory_interface::release_guest_address_range()` distinguishes decommit (keep the address claimed, discard contents) from release (actually free) so a decommitted-but-still-reserved guest range can't be silently handed to an unrelated host allocation. Both have dedicated regression tests in `windows-emulator-test`.
- Smaller correctness fixes in the same area: no host-claim leak when an allocation fails, `try_write_memory`'s permission check covers the full write range instead of just the first byte, and a hard single-instance guard on the global fault-routing pointer (this backend installs a process-wide signal handler, so a second instance would silently clobber the first's).
- One backend-agnostic fix, included here directly rather than as a separate PR since it's a prerequisite for this backend to run FEX-heavy workloads reliably: every new guest thread now seeds its FPU control word / MXCSR to the real x86 power-on reset defaults (`FPCW=0x037F`, `MXCSR=0x1F80`) instead of a zero-initialized state, which previously left every floating-point exception unmasked (including the near-ubiquitous inexact/precision condition) — guest code that reads the control word (e.g. the CRT's `pow`/`_except1` path) would otherwise raise `STATUS_FLOAT_INEXACT_RESULT` on first use, a crash real Windows never produces. Applies to Unicorn and Icicle too, not just this backend.

A related scheduler fix (`windows_emulator::start()`'s quantum-timer watchdog racing `perform_thread_switch`) has been extracted to #1244, at your request, so it can merge independently of this PR.

Linux support is build-tested only — it compiles in CI but has no signal handlers or MMIO emulation implemented yet, and is not expected to run correctly.

## Known limitations (not blockers)

- A guest write to read-only-mapped memory via a plain (non-atomic) `STR` instruction is misclassified by the ARM64 store decoder and aborts the whole emulator process instead of delivering a graceful access violation to the guest (`decode_arm64_store` only recognizes the `STLR` family). Affects code that deliberately writes RO memory and catches the fault (packers/DRM).
- `sync_host_page_apple` unions permissions across 4KB guest slots that share a 16KB host page — a deliberate simplification, documented in-code.
- A fixed-address `NtAllocateVirtualMemory(MEM_RESERVE, explicit base)` doesn't claim at the host level (only the size-only allocation path does), and `commit_memory` doesn't guard against committing inside a `host_reserved` region if a guest probes the right address. Same bug class as the two memory-safety fixes above; a follow-up, not addressed in this PR.

## Testing scope

This PR's `smoke-test-fex` CI job (macOS ARM64, required/gating) runs the real `test-sample.exe` under `EMULATOR_FEX=1`, with retries for a known ASLR-related flaky class (FEX shares the guest address space directly with the host process, so its free-region search competes with host ASLR; a fresh process retry works around it). That's distinct from the separate `build-fex` job (Linux ARM64), which only compile-checks the target.

**This PR has no WoW64 code**, so it can't provide WoW64 test coverage — that's the follow-up `feat/wow64-fex` PR, which runs the real x86 `test-sample.exe` as a WoW64 guest under FEX in its own CI job.

## Verification

Builds clean; `windows-emulator-test` 33/33; the FEX smoke test (`analyzer -e root c:/test-sample.exe` under `EMULATOR_FEX=1`) passes cleanly across 5 consecutive runs; `clang-format --dry-run` clean on every file this PR touches.


momo5502:
@claude review

src/windows-emulator/process_context.hpp
```diff
@@ -28,9 +28,21 @@ namespace sogen
...
-#define GDT_ENTRY_SIZE   0x8
+#ifdef __APPLE__
+// Darwin refuses MAP_FIXED anywhere in the low ~4GB regardless of ASLR (the standard 64-bit
+// Mach-O __PAGEZERO convention, enforced at the mmap syscall level) - a backend sharing the guest
+// address space with the host process (guest VA == host VA, e.g. FEX) can never place anything
+// there. GDT_ADDR is a fixed constant sogen always uses directly (not chosen dynamically via
+// find_free_allocation_base, so the reserved-host-ranges mechanism can't route around it), so it
+// has to live well above that floor here. Chosen far from typical host dyld/heap/stack placement
+// (which stays within a few GB above 4GB) to also avoid the *dynamic*, ASLR-dependent collisions
+// that reserved-host-ranges handles for everything else.
```
momo5502: interesting. does FEX always share the address space with the host? :O
this feels like something that should clearly be documented, as this has clear security implications

src/windows-emulator/windows_emulator.cpp
```diff
@@ -1331,6 +1331,15 @@ namespace sogen
...
                     {
+                        // Under the kernel lock so this switch_thread/stop() pair can't straddle a vCPU's
+                        // own scheduling step: perform_thread_switch consumes switch_thread (exchange to
+                        // false) under the lock, and a preemption whose switch request lands before that
+                        // consume while its stop() only lands inside the next quantum surfaces there as a
+                        // stop with no pending switch - the exact shape of a fatal wind-down, tearing the
+                        // whole run off at a random parked rip. Serialized against the scheduler, the pair
+                        // lands either fully before the consume (plain early switch) or fully inside the
+                        // running quantum (ordinary preemption), never split across it.
+                        const std::scoped_lock kernel_lock(this->kernel_lock_);
```
momo5502: feels like a criticial fix, maybe we can extract that in a separate PR, feels like this should be merged in isolation :D

src/windows-emulator/exception_dispatch.cpp
```diff
@@ -264,6 +264,19 @@ namespace sogen
...
 
+        // FEXCore's JIT translation of INT3 reports RIP already advanced past the trapping 0xCC
+        // (see reports_breakpoint_rip_past_instruction's doc comment) - real hardware/NT
+        // (KiBreakpointTrap) always reports #BP at the INT3 itself, which is what guest SEH/VEH
+        // handlers and ContextRecord->Rip adjustments (e.g. `Rip += 1` to step past it) expect.
+        // KVM/WHP already report the pre-advance address like real hardware, so this is scoped to
+        // the one backend that actually needs it - applying it more broadly landed one byte short
+        // of where KVM/WHP's own guest-visible breakpoints (e.g. anti-debug INT3 padding probes)
+        // expect to resume, causing an infinite re-fault loop there.
+        if (status == STATUS_BREAKPOINT && vcpu.cpu.reports_breakpoint_rip_past_instruction())
```
momo5502: not a big fan of such workarounds, i know they are sometimes needed, we already have such for icicle and unicorn. so if there really is no cleaner solution, let's keep this.

but maybe there is, i feel like it's worth rechecking. maybe the reported rip can be adjusted in the backend, before it reports it.

.github/workflows/ci-reusable.yml
```diff
@@ -498,6 +498,58 @@ jobs:
           path: "build/${{matrix.preset}}/**/CTestTestfile.cmake"
           retention-days: 1
 
+  build-fex:
```
momo5502: why is this not part of the regular build matrix?

deps/CMakeLists.txt
```diff
@@ -121,3 +121,94 @@ add_subdirectory(capstone)
...
+  # keeps fmt fully inline in the consumer TU, avoiding a link dependency on fmt symbols that the
+  # hidden-visibility FEXCore library does not re-export. FEXCore also generates some of its own
+  # headers at build time (Config/ConfigValues.inl, IR/IRDefines.inc, ...) into its binary dir's
+  # include/ - consumers need that include dir too.
+  add_library(fexcore SHARED IMPORTED GLOBAL)
+  set_target_properties(fexcore PROPERTIES
+    IMPORTED_LOCATION "${BINARY_DIR}/FEXCore/Source/${_FEXCORE_SHARED_LIB}"
+    INTERFACE_INCLUDE_DIRECTORIES "${_FEX_SRC}/FEXCore/include;${BINARY_DIR}/include;${_FEX_SRC}/External/fmt/include;${_FEX_SRC}/External/unordered_dense/include"
+    INTERFACE_COMPILE_DEFINITIONS "FMT_HEADER_ONLY=1")
+endif()
```
momo5502: should probably be extracted into a FEX.cmake file and included

src/samples/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ add_subdirectory(test-sample)
 add_subdirectory(vulkan-shim)
 add_subdirectory(vulkan-shim-test)
 add_subdirectory(steam-shim)
+add_subdirectory(cpu-bench-sample)
```
momo5502: do we still need the sample?

src/windows-emulator/memory_manager.hpp
momo5502, line 1: big interface changes here. i get why most of them are needed. give me another day or two to properly think about them.

momo5502:
overall great PR. feels almost perfectly isolated now. one thing i would extract, just because i think it's a critical fix, and in case we ever decide to revert FEX commits, for whatever reason, that fix should absolutely stay.

i haven't had the time to fully finish the review, but given there are a already a few comments to act on, i will revisit once they are addressed.

src/windows-emulator/process_context.hpp
JackTYM: [It does](https://github.com/FEX-Emu/FEX/blob/main/docs/ProgrammingConcerns.md#:~:text=This%20problem%20happens%20because%20FEX%20runs%20both%2032%2Dbit%20and%2064%2Dbit%20guest%20applications%20in%20the%20same%20address%20space%20as%20FEX%20itself), should this be documented in docs?

src/windows-emulator/windows_emulator.cpp
JackTYM: Extracted to #1244

.github/workflows/ci-reusable.yml
JackTYM: An entire project build currently fails on ARM Linux (Though this is likely easily fixable), this task was gated for FEX backend regressions specifically

src/samples/CMakeLists.txt
JackTYM: Nope, removed

src/windows-emulator/exception_dispatch.cpp
JackTYM: This should be fixed in 8fdb591

.github/workflows/ci-reusable.yml
momo5502: i think we should either fix ARM linux or disable fex on that platform and move fex to the build matrix

JackTYM: Should be fixed with #1254

src/windows-emulator/process_context.hpp
momo5502: ah yikes, this is pretty ugly, ngl. but not your fault and that shouldn't stop FEX from getting merged. 
yes, we should document that somewhere, not sure where though.

JackTYM: For now, added a disclosure to fex-backend.md. Maybe once FEX is more viable for real programs, could reference in Wiki

momo5502:
i hope i can find the time to finish review tomorrow. in the meantime: @claude review

cmake/fex.cmake
```diff
@@ -0,0 +1,88 @@
...
+  ExternalProject_Add(fex_external
+    SOURCE_DIR "${_FEX_SRC}"
+    CMAKE_GENERATOR "Ninja"
+    CMAKE_ARGS
+      -DCMAKE_BUILD_TYPE=Release
+      -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+      -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+      ${_FEXCORE_OSX_ARGS}
+      ${_FEXCORE_SANITIZER_ARGS}
+      -DENABLE_LTO=OFF
```
momo5502: why no LTO? i think this should be dependent on SOGEN_ENABLE_LTO

cmake/fex.cmake
```diff
@@ -0,0 +1,88 @@
...
+    set(_FEXCORE_SANITIZER_ARGS -DENABLE_ASAN=ON)
+  else()
+    set(_FEXCORE_SANITIZER_ARGS "")
+  endif()
+
+  ExternalProject_Add(fex_external
+    SOURCE_DIR "${_FEX_SRC}"
+    CMAKE_GENERATOR "Ninja"
+    CMAKE_ARGS
+      -DCMAKE_BUILD_TYPE=Release
```
momo5502: this should use the CMAKE_BUILD_TYPE we passed, and if it's a multi generator build, just fall back to RelWithDebInfo, imo

src/windows-emulator/memory_manager.hpp
```diff
@@ -173,6 +225,17 @@ namespace sogen
         void update_layout_version();
         bool commit_memory(uint64_t address, size_t size, nt_memory_permission permissions, bool allow_image_section);
         memory_permission get_effective_permissions(nt_memory_permission permissions) const;
+
+        // The actual fixed-address allocation logic, without the reserve_host_memory_ranges rescan
+        // the public allocate_memory(address, ...) performs first. reserve_host_memory_ranges itself
+        // calls this (not the public overload) for each backend-reported range, to avoid recursing
+        // back into reserve_host_memory_ranges.
+        bool allocate_memory_raw(uint64_t address, size_t size, nt_memory_permission permissions, bool reserve_only,
```
momo5502: the comments here are very noisy. i think many of them can be condensed

src/emulator/arch_emulator.hpp
```diff
@@ -45,6 +45,22 @@ namespace sogen
...
+        // backends) ABI-compatible with the analyzer that calls through this interface.
+        //
+        // Called once, from module_manager::map_main_modules() right after the process's execution
+        // mode is determined (PE-header-driven detection) and before any module - including this
+        // process's own executable - is mapped. Backends that execute guest code natively on real
+        // x86-64 hardware (KVM/Unicorn/WHP) don't need this - the CPU transparently switches to
+        // compatibility mode on the CS segment load alone, so the default implementation is a no-op.
+        // A JIT-based backend (FEXCore) is fixed-bitness per compiled context and only stands up the
+        // 64-bit one, so it uses this to reject a WoW64 (32-bit) process with a clear error up
+        // front, before its first block would be mis-decoded as 64-bit code.
```
momo5502: comment is very noisy, the vtable movement can be ignored. we compile from source, we don't care if vtable moves at all.

momo5502: we don't guarantee that backend shared libraries are backwards compatible. everything is compiled from one source. mixing artifacts from different builds is not supported.

src/windows-emulator-test/fex_register_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        std::unique_ptr<x86_64_emulator> try_create_fex_emulator()
+        {
+            try
+            {
+                return create_x86_64_emulator(backend_type::fex);
+            }
+            catch (const std::exception&)
+            {
+                return nullptr;
```
momo5502:
```suggestion
                return {};
```

src/windows-emulator/module/module_mapping.cpp
```diff
@@ -220,6 +223,96 @@ namespace sogen
...
+            }
+
+            if (!has_fixup)
+            {
+                return;
+            }
+
+            const auto span_size = static_cast<size_t>(max_end - min_offset);
+            std::vector<std::byte> block_buffer(span_size);
+            memory.read_memory(image_base + min_offset, block_buffer.data(), span_size);
```
momo5502: there is this overload: std::vector<std::byte> read_memory(const uint64_t address, const size_t size) const
no need to preconstruct the vector

src/windows-emulator/syscalls/memory.cpp
```diff
@@ -481,9 +486,46 @@ namespace sogen
...
+                // while the windowed record covers ranges the full scan deliberately omits but the windowed
+                // host_window_is_free probe still reports occupied - without it, a pick landing on such a
+                // range is never recorded as reserved and the loop re-picks the same base until exhaustion.
+                for (int attempt = 0;; ++attempt)
+                {
+                    potential_base = c.win_emu.memory.find_free_allocation_base(
+                        static_cast<size_t>(allocation_bytes), 0, address_requirements.alignment, address_requirements.lowest_address,
+                        address_requirements.highest_address);
+
+                    if (!potential_base || c.win_emu.memory.host_window_is_free(potential_base, static_cast<size_t>(allocation_bytes)))
```
momo5502, line 500: this can race, another thread reserving that address is a classical toctou

src/windows-emulator/module/module_mapping.cpp
```diff
@@ -220,6 +223,96 @@ namespace sogen
             }
         }
 
+        // Applies one relocation block's fixups against a host-side mirror of the block's guest byte
+        // span, then commits the whole span with a single memory.write_memory call. A freshly-mapped
```
momo5502: interestingly, it used to be like that. for some reason this got changed. i like that this optimization is naturally coming back though 😂

src/windows-emulator/module/module_mapping.cpp
```diff
@@ -488,27 +566,55 @@ namespace sogen
...
+            // so the next pick steps past it. Bounded so a genuinely exhausted address space still
+            // terminates rather than spinning.
+            constexpr int max_host_relocation_retries = 8;
+            bool mapped = false;
+            // The free-pick retry loop only makes sense when the caller left the target address up to
+            // us (relocation_base == 0) - if the caller specified a real target (mapping a view of an
+            // already-loaded image at that image's own base, so the view's internal absolute pointers
+            // stay correct), picking a different free host address instead would silently relocate the
+            // view away from where the caller actually needs it.
+            if (relocation_base == 0)
```
momo5502: very noisy comments. the message the comments convey is fine, but bro. your clanker is writing whole books

momo5502: i feel like a whole sweep over the entire PR is needed to rework the comments. there is an explicit note in AGENTS.md/CLAUDE.md, your clanker seems to have ignored it.

momo5502 (changes requested):
Btw, I was done with the review. PR looks good, a few minor things to act on though

src/windows-emulator/module/module_mapping.cpp
JackTYM: Comments should be improved, still a few long-ish ones, but should be significantly better.

JackTYM:
@momo5502 Thanks for this! All changes should be made properly in recent commits, finally got a chance to come back to this

momo5502:
Checked it. PR looks fine. one last pass of @claude review

momo5502 (approved):

PR #1240, opened 2026-07-24
Stop emulator on unimplemented syscalls and callback exceptions

Igoorx:
This PR complements https://github.com/momo5502/sogen/pull/1232 by also stopping on unimplemented syscalls and callback exceptions. The new handling of unimplemented syscalls doesn't seem to break anything per se, but I assume this wasn't an intentional change.


momo5502:
@claude review

PR #1241, opened 2026-07-24
Match the native window creation message flow hierarchy

Igoorx:
This PR improves the Windows emulator’s window-message behavior, bringing it closer to native Win32 semantics through several key fixes:

* It distinguishes child, top-level, and message-only windows during creation and visibility changes:

  * Child windows no longer receive top-level-only messages.
  * Visible child windows now receive child-specific creation messages.
  * Parents are notified through `WM_PARENTNOTIFY` on child creation.

* It reworks paint scheduling around pending invalidation and internal-paint state:

  * `WM_PAINT` is synthesized during message retrieval instead of being persistently posted.
  * Queue-status handling reports paint work only for effectively visible windows.
  * Visibility checks now account for hidden ancestors.
  * `BeginPaint` validates the update region and clears internal-paint state.

* It refines the messages dispatched during creation and show/hide transitions:

  * `WINDOWPOS` values and `SWP_*` flags now more closely match native behavior.
  * Non-message-only windows now receive `WM_ACTIVATEAPP` during activation and deactivation.

* It makes recursive destruction more faithful to native behavior:

  * The full hide and deactivation sequence is limited to the directly destroyed window.
  * Dependent windows receive `WM_DESTROY` without the top-level hide sequence.
  * `WM_DESTROY` is delivered across the hierarchy before `WM_NCDESTROY`.

### DIFF
--- a/src/emulator-platform/platform/window.hpp
+++ b/src/emulator-platform/platform/window.hpp
@@ -171,6 +171,9 @@ namespace sogen
 
 #define EMU_HWND_MESSAGE ((hwnd) - 3)
 
+#define SWP_NOCLIENTSIZE 0x0800
+#define SWP_NOCLIENTMOVE 0x1000
+
 #ifndef OS_WINDOWS
 #define MAXINTATOM                  0xC000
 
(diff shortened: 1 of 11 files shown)


momo5502:
@claude review

momo5502 (approved):

momo5502:
sad. ci fails on windows, now that this was merged. i assume it's still some non-determinism that surfaces due to performance regressions. I hope claude can figure out the root cause, while i'm at the gym.

momo5502:
i think this PR is actually the cause of the CI failure. not 100% confirmed yet, but `process_context::cursor_x` / `cursor_y` is what's diverging

momo5502:
i think this PR is only the trigger, the actually cause might be that the SDL ui backend is simply non-deterministic. i will add a custom ui backend for tests, that should fix it.

Igoorx:
> i think this PR is only the trigger, the actually cause might be that the SDL ui backend is simply non-deterministic. i will add a custom ui backend for tests, that should fix it.

Yeah, that definitely seems like the right call 👍

momo5502:
yep, #1252 fixed it 🎉

PR #1242, opened 2026-07-25
Fixes for WOW64 callbacks

Igoorx:
This PR improves `WINDOWPOS` callback handling to more closely match native Windows behavior:

* Changes made by applications during `WM_WINDOWPOSCHANGING` are now read back from the callback output and applied to the window.
* Window creation, show/hide transitions, and destruction now use the position, size, and `SetWindowPos` flags returned from `WM_WINDOWPOSCHANGING`.
* `WM_WINDOWPOSCHANGED`, `WM_MOVE`, and `WM_SIZE` now report the final window geometry after any callback changes have been applied.
* `WM_WINDOWPOSCHANGED` now receives its own `WINDOWPOS` payload, including the appropriate `SWP_NOCLIENTMOVE` and `SWP_NOCLIENTSIZE` flags.
* `WM_WINDOWPOSCHANGING` callback data is now handled correctly for both native 64-bit and WOW64 processes.

### DIFF
--- a/src/emulator-platform/platform/window.hpp
+++ b/src/emulator-platform/platform/window.hpp
@@ -127,6 +127,17 @@ namespace sogen
         uint32_t flags;
     };
 
+    struct EMU_WINDOWPOS32
+    {
+        uint32_t hwnd;
+        uint32_t hwndInsertAfter;
+        int x;
+        int y;
+        int cx;
+        int cy;
+        uint32_t flags;
+    };
+
     struct EMU_CREATESTRUCT
     {
         pointer lpCreateParams;
--- a/src/samples/test-sample/test.cpp
+++ b/src/samples/test-sample/test.cpp
@@ -1378,6 +1378,96 @@ namespace
         return true;
     }
 
+    bool test_mutable_callbacks()
+    {
+        struct test_state
+        {
+            int changing_count{};
+            int size_width{};
+            int size_height{};
+            UINT changed_flags{};
+            bool mutate{true};
+            bool saw_size{};
+            bool saw_changed{};
+        };
+
+        thread_local test_state* active_state{};
+        test_state state{};
+        active_state = &state;
+
+        WNDCLASSEXA wc{};
+        wc.cbSize = sizeof(wc);
+        wc.lpszClassName = "TestMutMsgQueueClass";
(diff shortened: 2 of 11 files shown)


momo5502:
Even if this is still a draft, I guess it doesn't hurt: @claude review

Igoorx:
> Even if this is still a draft, I guess it doesn't hurt:

Yeah, it's almost finished, so it's alright. Later, I'll just make some fixes to a few other relevant messages, which shouldn't change the current state of things that much.

> Lost documentation comment — src/windows-emulator/syscalls/thread.cpp (handle_NtCallbackReturn, around line 986). The prior code had a comment explaining a non-obvious constraint: user32's SFI-table-driven dispatch stubs always pass a real result pointer with ResultLength=0x18, a 24-byte structure whose first 8 bytes are the actual LRESULT. That comment was dropped in this rewrite and not restored anywhere near the new user_callback_result struct (syscall_utils.hpp:13-21). Per this repo's comment guidelines, that's exactly the kind of external/reverse-engineered constraint worth keeping — recommend re-adding it near the static_asserts in syscall_utils.hpp or at the read site in thread.cpp.

I removed that comment because it's unnecessarily verbose, but I will add a comment back that makes it explicit that overwriting the result value is intentional.

> No WOW64 regression coverage — the new test_mutable_callbacks test in src/samples/test-sample/test.cpp only runs as a native 64-bit test-sample.exe (per the project's smoke-test setup), so it exercises the EMU_WINDOWPOS path but not the EMU_WINDOWPOS32/WOW64 path that this PR specifically claims to fix. Since that's one of the two headline bullets in the PR description, it'd be worth either a WOW64-specific test/build target, or a manual verification note in the PR body confirming it was tested against a real 32-bit binary under WOW64.

I guess Claude missed the `Smoke Test Windows x86 (pull_request)` 😅

Igoorx:
I think this is okay now, although it still has some rough edges that would probably be better addressed on another occasion.

momo5502:
I checked the code, looks fine. I might have some super minor nits here and there, but too minor to be worth commenting, so fuck it. I guess a last round of claude does not hurt though:  @claude review

momo5502:
can be dealt with in follow up PRs

PR #1247, opened 2026-07-25
Fix KSecDD RNG query

Igoorx:
This PR fixes a regression in the KSecDD device introduced by https://github.com/momo5502/sogen/pull/1186, which assumed that the algorithm name would always be provided. However, an RNG query can be made without writing an algorithm name.

### DIFF
--- a/src/emulator/memory_interface.hpp
+++ b/src/emulator/memory_interface.hpp
@@ -62,6 +62,14 @@ namespace sogen
         }
 
         template <typename T>
+        T read_memory(const uint64_t address, const size_t size) const
+        {
+            T value{};
+            this->read_memory(address, &value, std::min(size, sizeof(T)));
+            return value;
+        }
+
+        template <typename T>
         T read_memory(const void* address) const
         {
             return this->read_memory<T>(reinterpret_cast<uint64_t>(address));
--- a/src/windows-emulator/devices/security_support_provider.cpp
+++ b/src/windows-emulator/devices/security_support_provider.cpp
@@ -10,8 +10,16 @@ namespace sogen
 
     namespace
     {
-        // Partial layout of the KsecDD ioctl 0x390400 request, covering the two fields the handler
-        // consumes: the operation selector and the requested algorithm name.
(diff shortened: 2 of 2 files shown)


momo5502:
Failing icicle tests require this PR to be merged: #1248

This bug has always been there, but I think the audio PR introduced a performance regression, which causes the bug to surface.

momo5502:
@claude review

Igoorx:
>This bug has always been there, but I think the audio PR introduced a performance regression, which causes the bug to surface.

That's quite unexpected, I would've assumed that the audio pr would make no difference when the audio isn't being used 🤔 

>Lines 80 and 91 read overlapping memory twice (the base struct is read again as part of the full struct read at line 91) when the algorithm name is present. This is a very small inefficiency, not a correctness issue — likely fine to leave as is for clarity.

Now that I thought more about it, I realized a smarter way to do it :D

src/emulator/memory_interface.hpp
```diff
@@ -61,6 +61,19 @@ namespace sogen
...
+        template <typename T>
+        T read_memory(const uint64_t address, const size_t size) const
+        {
+            if (size > sizeof(T))
+            {
+                throw std::invalid_argument("Memory read size exceeds destination size");
+            }
+
+            T value{};
+            this->read_memory(address, &value, size);
```
momo5502: why not std::min(size, sizeof(T))? then you can drop the throw and the std::min used when calling this function

Igoorx: One thing that comes to mind is that this wouldn’t be entirely obvious to the caller and could theoretically hide bugs in the calling code. However, I think this is only a minor concern for this function, so I’ll go with your suggestion.

src/emulator/memory_interface.hpp
momo5502, line 75: i think both of these methods should assert that T is trivially copyable, but that can be done in another PR. just a generic note

Igoorx: I'll include that in the PR I'm opening next.

PR #1210, opened 2026-07-10
Improve desktop, file-system, and system-information syscalls

Igoorx:
This PR does the following improvements to Sogen:

- [Populates more stack metadata in the TEB](https://github.com/momo5502/sogen/commit/06fd859399ddb49cb31278ce5257a7d56f24e7a5). This is actually just a temporary approximation because stack guard pages are not yet supported. See [SetThreadStackGuarantee](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreadstackguarantee) for more info.
- [Improves directory enumeration metadata reporting](https://github.com/momo5502/sogen/commit/d24d3d454d545d66b9a5d0d1bd0e80a704fa09a1), so directory entries report the host file timestamps.
- [Adds support for ANSI window-class names](https://github.com/momo5502/sogen/commit/4e4aa32d3050707d3ead99d962cdcd607a9449df), so `GetClassNameA` works in the emulator.
- [Creates default desktop shell windows](https://github.com/momo5502/sogen/commit/6bb1eb55d814538923947cdecf74d4470ed71c76), so anti-debug checks do not fail merely because the emulator reports no windows.
- [Reports processor-core topology through `NtQuerySystemInformation`](https://github.com/momo5502/sogen/commit/538eb79c56096e57b0b33b0a5e99368f5268c855).
- [Reports boot-environment information through `NtQuerySystemInformation`](https://github.com/momo5502/sogen/commit/29bd1d5503cb5d4cb14e22ac4738ca360d5e7778).
- [Handles additional unsupported `NtQuerySystemInformation` classes](https://github.com/momo5502/sogen/commit/8ce006c0e97533582afb0904a8e23b1d730cfa9f).
- [Adds more syscall stubs](https://github.com/momo5502/sogen/commit/831921bc93e295ebbbf5b9d6568e9c18749731f3): `NtUserGetPointerDevices`, `NtUserHwndQueryRedirectionInfo`, and `NtGdiDdDDINetDispQueryMiracastDisplayDeviceSupport`.

PR #1248, opened 2026-07-25
Make file timestamps deterministic under the relative-time clock

momo5502:
## Summary
- Host filesystem timestamps (from `stat()`/`compat_stat`) are real wall-clock time, unlike `KUSER_SHARED_DATA`'s clock which is properly virtualized through the injected `utils::clock` abstraction.
- A guest that queries a file's attributes at different real moments across two otherwise-identical runs (e.g. `SerializationTest`'s `ResettingEmulatorWorks`/`EmulationIsReproducible`) observes different `CreationTime`/`LastAccessTime`/`LastWriteTime`/`ChangeTime` values, breaking serialization determinism.
- This is what caused the intermittent Linux Icicle CI failures on `SerializationTest` after #1235 merged: Icicle is slow enough per run to reliably cross the host filesystem's ~1-second timestamp granularity between two touches of the same file, while Unicorn usually finishes fast enough that both touches land in the same second.
- Fix: under `windows_emulator::uses_relative_time()` (the deterministic instruction-tick clock), zero these four timestamp fields instead of reading the host's real `stat()` values, mirroring how real Windows returns no timestamp when one isn't available. Covers all five call sites in `syscalls/file.cpp` (directory enumeration, `NtQueryInformationFile`, `NtQueryInformationByName`, `NtQueryFullAttributesFile`, `NtQueryAttributesFile`).

## Test plan
- [ ] `cmake --build --preset=release` and the existing `windows-emulator-test` suite pass
- [ ] `SerializationTest.*` passes reliably under `EMULATOR_ICICLE=1` (the failure mode this fixes)
- [ ] `analyzer -s test-sample.exe` smoke test still passes

### DIFF
--- a/src/windows-emulator/syscalls/file.cpp
+++ b/src/windows-emulator/syscalls/file.cpp
@@ -350,23 +350,37 @@ namespace sogen
             }
         }
 
+        // Host filesystem timestamps are real wall-clock time, not virtualized like KUSER_SHARED_DATA's clock -
+        // under the deterministic instruction-tick clock (windows_emulator::uses_relative_time), two runs that
+        // touch the same file at different real moments would otherwise observe different values. Zero them
+        // instead; that mirrors real Windows returning no timestamp when one isn't available.
+        LARGE_INTEGER get_file_time(const bool deterministic, const timespec ts)
+        {
+            if (deterministic)
+            {
+                return {};
+            }
+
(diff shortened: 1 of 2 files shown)


Igoorx:
Note that some file enumeration APIs fail if the file time is older than 1970, which was why I made the fix here: https://github.com/momo5502/sogen/pull/1210

PR #1249, opened 2026-07-25
Fix NtTerminateProcess

Igoorx:
This PR is another follow-up to https://github.com/momo5502/sogen/pull/1232 and https://github.com/momo5502/sogen/pull/1240. It ensures that the emulator actually stops when the process is terminated.

This appears to have been a pervasive issue in the multi-vCPU implementation. I suspect that most, if not all, current uses of `c.emu.stop();` should instead be `c.win_emu.stop();`. I may open a PR for this in the future if I encounter this kind of issue again, but of course, if anyone wants to tackle it sooner, feel free!

This PR also adds trivially copyable assertions to the `memory_interface` helpers.

### DIFF
--- a/src/emulator/memory_interface.hpp
+++ b/src/emulator/memory_interface.hpp
@@ -56,6 +56,7 @@ namespace sogen
         template <typename T>
         T read_memory(const uint64_t address) const
         {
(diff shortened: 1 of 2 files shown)


momo5502:
@claude review

PR #1250, opened 2026-07-25
Redesign landing page with ink-on-paper theme

0xeeffoc:


momo5502:
Can you give some context on the redesign, please, e.g. your intentions?

Also, there are a few weirdnesses I want to highlight:

1. darkmode is missing
2. playground and landingpage are inconsistent now, e.g. round buttons + light mode vs. square buttons + dark mode
3. there is quite a lot of reduncancy now: undetectable debugging mentioned multiple times, cpu backends multiple times, system DLLs mentioned multiple times, and much more
4. Not a big fan of this: `Free & open source · GPL-2.0 · Nothing to install`. The only purpose seems to be to fill in space, without providing much value, feels like adlibs in a song. AI seems to like adlibs. "Free & open source" is implied by either the "view source" button right above the text, or by "GPL-2.0" next to it. "Nothing to install" is not fully correct.
5. weird borders around elements:

image

image

image

So, I do like the style, it feels professional, but content wise I'm not too happy.

0xeeffoc:
Mostly wanted to try out Fable 5 😄, and it took inspiration from mixpanel and hexrays
Anyways, added dark mode switch and some other fixes you said

PR #1251, opened 2026-07-25
Implement some missing information classes

Igoorx:
This PR improves support for some process, thread, and system information queries:

- `NtQueryInformationProcess`: Adds support for `ProcessGroupInformation` and reports the process affinity mask through `ProcessBasicInformation`. 
- `NtQueryInformationThread`: Adds support for `ThreadGroupInformation`.
- `NtQuerySystemInformation`: Adds support for `SystemBasicProcessInformation`, `SystemProcessorFeaturesInformation`, and `SystemProcessorFeaturesBitMapInformation`.

These changes fix a regression in an app that previously ran on sogen but broke after the `NtOpenProcess` change introduced in commit 6ae1a1204b5f28bb9cd699049a9e3aa1cf1d7d08. (That change exposed an execution path that had not been exercised before)

### DIFF
--- a/src/emulator-platform/platform/kernel_mapped.hpp
+++ b/src/emulator-platform/platform/kernel_mapped.hpp
@@ -1703,10 +1703,12 @@ namespace sogen
 
     typedef struct _GROUP_AFFINITY
     {
-        KAFFINITY Mask;
+        EMULATOR_CAST(EmulatorTraits<Emu64>::ULONG_PTR, KAFFINITY) Mask;
         WORD Group;
         WORD Reserved[3];
     } GROUP_AFFINITY, *PGROUP_AFFINITY;
+
+    static_assert(sizeof(GROUP_AFFINITY) == 0x10);
 
     struct THREAD_TLS_INFORMATION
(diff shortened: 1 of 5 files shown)


momo5502:
@claude review

Igoorx:
> Missing the leading S — should be SystemCodeIntegrityPolicyManagementInformation. It still compiles (nothing references it elsewhere) and its ordinal value (248) is unaffected since it's positional, but it's clearly unintentional and should be fixed before merge.

Copy-paste mistake, fixed.

> handle_system_basicprocess_information (system.cpp:316) leaves SYSTEM_BASICPROCESS_INFORMATION::ImageName zero-initialized (Buffer/Length = 0) rather than populating it with the process image name. Likely fine as a best-effort stub like other emulated info classes here, but worth a // TODO if intentional, since callers of SystemBasicProcessInformation typically rely on this field.

That's a pre-existing gap from `handle_system_process_information`, I guess I will go ahead and fix that in both places.

momo5502 (approved):

PR #1252, opened 2026-07-26
Dump serialization-mismatch buffers as CI artifacts, add offline bisection tooling

momo5502:
## Summary

We've been chasing a probabilistic `SerializationTest` failure on Windows/Unicorn CI that doesn't reproduce locally and only shows a truncated gtest diff in the log — no way to actually see where the two serialized states diverge. This PR adds the tooling to catch and diagnose it, and then uses that same tooling to find and fix the actual bug.

### Dump/artifact tooling

- On a mismatch, all four `SerializationTest` cases now dump both buffers to `<label>.a.bin`/`<label>.b.bin` under `$GITHUB_WORKSPACE`.
- Every CI job that actually runs the suite (`test`, `win-test`, `smoke-test-node`, `smoke-test-mingw`) uploads them as a `Serialization Mismatch ...` artifact on failure (`if-no-files-found: ignore`, so normal green runs are unaffected). `smoke-test-win32`/`smoke-test-whp`/`smoke-test-kvm`/`smoke-test-android` only run the `analyzer` sample directly, not the gtest suite, so they're not touched.
- Added env-gated tooling to turn a downloaded dump into an actual diagnosis instead of a raw hex dump:
  - `SerializationOffsetFinder.Run` (new test, `SOGEN_OFFSET_FINDER_FILE`) replays a dump through the real `deserialize()` path.
  - `SOGEN_TRACE_OFFSETS=1` prints the byte range of each top-level component (`memory`, `dispatcher`, `process`, ...) as it deserializes.
  - `SOGEN_TRACE_TARGET_OFFSET=<n>` bisects a specific offset down to the owning guest memory region, dumps a hex window, and scans nearby for a plausible embedded FILETIME (the most common cause of this class of bug).
  - All gated behind env vars — zero cost/behavior change on a normal test run.

Along the way, found and fixed a dead assertion in `SerializedDataIsReproducible`: it called `move_buffer()` on both serializers (leaving them moved-from/empty) and then asserted on `get_buffer()` of those now-empty buffers — so the comparison always trivially passed regardless of actual content. It now compares the real captured buffers.

### The actual bug, found using this tooling

The very next push-triggered CI run after this landed caught two real failures. Downloading the dumps and running them through `SerializationOffsetFinder.Run` with `SOGEN_TRACE_OFFSETS`/`SOGEN_TRACE_TARGET_OFFSET` pinpointed the exact diverging field: `process_context::cursor_x`/`cursor_y`, e.g. `(49, 73)` in one run vs `(0, 0)` in the other, with everything else in both ~89MB buffers matching byte-for-byte.

`cursor_x`/`cursor_y` are only updated in `windows_emulator::handle_ui_event()` when a pointer message is delivered through the real (SDL) UI backend's event pump — and that only happens once `pump_events()` drains a pending host-originated SDL event (e.g. on window show), which is a real wall-clock race against `emu.start()` returning, not something gated by the deterministic instruction clock. `test-sample.exe` never calls `SetCursorPos` itself, so this was a race, not guest behavior.

PR #1241 made this newly reachable: it added `test_paint_message_queue()`, the **first** `WS_VISIBLE` top-level window created anywhere in the test suite (every other window, e.g. in `test_message_queue()`, is message-only, parented to `HWND_MESSAGE`, and never actually shown). That's what causes the SDL backend to actually display a window and receive host-originated events for the first time.

Rather than trying to make SDL's real event timing deterministic, the fix removes the race by construction: tests don't need a real window, so `emulation_test_utils.hpp` now defaults `emulator_interfaces.ui` to the existing `null_ui_backend` (a complete no-op already defined in `ui_backend.hpp`, previously only used as an Emscripten/Node fallback) unless a caller explicitly injects one.

## Test plan

- [x] `windows-emulator-test.exe --gtest_filter=SerializationTest.*` — all 4 pass locally (including the fixed `SerializedDataIsReproducible`, now actually comparing real data)
- [x] Verified the dump → artifact → offline-replay pipeline end-to-end: forced a mismatch locally, confirmed `.a.bin`/`.b.bin` are written, then loaded the dump via `SerializationOffsetFinder.Run` with `SOGEN_TRACE_OFFSETS=1` and `SOGEN_TRACE_TARGET_OFFSET` and confirmed it correctly narrows down to the exact committed memory region/offset
- [x] Downloaded the real CI mismatch artifacts from the run this PR's own tooling caught, used the same tooling to root-cause them to `cursor_x`/`cursor_y`
- [x] Full local test suite (`windows-emulator-test.exe`, no filter) passes 27/27 with the `null_ui_backend` change, including `EmulationTest` (which exercises `test-sample.exe`'s new visible-window test)
- [x] `clang-format --dry-run --Werror` clean on all touched files (same clang-format version CI uses)
- [x] YAML validated (`yaml.safe_load`)
- [x] CI green on this PR (`gh pr checks 1252`)


src/windows-emulator/windows_emulator.cpp
momo5502, line 1: can be removed, it was debugging only

src/windows-emulator/memory_manager.cpp
momo5502, line 1: can be removed as well

src/windows-emulator-test/serialization_test.cpp
```diff
@@ -1,7 +1,49 @@
...
+    namespace
+    {
+        // On a serialization mismatch, dumps both buffers to <label>.a.bin/<label>.b.bin under
+        // $GITHUB_WORKSPACE (or the working directory outside CI) and fails the test. CI uploads these as
+        // artifacts; SerializationOffsetFinder.Run (serialization_offset_finder_test.cpp) replays a downloaded
+        // dump through the real deserialize() code with SOGEN_TRACE_OFFSETS/SOGEN_TRACE_TARGET_OFFSET
+        // (windows_emulator.cpp / memory_manager.cpp) to pin down which component owns the divergence.
+        void dump_buffer_to_file(const char* label, const char* suffix, const std::vector<std::byte>& buffer)
+        {
+            const char* dir = std::getenv("GITHUB_WORKSPACE");
```
momo5502: we should not rely on this. it should also work locally. just dump them relative to the working directory

src/windows-emulator-test/serialization_offset_finder_test.cpp
momo5502, line 1: can be removed

src/windows-emulator-test/serialization_test.cpp
momo5502: cmake should tell you what working dir is used

src/windows-emulator-test/serialization_test.cpp
```diff
@@ -1,7 +1,49 @@
...
+            const std::string path = (dir ? std::string(dir) + "/" : std::string()) + label + "." + suffix + ".bin";
+
+            FILE* file = std::fopen(path.c_str(), "wb");
```
momo5502: we also have utils::io for that

src/windows-emulator-test/serialization_test.cpp
```diff
@@ -1,7 +1,49 @@
...
+
 namespace sogen::test
 {
+    namespace
+    {
+        // On a serialization mismatch, dumps both buffers to <label>.a.bin/<label>.b.bin under
+        // $GITHUB_WORKSPACE (or the working directory outside CI) and fails the test. CI uploads these as
+        // artifacts; SerializationOffsetFinder.Run (serialization_offset_finder_test.cpp) replays a downloaded
+        // dump through the real deserialize() code with SOGEN_TRACE_OFFSETS/SOGEN_TRACE_TARGET_OFFSET
+        // (windows_emulator.cpp / memory_manager.cpp) to pin down which component owns the divergence.
```
momo5502: comments are too verbose. be ultra concise. respect the CLUDE.md

momo5502:
Addressed in bfefc86b:
- Removed the SOGEN_TRACE_OFFSETS/SOGEN_TRACE_TARGET_OFFSET instrumentation in `windows_emulator.cpp`/`memory_manager.cpp` and deleted `serialization_offset_finder_test.cpp` entirely - debugging-only, not shipping it.
- `dump_and_expect_equal` now writes relative to the working directory instead of `$GITHUB_WORKSPACE`, so it works identically locally and in CI (CMake's `WORKING_DIRECTORY` already controls where the test runs from).
- Switched to `utils::io::write_file` instead of raw fopen/fwrite.
- Trimmed the comments.
- Updated the CI artifact-upload `path:` globs to match: dumps land in `build/<preset>/artifacts` (the test's working directory), not the workspace root.

PR #1255, opened 2026-07-31
Fix WHP XSAVE state saving

Igoorx:
This PR adds XSAVE state saving/restoration in WHP. I discovered this was necessary after noticing that WHP guests had issues when running code that worked correctly in Unicorn. It turns out that WHP requires the XSAVE state to be saved and restored separately.

### DIFF
--- a/src/backends/whp-emulator/whp_x86_64_emulator.cpp
+++ b/src/backends/whp-emulator/whp_x86_64_emulator.cpp
@@ -42,6 +42,7 @@ namespace sogen::whp
         constexpr uint64_t internal_page_table_base = 0x0000007000000000ull;
         constexpr uint64_t syscall_hook_virtual_address = 0xFFFF800000001000ull;
         constexpr uint64_t unmapped_guest_page = (std::numeric_limits<uint64_t>::max)();
+        constexpr UINT32 xsave_state_capacity = 960;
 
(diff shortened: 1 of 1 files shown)


momo5502:
@claude review

Igoorx:
> This is safe today: the guest XCR0 set in initialize_virtual_processor_state (line 2085-2094) only ever enables bits 0-2 (x87 + SSE + AVX), and for non-compacted XSAVE the resulting area is always exactly 512 (legacy) + 64 (header) + 256 (YMM_Hi128) = 832 bytes, so 960 leaves comfortable margin. That said, it's a magic number with no derivation, and if XCR0 ever gains another bit here (AVX-512 mask/ZMM state, etc.) without updating this constant, WHvGetVirtualProcessorState would simply fail with an E_... HRESULT via WHP_CHECK_HR — not corrupt anything, but worth a short comment on how 960 was derived, or computing it from CPUID leaf 0x0D instead of hardcoding, for future-proofing.

960 is the size required today, not 832. Hard-coding the size has the benefit of producing a clear failure if the required size ever changes, rather than causing an obscure state-restoration failure. However, that may be overthinking it... So I think either approach is fine, and I can make it probe for the correct size if you think it would be worth it, @​momo5502.

PR #1259, opened 2026-08-01
Further improve window message sequencing accuracy

Igoorx:
This PR builds on the sequence-accuracy fixes introduced in https://github.com/momo5502/sogen/pull/1241, further improving the accuracy of the message sequence sent by the Windows emulator.

* Window creation and visibility changes now follow a more complete native message flow:
  
  * Visible top-level windows use the same show sequence during creation as they do through "ShowWindow".
  * Showing a window includes "WM_NCPAINT" and "WM_ERASEBKGND" at the appropriate points in the sequence.
  * Visible descendants also receive non-client and background-paint messages when their parent is shown.
  * Showing a child window can erase its parent’s background before continuing the child’s paint sequence.
  * Activation and "WINDOWPOS" flags are selected based on the window type and the requested show command.

* Paint and invalidation state now behaves more like Windows:
  
  * Handling "WM_ERASEBKGND" does not clear the window’s pending update region.
  * "UpdateWindow" sends "WM_PAINT" and validates the update region through "BeginPaint".
  * Reapplying unchanged window geometry does not create a new update region.
  * Showing a previously hidden window through "SetWindowPos" avoids an additional geometry invalidation.
  * Paint tests now cover the "WM_NCPAINT", "WM_ERASEBKGND", and "WM_PAINT" sequence produced by "ShowWindow" and "UpdateWindow".

* DWM-related calls now expose the non-client-rendering behavior expected by applications:
  
  * Top-level windows receive "WM_DWMNCRENDERINGCHANGED" during creation.
  * Setting "WCA_NCRENDERING_ENABLED" or "WCA_NCRENDERING_POLICY" queues the same notification.
  * Querying "WCA_NCRENDERING_ENABLED" reports non-client rendering as enabled.
  * Window-composition attribute handling now uses the expected guest structure layout.

* A few related compatibility and cleanup issues were also fixed:
  
  * Message retrieval checks queued messages first, pending paint work second, and only then synthesizes due timer messages.
  * "WM_PARENTNOTIFY" during destruction is only sent for the child window directly passed to "DestroyWindow".
  * Desktop thread information now includes the expected flags.
  * Guest GDI allocations are released when handle creation fails or when the corresponding object is deleted.

Working on this PR made me painfully aware of how some software seems to be hanging by a thread, haha. One game would only start if it received the DWM message, while another would enter a loop if it did not receive the ERASEBKGND message, due to an invariant caused by the lack of a `BeginPaint` call.

An current message-sequence diff against native Windows can be found here: https://pastebin.com/raw/9MLkqUNe

### DIFF
--- a/src/emulator-platform/platform/user.hpp
+++ b/src/emulator-platform/platform/user.hpp
@@ -382,6 +382,7 @@ namespace sogen
     enum USER_WINDOWCOMPOSITIONATTRIB : uint32_t
     {
         WCA_NCRENDERING_ENABLED = 1,
+        WCA_NCRENDERING_POLICY = 2,
         WCA_EXTENDED_FRAME_BOUNDS = 8,
         WCA_CLOAKED = 18,
     };
@@ -399,7 +400,9 @@ namespace sogen
     {
         uint8_t unknown0[0x8];
         uint64_t spwndDesktop;
-        uint8_t unknown10[0xEF];
+        uint8_t unknown10[0x30];
+        uint32_t flags;
+        uint8_t unknown44[0xFF];
     };
 
     // NOLINTEND(modernize-use-using,cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays,cppcoreguidelines-use-enum-class)
--- a/src/emulator-platform/platform/window.hpp
+++ b/src/emulator-platform/platform/window.hpp
@@ -262,11 +262,13 @@ namespace sogen
 #define WM_COMMAND                  0x0111
 #define WM_TIMER                    0x0113
 #define WM_HOTKEY                   0x0312
+#define WM_DWMNCRENDERINGCHANGED    0x031F
 #define WM_WINDOWPOSCHANGING        0x0046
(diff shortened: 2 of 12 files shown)


momo5502:
@claude review

Igoorx:
> The previous handle_NtUserGetWindowCompositionAttribute branched on c.proc.is_wow64_process to read a 12-byte (uint32_t-pointer) USER_WINDOWCOMPOSITIONATTRIBDATA layout for WOW64 guests, vs. the native 24-byte layout otherwise. That branch has been removed

Yes, because that branch was incorrect. WoW64 already handles this internally.

> unknown10[0xEF] (239 bytes) was replaced with unknown10[0x30] + flags + unknown44[0xFF], growing the total struct from 0xFF to a much larger size.

That does not matter because the total size is unknown to begin with.

> orchestrator.advance() already validates window existence internally for every path it returns a step from (main queue, descendants, parent-erase), so the !win recursive-retry branch looks unreachable in practice. Not a correctness issue, just redundant defensive code — could be a while loop instead of recursion for clarity, but low priority.

Yes, this is defensive code, so turning it into a while loop would give it more prominence than it needs.

PR #1268, opened 2026-08-03
Implement and fix miscellaneous syscalls

Igoorx:
This PR does the following improvements to Sogen:

- [Implements `NtUserSetMenu`.](https://github.com/momo5502/sogen/pull/1268/commits/653579b3ef215444ebca9690fb08f7d25a9b23ca)
- [Implements `NtUserLoadKeyboardLayoutEx`.](https://github.com/momo5502/sogen/commit/8d8b22b225690822a71d921630227bcb07acfdb7)
- [Fixes dialog base-unit reporting.](https://github.com/momo5502/sogen/commit/e1c7dee9866e2ef442e6196ccd0aecec142c8391)
- [Implements `NtAllocateUuids`.](https://github.com/momo5502/sogen/commit/031e470272f0b8ed3ed636f1b6018912d80b74bf)
- [Implements `NtGdiGetBitmapBits`.](https://github.com/momo5502/sogen/commit/4c9f95c3dd2cc931c4d72fd7a2bc7cee34208486)
- [Implements `NtUserGetWindowPlacement`.](https://github.com/momo5502/sogen/commit/e6dc886abc6156180b386af1c9407bafbe7802a0)
- [Fixes `GetAsyncKeyState` handling for key codes below `0x20`.](https://github.com/momo5502/sogen/commit/44db393f68cb2ab4d1d81ff1ada3c4d3bed34d18)
- [Implements `NtUserFillWindow`.](https://github.com/momo5502/sogen/commit/1805b85faa64d45e1824a8173202e43147524666)
- [Adds support for `SystemPageFileInformation` in `NtQuerySystemInformation`.](https://github.com/momo5502/sogen/commit/71b2e02c47d7408098192e120e9108e04c52782a)
- [Implements `KeyFullInformation` registry-key queries.](https://github.com/momo5502/sogen/commit/0836a1323b73d2ea831ccf2a6c92bbaa17547d86)
- [Adds the Shift JIS charset to `NtGdiEnumFonts`.](https://github.com/momo5502/sogen/commit/ca78cce5249eb2da1203ff162270843104e665b0)
- [Adds additional syscall stubs.](https://github.com/momo5502/sogen/commit/c3e8272d6b371e9ce10f462db7e47b8914690e0d)

### DIFF
--- a/src/backends/whp-emulator/whp_x86_64_emulator.cpp
+++ b/src/backends/whp-emulator/whp_x86_64_emulator.cpp
@@ -42,7 +42,7 @@ namespace sogen::whp
         constexpr uint64_t internal_page_table_base = 0x0000007000000000ull;
         constexpr uint64_t syscall_hook_virtual_address = 0xFFFF800000001000ull;
         constexpr uint64_t unmapped_guest_page = (std::numeric_limits<uint64_t>::max)();
-        constexpr UINT32 xsave_state_capacity = 960;
+        constexpr uint32_t xsave_state_capacity = 0xFFF;
 
         uint64_t align_down_to_page(const uint64_t value)
         {
--- a/src/emulator-platform/platform/gdi.hpp
+++ b/src/emulator-platform/platform/gdi.hpp
@@ -7,6 +7,7 @@ namespace sogen
 #ifndef OS_WINDOWS
 #define ANSI_CHARSET       0
 #define DEFAULT_CHARSET    1
+#define SHIFTJIS_CHARSET   128
 #define GREEK_CHARSET      161
 #define TURKISH_CHARSET    162
 #define VIETNAMESE_CHARSET 163
--- a/src/emulator-platform/platform/process.hpp
+++ b/src/emulator-platform/platform/process.hpp
@@ -1325,8 +1325,19 @@ namespace sogen
         // Followed by SYSTEM_THREAD_INFORMATION<Traits> Threads[NumberOfThreads]
     };
 
+    template <typename Traits>
+    struct SYSTEM_PAGEFILE_INFORMATION
+    {
+        ULONG NextEntryOffset;
+        ULONG TotalSize;
+        ULONG TotalInUse;
+        ULONG PeakUsage;
+        UNICODE_STRING<Traits> PageFileName;
+    };
+
     static_assert(sizeof(SYSTEM_THREAD_INFORMATION<EmulatorTraits<Emu64>>) == 0x50);
     static_assert(sizeof(SYSTEM_PROCESS_INFORMATION<EmulatorTraits<Emu64>>) == 0x100);
+    static_assert(sizeof(SYSTEM_PAGEFILE_INFORMATION<EmulatorTraits<Emu64>>) == 0x20);
 
     struct PROCESS_PRIORITY_CLASS
     {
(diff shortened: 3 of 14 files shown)


momo5502:
@claude review

Igoorx:
Looks like there are still some issues worth addressing, brought up by ChatGPT:

---

**[P1] Preserve `GetBitmapBits` size-query and bitmap-format semantics**
`src/windows-emulator/syscalls/gdi.cpp:2005–2021`

The handler immediately returns `0` when `bits == 0` and calculates the result as `surface.pixels.size() * sizeof(uint32_t)`. That exposes the emulator’s internal 32-bpp surface rather than the bitmap’s actual format. The surface model separately tracks `guest_bpp`, stride, orientation, and already supports 24-bpp DIB sections.

`GetBitmapBits(hbitmap, count, nullptr)` is a valid size query and should return the bitmap’s full byte size regardless of `count`. Negative counts also request the full bitmap. The size and copied layout use the bitmap’s actual bit depth with 16-bit-aligned scanlines; existing compatibility tests explicitly cover null buffers, negative counts, and non-32-bpp bitmaps.

This breaks the common two-call pattern:

```cpp
const auto size = GetBitmapBits(bitmap, 0, nullptr);
std::vector<std::byte> data(size);
GetBitmapBits(bitmap, size, data.data());
```

Compute the DDB stride from the bitmap’s real bpp, return the complete size for a null destination, and convert the internal pixels into the appropriate depth and row order before copying.

---

**[P2] Use the classless sentinel in `KeyFullInformation`**
`src/windows-emulator/syscalls/registry.cpp:132`

The implementation writes:

```cpp
info.ClassOffset = required_size;
```

but leaves `ClassLength` as zero and writes no class bytes.

For a key without a class, Windows reports `ClassOffset == 0xffffffff` and `ClassLength == 0`. `ClassOffset == offsetof(KEY_FULL_INFORMATION, Class)` is used only when class data actually follows the fixed header.

Since Sogen currently has no class data here, this should be:

```cpp
info.ClassOffset = UINT32_MAX;
```

---

**[P2] Return the correct failure value from `NtUserVkKeyScanEx`**
`src/windows-emulator/syscalls/user.cpp:6310`

The new stub is declared as `BOOL` and returns `FALSE`, so a failed lookup returns zero.

The syscall’s return type is a 32-bit packed result, with `0xffffffff` as the failure sentinel. Reference implementations initialize the result to `(DWORD)-1`.

Even while stubbed, it should use a `uint32_t`/`DWORD` return type and return `UINT32_MAX`; otherwise callers cannot distinguish failure correctly.

---

**[P2] Preserve `NtUserEnableMenuItem`’s return contract**
`src/windows-emulator/syscalls/user.cpp:5582`

This stub is declared as `BOOL` and unconditionally returns `TRUE`.

The syscall returns a `UINT` containing the item’s **previous state**, or `UINT_MAX` on failure—not a Boolean success flag.

Until menu-state mutation is implemented, returning `UINT_MAX` is safer than reporting a fabricated previous state.

Igoorx:
> WHvGetVirtualProcessorState(this->partition_, this->vp_index_, WHvVirtualProcessorStateTypeXsaveState, bytes.data() + register_bytes + sizeof(xsave_size), xsave_state_capacity, &xsave_size) failed with HRESULT 0x80370301

Looks like the XSAVE state can have different sizes, perhaps depending on the CPU. I’ll go ahead increase its capacity to a safer value.

momo5502:
@claude review

PR #1269, opened 2026-08-03
VKD3D bring-up

Igoorx:
This PR implements several extensions and entry points that were missing from the GPU bridge and are required for VKD3D to work.

Tested using the Mario 64 PC port with DirectX 12 through VKD3D:

image

I also tested CoD MW3, and it still works and renders after this patch, so there was no DXVK regression.

**Disclaimer:** This PR was largely vibe-coded. I did multiple review passes with an LLM and made an effort to clean up the implementation, but I can’t confidently say I fully understand whether everything is as correct as it could be.

### DIFF
--- a/src/gpu-bridge-protocol/gpu_bridge_protocol.hpp
+++ b/src/gpu-bridge-protocol/gpu_bridge_protocol.hpp
@@ -181,9 +181,13 @@ namespace sogen::gpu_bridge
         cmd_draw_indirect_count = 0x88E,
         cmd_next_subpass = 0x88F,
         free_descriptor_sets = 0x890,
-        get_shader_module_identifier = 0x891,
-        get_shader_module_create_info_identifier = 0x892,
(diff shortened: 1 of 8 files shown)


momo5502:
@claude review

Igoorx:
To make this PR a little less annoying to deal with, I’ll split it into four parts, with this first one being basically just low-friction new entry points.

momo5502:
@claude review

Igoorx:
#1 addressed, about #2:

> feature_struct_size() is missing the EXT sType for vertex-attribute-divisor

`VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VERTEX_ATTRIBUTE_DIVISOR_FEATURES` has the same value as `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VERTEX_ATTRIBUTE_DIVISOR_FEATURES_EXT`, so this is fine.

momo5502 (approved):

PR #1302, opened 2026-08-22
Add FEX backend support for Android

Igoorx:
This PR gets the FEX backend working on Android, supporting both native compilation on Android and cross-compilation (though the latter is untested). Performance is good overall, though a few bottlenecks remain (rendering in particular).

Here's a short demo of Sonic Mania running on my phone through Sogen inside Termux. It combines the Steam assets with the executable from the [Sonic-Mania-Decompilation project](https://github.com/RSDKModding/Sonic-Mania-Decompilation), with Termux:X11 for display, PulseAudio for audio, and Freedreno for hardware-accelerated graphics:

https://streamable.com/wd63dr

As the video shows, audio and external keyboard input both work. The game holds full speed at 480p, but the frame rate drops noticeably at 720p.

Some additional notes:

* This PR covers only the FEX changes. It isn't enough to actually play games because a few SDL issues still need fixing, I plan to address those in a follow-up PR.
* CI now includes an Android ARM64 build to catch compilation/build regressions, but tests are not run, since the GitHub CI runner doesn't provide Android ARM64 hosts or ARM64 nested virtualization.
* The FEX backend still only supports x64 apps.

Thanks to @JackTYM for making this viable!

### DIFF
--- a/src/backends/fex-emulator/fex_x86_64_emulator.cpp
+++ b/src/backends/fex-emulator/fex_x86_64_emulator.cpp
@@ -1,4 +1,4 @@
-// _GNU_SOURCE is required for mremap() (used to alias host memory into the guest address space).
+// _GNU_SOURCE exposes the Linux host VM and signal definitions used by this backend.
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
@@ -6,6 +6,8 @@
 #define FEX_EMULATOR_IMPL
 #include "fex_x86_64_emulator.hpp"
 #include "fex_x86_64_common.hpp"
+
+#include "address_utils.hpp"
 
 // FEX (https://fex-emu.com) is an in-process x86-64 -> AArch64 binary translator. Unlike the
 // Unicorn/Icicle/KVM backends it does not manage a sandboxed guest address space: the translated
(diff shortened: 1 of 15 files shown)


src/backends/fex-emulator/fex_x86_64_emulator.cpp
```diff
@@ -78,6 +84,52 @@ namespace sogen::fex
...
+#ifdef __ANDROID__
+        constexpr uint64_t guest_address_space_end = 0x00007ffffffeffffULL + 1;
+
+#ifndef MAP_FIXED_NOREPLACE
+        constexpr int MAP_FIXED_NOREPLACE = 0x100000;
+#endif
+
+        std::vector<host_reserved_range> read_host_mappings_android(const uint64_t window_start, const uint64_t window_end)
+        {
+            std::unique_ptr<FILE, decltype(&::fclose)> maps{::fopen("/proc/self/maps", "r"), &::fclose};
```
momo5502: we do have these: https://github.com/momo5502/sogen/blob/main/src/common/utils/io.hpp#L16

Igoorx: Using these helpers makes the parsing a bit more involved, but if the goal is to avoid C APIs, then I guess it’s a reasonable tradeoff.

PR #1306, opened 2026-08-25
Add a seven backend (seven-jit-powered x86_64_emulator implementation)

pinefin:
## Summary
- Adds a new `x86_64_emulator` backend (`backend_type::seven`) implemented on top of [binsnake/seven-jit](https://github.com/binsnake/seven-jit) -- a from-scratch C++23 x86-64 semantic emulator (`seven_core`, sharing the iced-x86 decoder) with an asmjit-backed native-codegen fast path (`seven_jit::JitExecutor`) layered on top: hot code compiles to native x86-64 once per address, everything the compiler doesn't cover falls back to the plain interpreter one instruction at a time, so hook/trap fidelity is identical either way -- only throughput on the compiled subset changes.
- `src/backends/seven-emulator/`: the glue implementing SOGEN's `x86_64_emulator` interface (registers, memory, hooks, serialize/deserialize) against `seven_jit::JitExecutor`. Builds `iced_x86`/`softfloat`/`seven_core`/`seven_jit`/`asmjit` out of an out-of-tree seven-jit checkout (`SEVEN_ROOT` cache var), gated behind a new `SOGEN_ENABLE_SEVEN` option (default `OFF`), matching the existing `fex`/`whp`/`kvm` optional-backend pattern.
- `backend-selection`: adds `backend_type::seven` alongside the existing `unicorn`/`icicle`/`whp`/`kvm`/`fex` options.
- Also includes an on-demand stack/scratch-region growth fallback in `write_memory` for guests that write just past the currently-committed stack region (the classic guard-page stack-growth pattern) -- unicorn/whp accept this implicitly at the OS level, seven's page model is strict about only touching explicitly-mapped pages, so this backend grows the missing pages RW (never downgrading existing permissions) and retries, matching what a real OS does on a guard-page fault.

## Why
Wanted a backend with a real interpreter-plus-JIT throughput story instead of picking between "fast but not source-available/less inspectable" (WHP/KVM) and "slow but reference-quality" (a plain interpreter). seven_core's decoder/handler semantics are hardware-validated (libLISA measurements), and seven-jit's compiled fast path gets within a small constant factor of that fidelity at much higher throughput on repeatedly-executed code.

## Test plan
- [x] Builds cleanly as an optional backend (`SOGEN_ENABLE_SEVEN=ON`) alongside the existing backends, no changes to any backend that was already there.
- [x] Exercised as the default backend of a downstream consumer project driving `windows_emulator` through real Windows PE binaries -- register/memory/hook/serialize paths all exercised end to end, matches existing backends' behavior on the same inputs.

Two commits: the initial backend, and a follow-up that swaps the plain interpreter for seven-jit's `JitExecutor`. I'll keep appending commits here rather than force-pushing over history if anything else comes up.

### DIFF
--- a/src/backends/seven-emulator/CMakeLists.txt
+++ b/src/backends/seven-emulator/CMakeLists.txt
@@ -1,0 +1,105 @@
+# seven-emulator SOGEN backend.
+# Builds binsnake/seven-emulator's libraries (iced_x86, softfloat, seven_core) from an
+# out-of-tree source checkout and the glue that implements SOGEN's x86_64_emulator on top of
+# seven. seven is a from-scratch C++23 x86 semantic emulator that shares the iced-x86 decoder
+# with Mergen, which makes it the reference oracle for VM devirt divergence analysis.
+
+set(SEVEN_ROOT "C:/Users/vanilla/Desktop/alpha/claude/seven-emulator" CACHE PATH "seven-emulator source root")
+
+if(NOT EXISTS "${SEVEN_ROOT}/seven_core/include")
+  message(FATAL_ERROR "SEVEN_ROOT does not point at a seven-emulator checkout: ${SEVEN_ROOT}")
+endif()
+
+# SIMD profile — match seven's AVX512 default (emulated in software via wide-integer).
+set(_SEVEN_PROFILE "AVX512")
+set(_SEVEN_VBITS 512)
+set(_SEVEN_VBYTES 64)
+
+# ---- softfloat (C) ----
+if(NOT TARGET softfloat)
+  set(_SF "${SEVEN_ROOT}/third_party/softfloat")
+  file(GLOB _SF_SRCS "${_SF}/source/*.c" "${_SF}/source/8086/*.c")
+  list(FILTER _SF_SRCS EXCLUDE REGEX "M\\.c$")
+  list(FILTER _SF_SRCS EXCLUDE REGEX "M_")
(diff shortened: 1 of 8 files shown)


src/backends/seven-emulator/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
+# seven-emulator SOGEN backend.
+# Builds binsnake/seven-jit's libraries (iced_x86, softfloat, seven_core, seven_jit) from an
+# out-of-tree source checkout and the glue that implements SOGEN's x86_64_emulator on top of
+# seven_jit::JitExecutor. seven_core is a from-scratch C++23 x86 semantic emulator sharing the
+# iced-x86 decoder with Mergen; seven_jit adds an asmjit-backed native-codegen fast path on top
+# (JIT-compiled hot blocks, interpreter fallback for anything uncovered), which is what makes this
+# the reference oracle for VM devirt divergence analysis fast enough to use as the default trace
+# backend instead of unicorn.
+
+set(SEVEN_ROOT "C:/Users/vanilla/Desktop/alpha/binsnake/seven-jit" CACHE PATH "seven-jit source root")
```
momo5502, line 10: 😵‍💫

momo5502: can we wire it up as a submodule, like the other dependencies?

PR #1310, opened 2026-08-28
gdb-stub: use range execution hook for exec breakpoints

williballenthin:
To support an unpacking project, I'd like to be able to set hardware execution breakpoints on a range (actually, a full section). Seems like there's a lot of overhead today (per-instruction): one hashmap entry for icicle (O(n)), one list entry for Unicorn (O(n**2)).

It seems this handling can be simplified to use the `hook_memory_range_execution` API, which I *believe* was implemented after the GDB handler.

Of course, I'm not yet very intimate with this project, so don't hesitate to push back if this change doesn't make sense to you.

---

Claude told me:

The per-byte hook loop dates from April 2025 when the simplified hook interface only offered single-address execution hooks.  The range API was re-added in May 2026 (3f28e5ef) for WHP section tracking but the GDB stub was never updated.  Replace the O(N) individual hooks with a single hook_memory_range_execution call, matching what the read/write watchpoint paths already do.

### DIFF
--- a/src/linux-gdb-stub/x64_gdb_stub_handler.hpp
+++ b/src/linux-gdb-stub/x64_gdb_stub_handler.hpp
@@ -187,10 +187,7 @@ namespace sogen
             try
             {
                 return this->hooks_.access<bool>([&](hook_map& hooks) {
-                    auto hook_vector = this->create_hook(type, addr, size);
-
-                    hooks[{addr, size, type}] = scoped_hook(*this->emu_, std::move(hook_vector));
-
+                    hooks[{addr, size, type}] = this->create_hook(type, addr, size);
                     return true;
                 });
             }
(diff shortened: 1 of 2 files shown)


src/linux-gdb-stub/x64_gdb_stub_handler.hpp
```diff
@@ -256,19 +256,11 @@ namespace sogen
 
         std::vector<emulator_hook*> create_execute_hook(const uint64_t addr, const size_t size)
```
momo5502: yeah, that was added later on, question is if all emulation backends support it? i hope they fall back to byte-wise hooks if not. so it's definitely worth keeping this change. it's a great optimization and we should make adjustments in the backends instead, if this causes issues.

the question is now, is std::vector still the desired return type? if i saw it correctly, all other hook creation methods also only return a single hook. might make sense to just switch to `emulator_hook*`

williballenthin: sorry, the best I can do here is be a meat proxy:

Claude:
image

I looked into this (with Claude) before proposing the changes, and understood that the backends should all support `hook_memory_range_execution` cleanly, except KVM, which apparently already has an execution hook limitation (only fires during single stepping or after syscall/exception traps?).

I was hoping that CI would flag incorrect behavior, so will be watching that carefully.

Let me see if simplifying the return type works out - I agree that would be nice.

williballenthin: simplifying the return type seems reasonable, PR updated with second commit.

momo5502: yeah, it's fine. KVM limitation can probably be resolved in a similar way to WHP, but that's not in the scope of this PR.

I am willing to accept potential regressions in that case. Frontends implementing workarounds for backends is ugly and I'm glad this one's gone. 

If the return type is changed (or it's confirmed that vector is still needed), then the PR is fine to be merged :D

williballenthin: should be good to go, thanks for considering @momo5502

src/windows-gdb-stub/x86_64_gdb_stub_handler.hpp
```diff
@@ -293,60 +290,42 @@ namespace sogen
...
-                return this->create_write_hook(addr, size);
-            case hardware_read_write: {
-                auto hooks1 = this->create_hook(hardware_read, addr, size);
-                auto hooks2 = this->create_hook(hardware_write, addr, size);
-                hooks1.insert(hooks1.end(), hooks2.begin(), hooks2.end());
-                return hooks1;
-            }
+                return {*this->emu_, this->create_write_hook(addr, size)};
+            case hardware_read_write:
+                return {*this->emu_, {this->create_read_hook(addr, size), this->create_write_hook(addr, size)}};
```
momo5502, line 328: ah yeah, i see, this is why vector was still needed, but seems fine

momo5502 (approved):

PR #1312, opened 2026-08-31
feat(macos): emulate macOS arm64 user space

kaganisildak:
Adds a Darwin personality to sogen: enough of the Mach-O loader, arm64 ABI, BSD kernel, Mach IPC, WindowServer and AppKit surface to run a real, unmodified GUI application. /System/Applications/Calculator.app is the reference subject; it launches, composes its own window and responds to clicks and keystrokes.

Sogen applies no fixups, relocations or symbol binding of its own. It maps the guest's executable and /usr/lib/dyld and starts dyld's entry point, so the real dynamic loader does that work exactly as it does on hardware. That single decision is what keeps the surface tractable.

What is here:

  - Mach-O loader and the dyld shared cache, including a demand pager for hosts that cannot mmap 5.4 GiB
  - BSD syscalls, 128 of 456 measured against xnu-12377.121.6 and held by a ratcheted test, plus the commpage and a guest/host filesystem
  - Mach IPC: port namespace, mach_msg2, hand-written MIG subsystems, the IOKit and IOSurface user clients, XPC bootstrap
  - Concurrency: kqueue, workqueues, psynch mutexes and condition variables, the pieces libdispatch needs to schedule anything
  - A window server: SkyLight connection bring-up, the CALayer tree AppKit builds, a software compositor and an SDF rasteriser
  - A wasm64 browser build, served as a third mode of the playground beside windows and linux, with a synchronous range bridge into the shared cache and a service-worker block cache for the root

Guest writes are confined to the emulation root: any mutation whose real path resolves outside it is refused. Reads still resolve through the root's symlinks into the host, which is deliberate and documented.

Not yet working: the keypad tint is unread because it lives in SwiftUI's Swift-side state rather than in the layer tree, TextEdit trips a libdispatch workloop assertion, and the browser reaches a window less reliably than the native build because it interprets where native JITs.

docs/macos-emulation.md is the entry point; src/tools/setup-macos-env.sh prepares a machine from scratch.


kaganisildak:
@momo5502 did you checked ?

momo5502:
> @momo5502 did you checked ?

No, I never start reviewing until CI is green and AI review passes.

PR #1334, opened 2026-09-09
feat(registry): stub NtCreateRegistryTransaction and NtOpenRegistryTransaction

JackTYM:
## Summary

Register `NtCreateRegistryTransaction` and `NtOpenRegistryTransaction` (both exported by the Windows 10+ `ntdll.dll` in the emulation root, syscall ids `0xC0` / `0x12F`) as handlers that return `STATUS_NOT_SUPPORTED` (`0xC00000BB`). Sogen's registry is a plain hive with no Kernel Transaction Manager (TxR) support, and `STATUS_NOT_SUPPORTED` is what a Windows system without transacted-registry support returns, so callers take their ordinary non-transacted fallback path. The output handle is left untouched on the failure path, as NT does.

Before this change, the first call aborts the whole emulation through the unimplemented-syscall path.

## Reproduction

Any x64 guest that calls either export. The new `test-sample` section is the minimal one: resolve both from `ntdll.dll` via `GetProcAddress`, call `NtCreateRegistryTransaction(&h, TRANSACTION_ALL_ACCESS, &oa, 0)` and `NtOpenRegistryTransaction(&h, TRANSACTION_ALL_ACCESS, &oa)`.

Without the handlers registered (the two `add_handler` lines removed, everything else identical), `test-sample.exe` (x64, cross-built with the repo's `cmake/toolchain/mingw-w64.cmake`) under `--backend unicorn`:

```
Executing syscall: NtCreateRegistryTransaction (0xC0) at 0x1800a1bf2 via 0x1400e162d (test-sample.exe)
Unimplemented syscall: NtCreateRegistryTransaction - 0xC0 (raw: 0xC0)
Emulation failed at: 0x1800a1bf4 - Emulation terminated without status
```

The analyzer exits with status 1 and guest stdout stops at `Running test 'Registry Transaction': ` with no result; every test after it never runs.

## Verification

Same `test-sample.exe`, this branch:

- `--backend unicorn`: both syscalls dispatched (`NtCreateRegistryTransaction (0xC0)`, `NtOpenRegistryTransaction (0x12F)`), `Emulation terminated with status: 0`, guest stdout shows `Running test 'Registry Transaction': Success`, 30 `Success` / 0 `Fail` across the whole sample.
- `--backend fex`: identical result (both syscalls dispatched, status 0, 30/0).
- `windows-emulator-test` with `EMULATOR_ROOT` pointing at a root containing the rebuilt sample: `50 tests from 11 test suites ran ... [  PASSED  ] 50 tests.` The `EmulationTest.*` cases run the sample end to end, so the new section is exercised on every CI run that rebuilds `test-sample.exe`.
- Host: macOS arm64, `cmake --build --preset=release`.

Note the sample check accepts either `STATUS_NOT_SUPPORTED` or success from the create call (so a future real TxR implementation does not have to touch the test), and only requires the open call to fail, since no transaction object exists to open.

## Files

- `src/windows-emulator/syscalls/registry.cpp` — the two handlers
- `src/windows-emulator/syscalls.cpp` — declarations + `add_handler` registrations
- `src/samples/test-sample/test.cpp` — `test_registry_transaction` + `RUN_TEST` entry, `<winternl.h>` include

### DIFF
--- a/src/samples/test-sample/test.cpp
+++ b/src/samples/test-sample/test.cpp
@@ -20,6 +20,7 @@
 #include <intrin.h>
 
 #include <windows.h>
+#include <winternl.h>
 #include <timeapi.h>
 #include <winsock2.h>
 #include <ws2tcpip.h>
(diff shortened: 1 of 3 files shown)


src/samples/test-sample/test.cpp
momo5502, line 1: what's the point of adding a test for a feature that is unsupported?

momo5502: honestly, either implement the feature, or remove that test. but this is useless.

JackTYM: The test was just to show that the crash was gone though makes sense why you don't want it, will remove it

momo5502 (approved):

PR #1355, opened 2026-09-14
Fix RIP lifecycle for NtContinue and user callbacks

warsang:
**Problem**: After a syscall completes, KVM/WHP applied an unconditional rip += 2 to skip past the synthetic syscall instruction. 
Some syscall handlers: NtContinue, NtCallbackReturn, and user-callback dispatch (which installs ki_user_callback_dispatcher directly), already set RIP to an exact guest target before returning. The blind +2 then ran on top of that, landing execution 2 bytes into the target (typically inside INT3 padding), and in the exception path, producing a misleading crash instead of surfacing the guest's real fault.

**Fix:**

Adds an instruction_pointer_finalized flag on syscall_context, set via mark_instruction_pointer_finalized() at each site that installs an explicit RIP (NtContinue, NtCallbackReturn, callback dispatch).

This changes the syscall dispatch function's return type from void to instruction_hook_continuation so backends learn generically whether RIP was already finalized, instead of guessing.
Updates KVM, WHP, Icicle, and FEX to skip their automatic advance when finalized. (Unicorn's hook model already compensates for this differently).

Fixes exception-stop paths to call win_emu.stop() instead of emu.stop(), so a hard error / unhandled exception halts the whole emulator rather than falling through into KiUserExceptionDispatcher's padding.
Syncs thread.previous_ip/current_ip on NtContinue so a later exception is reported against the restored context, not the syscall that triggered it.

**Testing**: ninja windows-emulator-test builds and runs on Linux (release preset). 27/27 tests pass locally and CI passed

### DIFF
--- a/src/windows-emulator/handles.hpp
+++ b/src/windows-emulator/handles.hpp
@@ -188,6 +188,7 @@ namespace sogen
       public:
         using index_type = uint32_t;
         using value_map = std::map<index_type, T>;
+        using iterator = typename value_map::iterator;
 
         bool block_mutation(bool blocked)
         {
@@ -270,7 +271,7 @@ namespace sogen
             return h;
         }
 
-        std::pair<typename value_map::iterator, bool> erase(const value_map::iterator& entry)
+        std::pair<iterator, bool> erase(const iterator& entry)
         {
             if (this->block_mutation_)
             {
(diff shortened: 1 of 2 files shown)


momo5502:
@codex review

momo5502:
@claude review

PR #1356, opened 2026-09-14
Support ARM64EC/CHPE images in x64 backend mapping

warsang:
**What this adds**: Recognizes ARM64EC (0xA641) as a valid PE machine type in the module loader (previously only I386/AMD64 were accepted; ARM64EC images threw "Unsupported architecture!"). For ARM64EC images specifically, treats sections marked CNT_CODE (but not MEM_EXECUTE) as executable mirroring how Windows applies CHPE hybrid code-map permissions to the x64-compatible entry/bootstrap range when it builds the image VAD. Without this, that range would load non-executable and never be reachable.

(Quick note on Arm64EC: sogen just runs the embedded x64 machine code in that entry/bootstrap range directly on its normal x64 backend, exactly like any other x64 code it loads so we're not changing the emulation backend arch to ARM and emulating any arm instructions)

This diff also carries two unrelated cleanups: adding using iterator = typename value_map::iterator; and fixing what look like truncated declarations (std::pair<typename value_map:: / value_map:: on their own) into proper std::pair<iterator, bool> erase(...) and iterator end() signatures, in both module_mapping.cpp and user_handle_table.hpp. 

**Testing**: Emulated an Arm64EC PE file; saw a trace

### DIFF
--- a/src/windows-emulator/handles.hpp
+++ b/src/windows-emulator/handles.hpp
@@ -188,6 +188,7 @@ namespace sogen
       public:
         using index_type = uint32_t;
         using value_map = std::map<index_type, T>;
+        using iterator = typename value_map::iterator;
 
         bool block_mutation(bool blocked)
         {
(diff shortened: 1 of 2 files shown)


momo5502:
@codex review

momo5502:
@claude review

PR #1360, opened 2026-09-14
Add generic NtUserCallHwnd/Lock/OneParam dispatch table

warsang:
**What this adds:** 

Implements the previously-unhandled NtUserCallHwndLock, NtUserCallHwnd, and NtUserCallOneParam syscalls, plus a new dialog-pointer case in NtUserCallHwndParam.

 Routine numbers are wired to the Windows 10 19041–19045 user-call table: 99 → SetDialogPointer, 111 → SetDialogSystemMenu, 115 → UpdateWindow, 89 → SetMsgBox, 0x39 → ReleaseDC, 0x3B → PostQuitMessage. Unimplemented routine numbers log and return 0/FALSE rather than crashing. 

The windows version is gated with c.win_emu.version.is_build_within(19041,19046) so we don't apply this to other windows versions that would be using different call ids.
```

            if (c.win_emu.version.is_build_within(19041, 19046) && code == user_call_set_dialog_pointer)
            {
                return handle_NtUserSetDialogPointer(c, hwnd, param);
            }
// else falls through to old behavior
```

**Scope/limitations:**

Routine numbers are specific to the 19041–19045 build range; other Windows versions use different numbers and aren't covered and will fall back to previous behavior. This gives a good boilerplate for anyone that wants to implement the same.

UpdateWindow dispatches its paint synchronously; there's no async completion path for callers expecting real async paint semantics under active dispatch.

**Testing**: user.cpp/syscalls.cpp compile, test-sample.exe links, clang-format is clean. Added a test_dialog_table sample test exercising CreateDialogIndirectParamW create/destroy through the new path, which passes. CI green.


momo5502:
@codex review

momo5502:
@claude review

PR #1361, opened 2026-09-15
fix(module): collect ordinal-only exports in both collect_exports overloads

JackTYM:
## Problem

Both `collect_exports` overloads in `src/windows-emulator/module/module_mapping.cpp` (the file-buffer walker and the `map_module_from_memory` walker) iterate `AddressOfNames` for `NumberOfNames` entries and reach the function slot through `AddressOfNameOrdinals`. The PE export directory does not require every function slot to have a name-table entry, so:

- A module whose exports are ordinal-only (`NumberOfNames == 0`) produces an **empty** `exports` vector and an empty `address_names` map.
- A mixed module silently drops every export that has no name entry.

Because `mapped_module::address_names` is what `windows-analyzer/analysis.cpp` uses to emit `function_execution_event` / `foreign_code_transition_event`, calls into an ordinal-only DLL are never logged at all, and calls into an unnamed ordinal of a mixed DLL are attributed via `upper_bound` to the nearest *named* export with a wrong offset. `find_export` cannot resolve them either, and the export count reported by `minidump_loader.cpp` is wrong.

## Fix

Walk `AddressOfFunctions` for `NumberOfFunctions` slots, skip zero (unused) slots, and attach the name through a reverse ordinal -> name-index map built from `AddressOfNameOrdinals`. Exports with no name entry get the conventional `#<ordinal>` name so `find_export("#N")` and `address_names` resolve them. Both overloads get the same change so file-mapped and memory-mapped modules produce identical export lists.

## Reproduction

`C:\Windows\System32\profapi.dll` (x64, from the emulation root) has `NumberOfNames == 0` and 17 non-zero slots in `AddressOfFunctions`. The new test `src/windows-emulator-test/module_exports_test.cpp` maps it through `module_manager::map_module` and compares `mod.exports.size()` against an independent walk of the export directory read back from guest memory.

Against `origin/main`'s `module_mapping.cpp` (swapped in, rebuilt, run, then restored):

```
module_exports_test.cpp:55: Failure
Expected equality of these values:
  mod.exports.size()
    Which is: 0
  count_used_export_slots(emu.memory, mod, directory)
    Which is: 17
[  FAILED  ] ModuleExportsTest.OrdinalOnlyExportsAreCollected
[  FAILED  ] ModuleExportsTest.MemoryMappedModuleCollectsSameExportsAsFileMappedModule
```

With this branch:

```
[       OK ] ModuleExportsTest.OrdinalOnlyExportsAreCollected (7 ms)
[       OK ] ModuleExportsTest.MemoryMappedModuleCollectsSameExportsAsFileMappedModule (3 ms)
```

## How widespread this is

Independent survey (a standalone PE parser, not sogen) of the DLLs in the emulation root that carry an export directory:

- `system32` (x64): 158 DLLs. 2 are ordinal-only and previously mapped with **zero** exports (`profapi.dll` 17, `ntasn1.dll` 60). 37 mixed DLLs previously dropped **2,268** unnamed exports in total, e.g. `dxcore.dll` names=1 used=224 (+223), `combase.dll` 461/598 (+137), `comctl32.dll` 119/181 (+62), `dwmapi.dll` 44/101 (+57), `dcomp.dll` 12/32 (+20), `d3d9.dll` 17/23 (+6).
- `syswow64` (x86): 2,339 DLLs. 38 ordinal-only (including every MFC runtime, `mfc140u.dll` alone has 14,482 exports, all unnamed; `edgeIso.dll` 232, `shunimpl.dll` 453). 134 mixed DLLs previously dropped **27,725** unnamed exports.

## Tests

- New `ModuleExportsTest.OrdinalOnlyExportsAreCollected`: asserts `NumberOfNames == 0`, `exports.size()` equals the number of non-zero function slots, every symbol is named `#<ordinal>`, ordinal is within `[Base, Base + NumberOfFunctions)`, `address == image_base + rva`, `find_export(name)` resolves, and `address_names` contains the address.
- New `ModuleExportsTest.MemoryMappedModuleCollectsSameExportsAsFileMappedModule`: `map_module_from_memory<uint64_t>` over the already-mapped image yields the same `exports` (name/ordinal/rva/address, in order) and an equal `address_names` map as the file-mapped module.
- Full `windows-emulator-test` suite via `gtest_parallel.py` with `EMULATOR_ROOT=root`: 52/52 pass, exit code 0 (macOS arm64, `--preset=release`).

## Notes

- No behavioural change for exports that already had a name: named symbols still get their string name; only previously-missing slots are added. `exports.reserve` now uses `NumberOfFunctions` instead of `NumberOfNames`.
- The `#<ordinal>` naming matches the existing `u"#"` convention already used for ordinal-based class/resource lookups elsewhere in the emulator (`process_context.cpp`, `syscalls/user.cpp`).

### DIFF
--- a/src/windows-emulator-test/module_exports_test.cpp
+++ b/src/windows-emulator-test/module_exports_test.cpp
@@ -1,0 +1,91 @@
+#include "emulation_test_utils.hpp"
+#include "module/module_mapping.hpp"
+
+namespace sogen::test
+{
+    namespace
+    {
+        // profapi.dll exports every function by ordinal only.
+        constexpr auto ordinal_only_module = R"(C:\Windows\System32\profapi.dll)";
+
+        mapped_module& map_ordinal_only_module(windows_emulator& emu)
+        {
+            auto* mod = emu.mod_manager.map_module(ordinal_only_module, emu.log);
+            if (!mod)
+            {
+                throw std::runtime_error("Failed to map profapi.dll");
+            }
+
+            return *mod;
+        }
+
+        IMAGE_EXPORT_DIRECTORY read_export_directory(const memory_manager& memory, const mapped_module& mod)
+        {
+            const auto dos_header = memory.read_memory<PEDosHeader_t>(mod.image_base);
+            const auto nt_headers = memory.read_memory<PENTHeaders_t<uint64_t>>(mod.image_base + dos_header.e_lfanew);
+            const auto& entry = nt_headers.OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_EXPORT];
+            return memory.read_memory<IMAGE_EXPORT_DIRECTORY>(mod.image_base + entry.VirtualAddress);
+        }
+
+        size_t count_used_export_slots(const memory_manager& memory, const mapped_module& mod, const IMAGE_EXPORT_DIRECTORY& directory)
+        {
+            size_t used_slots = 0;
+            for (DWORD slot = 0; slot < directory.NumberOfFunctions; ++slot)
+            {
+                const auto rva = memory.read_memory<DWORD>(mod.image_base + directory.AddressOfFunctions + slot * sizeof(DWORD));
+                if (rva != 0)
+                {
+                    ++used_slots;
+                }
+            }
+
+            return used_slots;
+        }
+    }
+
+    TEST(ModuleExportsTest, OrdinalOnlyExportsAreCollected)
+    {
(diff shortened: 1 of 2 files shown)


momo5502:
@codex review

momo5502:
> You have reached your Codex usage limits for code reviews. You can see your limits in the [Codex usage dashboard](https://chatgpt.com/codex/cloud/settings/usage).

fuck. @claude review

PR #1362, opened 2026-09-15
fix(handles): make handle values 4-aligned like real Windows

JackTYM:
## Problem

Sogen hands out kernel handle values with the object id starting at bit 0, so consecutive handles are `0x800002, 0x800003, 0x800004, ...`. Real NT handles are always multiples of 4: the kernel ignores the low two bits, and user-mode code is allowed to (and does) store private tag bits there.

The consequence is not cosmetic. When guest code tags a handle and later passes it to `NtClose`, the emulator resolves the *tagged* value as a different id and closes whichever neighbouring handle happens to live there. Concretely, on current `main` (x64 guest, unicorn backend), with `before=0x800001`, `target=0x800002`, `after=0x800003` all open:

```
NtClose(0x800003)   // target | 3
  -> STATUS_SUCCESS, `target` still open, `after` closed
NtClose(0x800001)   // target | 1
  -> STATUS_SUCCESS, none of the three probe handles closed (some other live handle was)
NtClose(0x800000)   // target & ~3
  -> STATUS_INVALID_HANDLE (0xC0000008)
```

The same happens for a 32-bit guest (`NtClose(0x800005)` with tag 1 closes `after` instead of `target`; tags 2 and 3 fail with `STATUS_INVALID_HANDLE`).

## Fix

- `handles.hpp`: `handle_value` reserves the low 2 bits (`reserved : 2`, `id : 21` instead of `id : 23`). Every `handle_store` therefore hands out 4-aligned values and lookups ignore the tag bits. The per-store `IndexShift` template parameter becomes redundant and is removed.
- `process_context.hpp`: drops the now-nonexistent `IndexShift = 2` argument from `sections` and `registry_keys`. Those two stores previously got 4-alignment by shifting their own index; every store now gets it from the layout.
- `user_handle_table.hpp` / `syscalls/user.cpp`: user32 indexes the shared `aheList` by the HANDLE's low 16 bits, not by our internal id. With the id now at bit 2, the slot the guest computes is `id << 2`, so `allocate`, `set_owner`, `free_index` and `NtUserMapDesktopObject` go through `handle_index_to_ahe_slot()`. The allocator is bounded by `MAX_HANDLE_INDICES = MAX_HANDLES >> 2` (the ids that still have a slot inside the 0xFFFF-entry table); `used_indices_` and the bound checks in `user.cpp` use the same constant.
- `test-sample/test.cpp`: new `Handle Tag Bits` case. It opens three files, calls `ntdll!NtClose` on the middle one with each low-2-bit pattern (0..3), and asserts `STATUS_SUCCESS`, the target closed, and both neighbours still readable.

## Verification

Host: macOS arm64 (Apple Clang), `--preset=release`. Guest binaries cross-built with mingw-w64 14.0.0 from this branch's `test.cpp` (x64) plus a standalone copy of the new test body for x86 (`test.cpp` itself only builds with MSVC on 32-bit).

`test-sample.exe` (x64), full run, this branch vs. a build with `origin/main`'s `handles.hpp`, `process_context.hpp`, `user.cpp`, `user_handle_table.hpp` swapped in (same test binary):

| backend | main-equivalent | this branch |
|---|---|---|
| unicorn | `Handle Tag Bits`: **Fail**, exit 1 (28 other tests pass) | 30/30 pass, exit 0 |
| FEX | `Handle Tag Bits`: **Fail**, exit 1 | 30/30 pass, exit 0 |
| icicle | `Handle Tag Bits`: **Fail**, exit 1 (27 other pass; Interrupts self-skips) | 29/29 pass, exit 0 |

Standalone probe (same logic as the new test), unicorn:

| guest | main-equivalent | this branch |
|---|---|---|
| x64 | FAIL (handles `0x800002/03/04`; tag 3 closes `after`, tag 1 closes an unrelated handle, tag 0 -> `0xC0000008`) | PASS (handles `0x800008/0c/10`; all 4 tags close exactly `target`) |
| x86 (WoW64) | FAIL (handles `0x800003/04/05`; tag 1 closes `after`, tags 2-3 -> `0xC0000008`) | PASS (handles `0x80000c/10/14`; all 4 tags close exactly `target`) |

`windows-emulator-test` on this branch: 50/50 passed (29s).

### DIFF
--- a/src/samples/test-sample/test.cpp
+++ b/src/samples/test-sample/test.cpp
@@ -1858,6 +1858,58 @@ namespace
         return true;
     }
 
+    bool test_handle_tag_bits()
+    {
+        using nt_close_t = LONG(NTAPI*)(HANDLE);
+        const auto nt_close =
+            reinterpret_cast<nt_close_t>(reinterpret_cast<void*>(GetProcAddress(GetModuleHandleA("ntdll.dll"), "NtClose")));
+        if (!nt_close)
+        {
+            puts("ntdll!NtClose not found");
+            return false;
+        }
+
+        const auto open_file = [](const char* path) {
+            return CreateFileA(path, GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr);
+        };
+
+        const auto is_open = [](const HANDLE file) {
+            char byte{};
+            DWORD read{};
+            return ReadFile(file, &byte, sizeof(byte), &read, nullptr) != FALSE;
+        };
+
+        bool valid = true;
+
+        for (uint32_t tag = 0; tag < 4; ++tag)
+        {
+            const HANDLE before = open_file(R"(C:\Windows\System32\ntdll.dll)");
+            const HANDLE target = open_file(R"(C:\Windows\System32\kernel32.dll)");
+            const HANDLE after = open_file(R"(C:\Windows\System32\kernelbase.dll)");
+
+            if (before == INVALID_HANDLE_VALUE || target == INVALID_HANDLE_VALUE || after == INVALID_HANDLE_VALUE)
+            {
+                puts("Failed to open the probe files");
+                return false;
+            }
+
+            auto* const tagged = reinterpret_cast<HANDLE>((reinterpret_cast<ULONG_PTR>(target) & ~ULONG_PTR{3}) | tag);
+            const auto status = nt_close(tagged);
+
+            if (status != 0 || !is_open(before) || is_open(target) || !is_open(after))
+            {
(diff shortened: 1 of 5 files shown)


momo5502:
hm, we deliberately chose not to follow this, because the leftover amount of bits is low.
an application tagging handles is just buggy, wtf is that even. admittedly, if someone wants to analyze such an app, behaving just like microsoft might make sense.

JackTYM:
> admittedly, if someone wants to analyze such an app, behaving just like microsoft might make sense.

This is mostly the reason I made this a PR, since I assume 1:1 correctness with the real Windows kernel is the end goal of running binaries under sogen

[Microsoft documentation supporting this as a feature](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getqueuedcompletionstatus#:~:text=Even%20if%20you,the%20completion%20port.)

momo5502 (approved):

PR #1369, opened 2026-09-21
fix(user): write the WoW64 userconnect reply in SHAREDINFO layout

JackTYM:
## Problem

A 32-bit (WoW64) guest that creates a window and calls `DefWindowProcA(hwnd, WM_NULL, 0, 0)` faults inside user32 on the first call. Since a typical message loop pumps `WM_NULL`, this hits any 32-bit GUI app that gets as far as a window.

Root cause: the 32-bit `user32!UserClientDllInitialize` raw-copies the 0x238-byte user-connect reply verbatim into `_gSharedInfo` and reads it with the same 8-byte-slotted `SHAREDINFO` layout the 64-bit client uses. Sogen's two WoW64 writers of that reply — the CSRSS ApiPort user-connect reply in `ports/api_port.cpp` and `NtUserProcessConnect` in `syscalls/user.cpp` — both filled a separate, hand-built `WIN32K_USERCONNECT32` struct whose `wndmsg_count`/`wndmsg_table` fields sat in the wrong slots. `SHAREDINFO.DefWindowMsgs` (0x218) and `.DefWindowSpecMsgs` (0x228) therefore stayed null in the 32-bit client, and user32 dereferenced them on the first `DefWindowProc`.

## Fix

- Route both WoW64 writers through the existing `win32k_userconnect::try_write_user_shared_info` / `populate_user_shared_info`, i.e. the same code the native 64-bit connect already uses. WoW64 user32 expects the 64-bit-slotted struct, so no narrowing layer is needed.
- Delete `WIN32K_USERCONNECT32`, its `static_assert`s, `build_wow64_userconnect`, `try_write_wow64_userconnect`, and the now-unused `k_wow64_wndmsg_count` / `k_wow64_ime_msg_count` constants. The `NtUserProcessConnect` writer was a second, independently maintained copy of the same wrong layout; it now shares the one path.
- Add `static_assert(sizeof(USER_SHAREDINFO) == 0x238)` so the reply size user32 copies is pinned in the type itself (the length checks in `api_port.cpp` and `resolve_wow64_destination` now key off `sizeof(USER_SHAREDINFO)`).
- Update `docs/windows-ui-emulation.md`: the section "Fix concept is validated; the clean location is NOT yet found" described this exact investigation as open, with `WIN32K_USERCONNECT32` named as a suspect and a proposed watchpoint-based next step. It is rewritten as resolved (the four globals are `SHAREDINFO.DefWindowMsgs`/`DefWindowSpecMsgs`), and the breadcrumb about the WoW64 path having "no working equivalent" of the message-bit tables is corrected.

Net: +38 / -146 lines across 6 files. No behaviour change for 64-bit guests — `populate_user_shared_info` itself is untouched.

## Reproduction / verification

Minimal 32-bit repro (`i686-w64-mingw32-g++ -static`): `RegisterClassA` + `CreateWindowExA`, then `DefWindowProcA(hwnd, WM_NULL, 0, 0)`, print the result, exit 0 iff it returned 0. Run with `analyzer -s --backend unicorn -e <root> c:/wm-null-repro-x86.exe` (FEX has no WoW64 support, so unicorn is the only applicable backend here).

Before (this branch's five source files swapped to their `origin/main` versions, rebuilt):
```
calling DefWindowProcA(WM_NULL)
Null-pointer call to 0x0; caller return address 0x0 (?+0x0)
analyzer exit code 1.
```

After (this commit, `1e11a8da`):
```
calling DefWindowProcA(WM_NULL)
DefWindowProcA(WM_NULL) returned 0
analyzer exit code 0.
```

`windows-emulator-test` with `EMULATOR_ROOT` set: 50/50 pass on this commit (release preset, macOS arm64 host).

### DIFF
--- a/docs/windows-ui-emulation.md
+++ b/docs/windows-ui-emulation.md
@@ -662,26 +662,25 @@ Dumping the live globals (guest `user32_base = fault_ip - 0x29378`; read `+0xccb
 reaches user32. With the limit at 0, only `DefWindowProc(WM_NULL=0)` reaches the read and `[0+0]`
 faults — and the load loop pumps `WM_NULL`, so it hits reliably.
 
-### Fix concept is validated; the clean location is NOT yet found
+### Resolved: the globals are `SHAREDINFO.DefWindowMsgs` / `DefWindowSpecMsgs`
+
+The four globals are `_gSharedInfo.DefWindowMsgs` (0x218) and `.DefWindowSpecMsgs` (0x228). The
+32-bit `user32!UserClientDllInitialize` raw-copies the 0x238-byte user-connect reply into
+`_gSharedInfo` and reads it with the same 8-byte-slotted `SHAREDINFO` layout as the 64-bit client
+(hence the 8-byte spacing of `…bc0/…bc8/…bd0/…bd8`). The WoW64 connect used to be written in a
+separate `WIN32K_USERCONNECT32` layout whose `wndmsg_count`/`wndmsg_table` fields landed in the
+wrong slots, so the message tables never reached user32. Both WoW64 writers (the CSRSS ApiPort
+reply in `ports/api_port.cpp` and `NtUserProcessConnect`) now go through the same
+`populate_user_shared_info` as the 64-bit path, and `WIN32K_USERCONNECT32` is gone.
+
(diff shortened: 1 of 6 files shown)


momo5502:
@codex review

momo5502:
@claude review

PR #1370, opened 2026-09-21
feat(gdi): decode 1/4/8/16bpp DIBs in NtGdiSetDIBitsToDeviceInternal

JackTYM:
## Summary

`NtGdiSetDIBitsToDeviceInternal` only handled 24/32bpp `BI_RGB` DIBs. Any other depth hit the `unsupported DIB` warning and returned 0 scanlines, so the destination surface was left untouched. This adds 1/4/8bpp (palettised, via the `RGBQUAD` colour table) and 16bpp (RGB555) decoding, and keeps 24/32bpp behaviour unchanged.

Files: `src/windows-emulator/syscalls/gdi.cpp`, `src/samples/test-sample/test.cpp` (+148/-10).

## What changed

- A `dib_pixel_to_bgra32` helper decodes one pixel per bit depth. 1bpp is MSB-first, 4bpp high nibble first, 16bpp is RGB555 (not 565) with 5->8 bit expansion, 24/32bpp read the bytes as B,G,R. Every path composes the pixel from individual bytes, so the result no longer depends on host endianness.
- The colour table is read from `info + biSize` (not a hard-coded 40) and sized by bit depth (`1 << bpp`), not by `biClrUsed`, because pixel data may legally index past `biClrUsed`; unspecified entries decode to opaque black instead of indexing out of range.
- `color_use` (previously ignored) is now honoured for the reject path only: the table is treated as an `RGBQUAD` array only for `DIB_RGB_COLORS` with a `BITMAPINFOHEADER` or later. `DIB_PAL_COLORS` (16-bit logical-palette indices) and `BITMAPCOREHEADER` (`RGBTRIPLE`s) DIBs at <=8bpp still go through the existing warning-and-return-0 path rather than being misread as `RGBQUAD`s. The warning now also prints `usage=`.
- Compression other than `BI_RGB`, `biWidth <= 0`, and bit counts outside {1,4,8,16,24,32} are still rejected as before.

## Test

`test-sample` gains a `GDI DIB` test: a 4x1 top-down DIB at each of 1/4/8/16bpp is blitted with `SetDIBitsToDevice(..., DIB_RGB_COLORS)` onto a 32bpp memory bitmap (`CreateBitmap(4, 1, 1, 32)` selected into a `CreateCompatibleDC`), after which every pixel is read back with `GetPixel` and compared against the expected colour. The pixel patterns are chosen so index order (0,1,2,3 or 0,1,0,1 for 1bpp) and nibble/bit ordering are actually checked, and the 16bpp case encodes pure R/G/B/white in RGB555 so a 565 decode would fail.

## Verification

Host: macOS arm64 (Apple Silicon), `cmake --build --preset=release`, worktree at this commit.

- `windows-emulator-test` (gtest, `EMULATOR_ROOT` set): **50/50 passed**.
- Full `test-sample` cross-compiled from this branch's `test.cpp` with mingw-w64 (x86-64 PE32+), run as `analyzer -s -e root --backend unicorn c:/test-sample-gdi-x86_64.exe`: **30/30 tests `Success` including `GDI DIB`, exit 0, zero `unsupported DIB` warnings** in the log.
- The `GDI DIB` function extracted verbatim into a minimal harness and built for **both i686 (WoW64) and x86-64**:
  - unicorn backend: i686 `GDI DIB: Success` (exit 0), x86-64 `GDI DIB: Success` (exit 0)
  - fex backend: x86-64 `GDI DIB: Success` (exit 0); i686 not runnable, analyzer reports the pre-existing `FEX backend does not support WoW64 (32-bit) processes yet`.
- Negative control (same binaries, same root, `gdi.cpp` temporarily swapped to `origin/main` and analyzer relinked): x86-64 full sample and i686 harness both fail with `NtGdiSetDIBitsToDeviceInternal: unsupported DIB (bpp=1 compression=0 width=4)` followed by `SetDIBitsToDevice(1bpp) copied 0 scanlines`, `GDI DIB: Failure`, exit 1. Restoring this branch's `gdi.cpp` returns them to `Success`.
- Side effect worth noting: on `origin/main` the existing `GDI` test already emits `unsupported DIB (bpp=1 compression=0 width=128)` (a 1bpp DIB it blits internally was silently dropped); with this change that warning disappears and the DIB is drawn.

Note: the `test-sample.exe` shipped in `build/release/artifacts` on this host predates this `test.cpp` and does not contain the `GDI DIB` test, which is why the sample was rebuilt from source with mingw-w64 for verification.


momo5502:
@codex review

momo5502:
@claude review

